PR #6, opened 2015-02-16
gtk3 clipboard fixes

mikesmitty:
We have been manually patching this old bug across various versions of Autokey, but it would be nice to not have to work around it all the time:
https://code.google.com/p/autokey/issues/detail?id=211


guoci:
Thanks. Uploaded to PyPI.

mikesmitty:
Awesome, thanks for merging it so quickly.

PR #71, opened 2017-02-14
return 'None' if key doesn't exist

HyperHamster:
Prevents KeyError if you try to **get_(global_)value** from a key that doesn't exist and also importantly allows existence checking.


HyperHamster:
Failure in check due to poor setup of Travis CI, not this commit.

troxor:
Yeah, Travis CI isn't set up. PR looks fine though, thanks!

PR #83, opened 2017-05-20
update description, update installation section

ersanchez:
Updated description to include TextExpander and Breevy keywords to help new Linux users (and search engines) find AutoKey.
Trimmed "Installation" section so that it points to the wiki - makes this README more concise.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,7 +11,7 @@ AutoKey-Py3
 
 About
 =====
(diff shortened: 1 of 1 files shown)


troxor:
Keeping the basic docs (installation, etc.) as part of the code is more convenient if ever the project needed to migrate off GitHub. The comparison to TextExpander is a nice touch though, but I'd like to keep the Installation docs in README.rst for now.

PR #105, opened 2017-11-05
Ported autokey-run from the legacy optparse module to the new Python3 argparse module

luziferius:
The documentation here ([docs.python.org/3.6/library/optparse](https://docs.python.org/3.6/library/optparse.html)) says this about the _optparse_ module:
`Deprecated since version 3.2: The optparse module is deprecated and will not be developed further; development will continue with the argparse module.`

The port replaces the _optparse_-based argument parser with a new _argparse_-based one.
Additionally, the whole file is now PEP 8-compliant.

### DIFF
--- a/autokey-run
+++ b/autokey-run
@@ -4,6 +4,7 @@
 # -*- mode: python -*-
 
 # Copyright (C) 2011 Chris Dekter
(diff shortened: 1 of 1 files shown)


troxor:
Neat, thanks!

PR #107, opened 2017-11-18
Use XDG_RUNTIME_DIR and XDG_DATA_HOME for lock and log file

### LINKED ISSUE #106: Lockfile and log file shouldn't be in XDG_CONFIG_HOME
## Classification:

Enhancement

## Reproducibility:

Always

## Summary

Lockfile and log file shouldn't be in XDG_CONFIG_HOME

## Steps to Reproduce

Launch the application and look for autokey.log and autokey.pid.

## Expected Results

The files should all follow the [XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).
- autokey.log should be in $XDG_CACHE_HOME/autokey or $XDG_DATA_HOME/autokey (the standard is not clear on that)
- autokey.pid should be in $XDG_RUNTIME_DIR/autokey

## Actual Results

- They both are located in $XDG_CONFIG_HOME/autokey where they should not be, as both files are not configuration files.

## Version

AutoKey-Py3 version 0.93.10, but probably applies to all versions.
Installed via AUR on Arch Linux.

## Notes

The relevant code looks to be located here: https://github.com/autokey-py3/autokey/blob/master/lib/autokey/common.py#L22
If accepted, I could provide a pull request for this issue.

founderio:
Put autokey.log and autokey.lock in their proper places instead of in .config, fixes #106

It could be useful to check for the old log file location and (re)move it? I did not include that as I'm not sure if that is wise.

Sorry for all the whitespace changes - that was my editor. I can revert those if requested.

### DIFF
--- a/lib/autokey/common.py
+++ b/lib/autokey/common.py
@@ -19,8 +19,14 @@
 import os, dbus.service
 
 CONFIG_DIR = os.path.join(os.environ.get('XDG_CONFIG_HOME', os.path.expanduser('~/.config')), "autokey")
-LOCK_FILE = CONFIG_DIR + "/autokey.pid"
-LOG_FILE = CONFIG_DIR + "/autokey.log"
+# Runtime dir falls back to cache dir, as a fallback is suggested by the spec
+XDG_CACHE_HOME = os.environ.get('XDG_CACHE_HOME', os.path.expanduser('~/.cache'))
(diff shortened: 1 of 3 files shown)


troxor (approved):
Looks good, thanks!

PR #121, opened 2018-02-20
[RFC] Code style cleanup / iomediator refactoring

luziferius:
I did some clean-ups.
During the first half, I (mostly) just randomly picked elements to fix from the PyCharm code inspector.

Tested and working:
- Qt4 GUI starts (when the local environment has all dependencies fixed)
- GTK GUI starts
- Setting/using an abbreviation
- Setting/using a hotkey

Not working:
- GTK window property detect dialog: on clicking OK, the dialog freezes (does the same in master branch)

Not all commits are stable:
- Converting Key into an enum [here](https://github.com/autokey/autokey/pull/121/commits/2798a43d5a3a51c2b88d48470b7e3309d339a945) caused IoMediator class to malfunction, this is fixed way later in [this](https://github.com/autokey/autokey/pull/121/commits/4cfdd892530a02110b270a965d78d1d54742ad21) commit.
- Somewhere I broke the listener thread in interface.py, This got unnoticed for a while and is fixed [here](https://github.com/autokey/autokey/pull/121/commits/f65708bb660aa1e5d8048d4f55c9ef33b4e0bb90).
- It seems that I lost quite some imports (looks like the code inspection failed for those), the last commit should fix the GTK GUI.

### Notable cleanups done:
- Removed all `from some_model_module import *` in non-GUI files. I kept those `from PyKDE4.x import *` as this will be tackled later.

- iomediator, iomediator00, iomediator_Key and iomediator_constants reorganised into a package. Now, there is a package called `iomediator` that contains everything. I decomposed the previous 4 modules into public/private modules, each containing a class or related classes. The public classes inside private modules are imported by the package `__init__.py`, so that the whole public interface is available using a simple `from . import iomediator`. (I defined the _public interface_ to be everything that is accessed from outside of iomediator.)


lib/autokey/configmanager.py
```diff
@@ -19,26 +19,19 @@
...
 import glob
 import threading
-import subprocess
+import re
 
-from . import common, monitor
+from . import common
 from .iomediator_constants import X_RECORD_INTERFACE
 
-try:
```
luziferius: The removed try block was a compatibility fix for Python 2.4-2.6. Just import the standard library json module.

luziferius:
@troxor Got time to review the PR?

troxor:
Thanks for the cleanup! Somehow I missed this, but could you please rebase the changes so Github's merging is happy?

luziferius:
Done.

lib/autokey/gtkui/popupmenu.py
```diff
@@ -28,13 +27,17 @@ class PopupMenu(Gtk.Menu):
     A popup menu that allows the user to select a phrase.
     """
 
-    def __init__(self, service, folders=[], items=[], onDesktop=True, title=None):
```
luziferius, line 31: This is a subtle change.
I hope that the code does not rely on implicit sharing of a single list instance across calls.

lib/autokey/model.py
```diff
@@ -849,7 +857,7 @@ def get_description(self, buffer):
         
     def parsePositionTokens(self, expansion):
         # Check the string for cursor positioning token and apply lefts and ups as appropriate
-        if CURSOR_POSITION_TOKEN in expansion.string:
+        if CURSOR_POSITION_TOKEN in expansion.string:  # TODO: Undefined local variable?
```
luziferius, line 860: This is either black magic or never called.

troxor:
Cool, thanks again! Let's see what happens!

lib/autokey/gtkui/dialogs.py
```diff
@@ -424,15 +424,15 @@ def _setKeyLabel(self, key):
         self.keyLabel.set_text(_("Key: ") + key)
         
     def valid(self):
-        if not validate(self.key is not None, _("You must specify a key for the hotkey."),
+        if not validate(self.key is not None, _("You must specify a key for the hot"),
```
luziferius, line 427: Uh. Find and replace did something fancy here…

PR #125, opened 2018-03-11
Add support for F13-F35

HactarCE:
X11 supports keys with not just F1-F12, but also F13-F35. (search [here](https://cgit.freedesktop.org/xorg/proto/x11proto/plain/keysymdef.h) or [here](https://www.safaribooksonline.com/library/view/xlib-reference-manual/9780937175262/16_appendix-h.html) for `_F35`) These are useful not only for keyboards that physically have these keys, but also for using peripherals like gaming mice with extra buttons that can be remapped to unused function keys.

PR #121 would have to be updated to include this, but that should be trivial.


josephj11:
Could someone provide an example/howto of how to use this on [Gitter](https://gitter.im/autokey/autokey) or on [our list](https://groups.google.com/forum/#!forum/autokey-users) ? There have been user requests for this recently. I'd like to add it to our wiki. Thanks.

vandemjh:
Should these keys be added to [this page](https://github.com/autokey/autokey/wiki/Special-Keys)?

josephj11:
If someone can verify that they actually work, then that page should be updated. And we still need examples of how to use them - especially since they do not naturally occur on normal keyboards.

luziferius:
Try a script with
`keyboard.send_keys("<F20>")`
either on `xev` and watch it's console output
or on any application that allows to rebind it's hotkeys. 

But support actually depends on the keyboard layout. My layout does not support them, it doesn’t have the symbols assigned.
You can try it yourself, execute `xmodmap -pk` in a terminal window to print the X11 key map. It outputs a _wide_ table-like structure. First column is the integer key code (side note: These are usable from within AutoKey by using the `<codeXX>` notation, replacing XX with the respective number), then columns with key symbols. Second column contains the symbols for _unmodified_ key presses, later columns contain the values for various modifier combinations.

If the second column contains \<F13> to \<F35>, it is supported by the current layout, otherwise it won’t work.

Edit:
@vandemij Nice catch. I updated the Wiki page.

vandemjh:
@josephj11 @luziferius I could not get fkeys above 12 to work, since as luz mentioned my keyboard layout does not support it.  Looking at `xev` it shows autokey sending individual keys angle bracket, f then 25.  I can also see this by opening any text editor and pressing the hotkey I get `<f25>`

It looks like one [could get them to work](https://askubuntu.com/questions/50751/how-to-generate-an-extended-f-key-press-f13-f20) with some fiddling with `.xmodmap` settings which might look like:
`keycode 320 = F17`
`keycode 321 = F18`

I haven't tested that but it could be useful for having a hotkey you won't ever accidentally press.

PR #134, opened 2018-04-12
Removed old testing stuff in interface.py.

luziferius:
Another small one: Remove old testing stuff in interface.py.
The XLibInterface instantiated during the test does not exist, thus the test won’t run even if tried.

### DIFF
--- a/lib/autokey/interface.py
+++ b/lib/autokey/interface.py
@@ -34,7 +34,7 @@ import Xlib.threaded as xlib_threaded
 del xlib_threaded
 
 
-from autokey import common
+from . import common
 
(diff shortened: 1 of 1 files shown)


troxor:
Last time I checked, absolute imports were considered superior to relative. I changed a bunch of them a while back. PEP8 or somesuch.

luziferius:
Awww. Now you have merged it… I’ve re-read PEP8, and actually yes, it says absolute is better than explicit relative imports. I’d have changed it, if the PR were still open.
For the next PRs, I’ll keep myself to (mostly) absolute imports.

PR #140, opened 2018-05-10
Fix autokey/autokey#139

### LINKED ISSUE #139: [BUG] Abbreviations fail to trigger following tab keystroke
## Classification:
Bug

## Reproducibility:
Always

## Summary

Abbreviations following a tab keystroke are not captured.

## Steps to Reproduce

* Scenario 1
    1. Press tab
    2. Enter abbreviation trigger (`.ask`) then space
        * Abbreviation doesn't trigger
* Scenario 2
    1. Press tab
    2. Enter abbreviation trigger (`.ask`) then space
        * Abbreviation doesn't trigger
    3. Backspace to remove trigger (`.ask`)
    4. Enter abbreviation trigger (`.ask`) then space
        * Abbreviation doesn't trigger
* Scenario 3
    1) Press tab
    2) Enter abbreviation trigger (`.ask`) then space
        * Abbreviation doesn't trigger
    3) Enter abbreviation trigger (`.ask`) then space
        * Abbreviation is triggered
        * First instance of abbreviation still present before expanded text

## Expected Results

- Abbreviation should trigger regardless of preceding whitespace keystrokes

## Actual Results

- Abbreviation fails to trigger if preceding keystroke is a tab

## Version

AutoKey-Py3 version: autokey-gtk 0.90.4, python 2.7.15rc1 and python 3.6.5 installed

Installed via: PPA

Distro: `Kubuntu 18.04 LTS 4.15.0-20-generic #21-Ubuntu x86_64`

## Notes
Issue presents in multiple applications, confirmed KVIrc (irc client), Discord, and yakuake/konsole

Noticed when pressing tab for name completion, tested with tab alone following mouse input.  Issue appears specifically tied to the tab keystroke.  Backspace to undo expansion is enabled.

Logs: [https://gist.github.com/sporkwitch/6869d80bf01ca35c09fb1a75b96e80a8](url)

Symptoms similar to #95 but does not appear related.

sporkwitch:
Small fix to resolve #139, have it treat any non-whitespace character as a match, rather than taking it from wordChars.

vim seems to have also nuked a bunch of extraneous whitespace characters on otherwise empty lines.  Can try to prevent that and resubmit if the messy visual comparison is a bother.

### DIFF
--- a/lib/autokey/model.py
+++ b/lib/autokey/model.py
@@ -54,7 +54,7 @@ def get_value_or_default(jsonData, key, default):
 def get_safe_path(basePath, name, ext=""):
     name = SPACES_RE.sub('_', name)
     safeName = ''.join([char for char in name if char.isalnum() or char in "_ -."])
-    
+
     if safeName == '':
         path = basePath + '/1' + ext
         jsonPath = basePath + "/.1.json"
@@ -68,7 +68,7 @@ def get_safe_path(basePath, name, ext=""):
         path = basePath + '/' + safeName + str(n) + ext
         jsonPath = basePath + '/.' + safeName + str(n) + ".json"
         n += 1
-        
+
     return path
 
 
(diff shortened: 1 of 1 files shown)


troxor:
Cool, thanks! 

You can also hide whitespace changes by adding ?w=1 to the end of the github url, like so:
.../autokey/pull/140/files?w=1

sporkwitch:
You would think that would be the default lol; the only thing that really matters with whitespace is indentation and tabs-vs-spaces (the answer is clearly spaces :P)

Issue #144, opened 2018-05-14
Main Window doesn't remember open folders


cbrchrt:
Classification:
UI/Usability

Reproducibility:
Always

Summary
When opening the main window I have to

Steps to Reproduce
Select 'Show Main Window' on the Autokey icon in the desktop panel bar.

Expected Results
Main window should open with the folders open from the previous time.

Actual Results
All folders are closed.

Version
0.94.1

Installed via: PPA

Distro: Ubuntu 18.04

Notes
Re-opening as suggested.

luziferius:
Side note: This happens both in the Qt and GTK version.

I think this happens because the GUI window is destroyed and completely removed from memory when closed and recreated from scratch when opened again. This loses the current state, including opened tree elements (folders). (The Qt code looks like working like this…)

@cbruchert Would it be sufficient for you to remember the opened folders during the runtime (from AutoKey start, across window closes/re-opens, until exit) or across program restarts (this includes OS/PC restarts)?

sporkwitch:
@luziferius worth noting, he would be working from the GTK version; I haven't included the Qt in the 18.04 PPA because it lacks the Qt deps

For my two cents, I think during a session would be sufficient.

cbrchrt:
@luziferius Maybe having it remember across program restarts would be nice too, but for me the most important is during the session.

luziferius:
Session-only allows the easiest solution, because there is no persistent storage involved that can desync from the file system content.

josephj11:
@luziferius 's repo is still there with all these PRs in it. IDK if any of this made it into develop.

There's another related issue to this. #703

Elliria:
I agree that it should remember its state during an open session, but having it remember its state from session to session should be optional. I would opt out of that, because I'd find it disorienting for the program to open in anything but its default state on its first load of a session.

josephj11:
@Elliria Persistence for the duration of the session makes the most sense to me. All these commits are so old though that they could only be used as suggestions at this point.

I think this should be fixed, but that it's pretty low priority ATM.

Elliria:
Yep. I agree. They're popping up in my [CodeTriage](https://www.codetriage.com/) emails, so I check them to see if they're easily fixable or comment on them or label them, as the case may be.

Issue #90, opened 2017-07-16
Cannot add phrases / scripts, anything.  I fail to find any means to do so.


deragon:
## Classification:

Bug

## Reproducibility:

Always

## Summary

Cannot add phrases / scripts, anything.   I fail to find any means to do so.

## Detailed description

I fail to find any button or means to add a phrase.  However, whenever I right click on an entry (such as 'My phrases', the following error appears on the terminal:

```
  File "/usr/local/lib/python3.5/dist-packages/autokey/qtui/configwindow.py", line 662, in on_treeWidget_customContextMenuRequested
    menu.popup(QCursor.pos())
AttributeError: 'NoneType' object has no attribute 'popup'
```

## Version

AutoKey-Py3 version:  latest, up to date master branch from git://github.com/autokey-py3/autokey-py3.git as of 2017-07-16 04:12:18 EDT

Distro:  Ubuntu 16.04 LTS Xenial Xerus, up to date as of 2017-07-16 04:12:18 EDT

## Notes

I tried AutoKey previously and found it buggy, so I never used it really; but I was able to add my phrases.  I now wanted to try it again, updated my git repository, fire it up again and to my surprise, I fail to find any button to add a phrase in the main window.  I can edit the current ones, but cannot add anything.

## Screenshot

Screenshot of what I see, with no button or means to add a phrase or entry.  Remember, right-click an entry shows nothing on screen, but an error message shows up.

![autokey_006](https://user-images.githubusercontent.com/7698377/28490888-fa1f1ae4-6eb2-11e7-8510-2aecb5707540.png)

troxor:
You might be stuck using the GTK frontend. The Qt version relies on a bunch of deprecated-upstream libraries, and may be deprecated itself soon unless someone knows Qt and wants to port the interface to Qt5. Just be sure to make sure any previous installations have been removed first, as there have been reports of conflict in the past.

luziferius:
@deragon Are you still interested in using autokey?
I refactored/rewrote the Qt based UI (here: https://github.com/luziferius/autokey/tree/qt4_ui_cleanup).
Would you mind checking it out and testing if the new version works for you?

luziferius:
Qt GUI is re-written, closing this for now.
Please comment/re-open if this bug still applies.

Issue #77, opened 2017-03-28
AutoKey is already running as pid 1776 but is not responding


davduf:
Hello great team !

## Summary

AutoKey is already running as pid 1776 but is not responding

## Actual Results

org.freedesktop.DBus.Error.ServiceUnknown: The name org.autokey.Service was not provided by any .service files

## Version

AutoKey-Py3 version qt

Installed via: (PPA, pip, …) DNF

Distro: Fedora 25

Thank you !

troxor:
Try removing ~/.config/autokey/autokey.pid . If you can reproduce this, it would be worth implementing a more permanent solution.

nick-s-b:
I'm getting the same error at least once a week. I think it might have something to do with with update of some of the python library packages but I'm not sure. I'm running it under Arch.

ineuw:
Using Linux Mint 18.1 Cinnamon 32bit and 64bit, I have the same error but it is easily correctable in my case.When  starting Autokey-qt in the Startup Applications, the system tray icon is not showing, (it is running, but I am never sure.), so I click on the menu and get the same error message.
The  "Show a notification icon" is selected in Preferences, but the icon is displayed only when Autokey is manually started on the menu.
PS. There is no autokey.pid created with this installation, but I know what you are referring to.

luziferius:
This should be fixed in 0.95.1, which contains a rewrite of the Qt GUI. It both contains a rewrite of the DBus handling and tray icon handling.

The report is quite old, so I just close it now. Please comment or re-open, if the issue still persists in 0.95.1 (or newer).

Vinzent03:
3 years later, I'm facing this issue.
version 0.95.10
installed via pip on KDE neon 5.23.3
Qt-Version: 5.15.3
```
AutoKey is already running as pid 3530 but is not responding

org.freedesktop.DBus.Error.ServiceUnknown: The name org.autokey.Service was not provided by any .service files
```
I enabled the autostart setting, which causes the error window to pop up, every time I start my pc.

josephj11:
@vinzent03 Are you on Arch or one of its derivatives as well?

The first thing to check is whether you have KDE set to restore the previous session when it starts. It might be launching AutoKey and then the AutoKey entry in ~/.config/autostart/autokey.desktop may be trying to do it as well.

I have no issues with this, but I'm way behind you with kubuntu 18.04.

[...]

Issue #56, opened 2017-01-05
Update autokey-qt frontend to PyQt5


troxor:
PyQt4 is being deprecated in some distributions (notably Archlinux).

troxor:
Suspending work on this for now due to the quantity of changes and limitations on time. I also don't run Qt-based desktop environments. Pull requests are welcome.

xuzhao9:
Hello @troxor  I am interested in porting the Qt app to Qt5.
However I found PyKDE5 is in pretty bad shape and is not usable.
Is it acceptable that I remove the PyKDE dependency and make it a pure Qt5 app?

troxor:
Hi @xuzhao9, that would be awesome. I'm sure there are many users that would appreciate an updated Qt5 interface!

supagu:
great, would love to see a qt5 interface so i can use this app!

xuzhao9:
Hi guys, how's everything going? I have finished porting a simple but runnable autokey-qt5 interface without PyKDE dependency. However, I have to write a middle-layer to translate the PyKDE related code to the PyQt so it may lose some of the original features. The translation layer code is in lib/autokey/qt5ui/qt5helper.py under the repo root.

You guys can go to my [forked repo](https://github.com/xuzhao9/autokey) to give it a try. Just run the autokey-qt5 script!
Remaining TODOs:
- Use QIcon to replace KIcon (some of the KDE global icons may not reachable?)
- I didn't implement KXmlGuiWindow, so setAutoSaveSettings doesn't work for now. Not sure if other things doesn't workout though.


Here is a [screenshot](https://ibb.co/k0AY7G) of my port, please give my comments and feedbacks before I can submit a PR.

troxor:
Thanks @xuzhao9 ! I think it looks good so far, but feedback from those who use the current autokey-qt would be valuable. We can replace anything Qt4/PyKDE-related with your port, and make it the only Qt interface if you agree that this would simplify things.

xuzhao9:
@troxor Hello! I met a problem when porting the app. At the starting of the autokey-qt5, the following error message is printed:
X protocol error:
<class 'Xlib.error.BadAccess'>: code = 10, resource_id = 666, sequence_number = 27, major_opcode = 33, minor_opcode = 0
X protocol error:
<class 'Xlib.error.BadAccess'>: code = 10, resource_id = 666, sequence_number = 28, major_opcode = 33, minor_opcode = 0
X protocol error:
[...]

PR #154, opened 2018-05-31
[RFC] Remove PyKDE4 dependency, Qt UI rewrite. Ported to PyQt5.

### LINKED ISSUE #131: compileuic.sh   breaks build
after a pykdeuic4-3.5   run, autokey is totally broken. this app is low pre-alpha quality
not even the paths in  compileuic.sh    are correct. what is this ?

pristine master does build + run though. bad first impression.

luziferius:
Let’s resurrect the Qt GUI…
This time, I’d like to get some feedback and testing done _before_ this gets merged. @troxor \*wink\*
Issue tracker that can be used for this: https://github.com/luziferius/autokey/issues
# Dependencies
- Dropped: PyKDE4, PyQt4
- New: PyQt5
- New install/build-time dependency (Optional): `pyrcc5` CLI tool to compile Qt resource files. (Ubuntu package: `pyqt5-dev-tools`)

This is used to compile the UI files and AutoKey Icons into a python module. Also copying the icons guarantees that all AutoKey (`A`) icons are present under all circumstances, even if not present in the system icon theme. Missing icons happens when AutoKey is installed in a local Python virtualenv.
(All other icons are still loaded from the system icon theme.)
Unlike previously with python classes generated by pykdeuic4, the generated module must not be added to the repository. It is automatically generated during setup.py execution. This guarantees that all users always have the latest UI changes, without us having to regenerate automatically generated files.
If this tool is not present, the resource folder and icons are copied as-is as additional package data. The Qt GUI transparently handles both cases.
Thus installing autokey does not pull a hard build-time PyQt5 dependency in, for users only interested in the GTK GUI.
- New: `python3-pyqt5.qtsvg`. This is used for a workaround. See https://github.com/autokey/autokey/pull/154/commits/fbbb9303e44f55723774dc0bd41fbbed4ac1ea33 for details
- Dropped dependency: Python module `dbus.mainloop.pyqt5`.

Previously, autokey-qt used `dbus.mainloop.qt` to integrate the dbus interface with Qt. This caused me some threading related problems after the refactoring. Thus I stopped using the generic implementation in `autokey.common` and instead use the DBus interface implementation directly built into Qt5/PyQt5.
# Things done
- Cleaned up all UI files in the UIC directory, removed all KDE specific widgets and replaced with plain Qt ones.
  - Removed the `compileuic.sh` script and moved all UI files out of the uic directory. Fixes #131
  - Now, those live under `autokey.qtui.resources.ui`
- Replaced all pykdeuic4 generated code with those cleaned up UI files.
  - Classes using designed UI widgets load the files once during startup and directly inherit from the runtime-generated UI classes.
- Decomposed large source files.
  - All dialogs are grouped in a new `dialogs` package
  - All Settings/Preferences related widgets are grouped in a new `settings` package
- Many code style cleanups
  - e.g. replacing camelCaseVariableNames with the recommended_underscore_lowercase_style.
  - Removed all `import *` and `from x import *` imports
- Removed all PyKDE4 dependencies
  - Designed a new main application window replacing the KXMLGUI window.
  - Designed a new Settings dialog.
- Streamlined the `autokey-qt` script and moved to `autokey.qtui.__main__`
Defined a new entry point in setup.py named `autokey-qt`. This now works like the GTk-GUI.
- Reworked the window hide logic. Instead of destroying the window and re-creating it, it is merely hidden.
  - Now with 90% less object destructions/re-creations
  - Those nasty try-except blocks catching RuntimeErrors/AttributeErrors used as normal control flow are gone.
  - All internal state is kept, including #144, even text/item selections are kept across window closes/re-opens
- The enable monitoring/expansion/service checkboxes both in the settings menu and tray icon sync with the internal service activation state, meaning that the checkbox visibly toggles when the user presses the global hotkey.
- Improved the popup menu functionality: The menu elements now have proper icons depending on their types (Script, Phrase or Folder).
  - By extension, this works for the tray icon context menu, too!
- I tried fixing non-code files (like packaging files in the `debian` folder, but I don’t know if this went well)
  - (Hopefully completely) removed KDE related packages from various dependency lists.
- Compile the AutoKey UI files and other resources used by the GUI into a resource file when installing.
Tell setup.py to use the pyrcc4 CLI tool to build the file and deploy this compiled resources file instead of shipping the plain resources directory like it is done now.
- Ship the application icon in the resources directory, so that it is displayed even if it is not found in the system icon theme (like when installed using pip).
- Designed a VERY basic About AutoKey dialog. 
- Port from PyQt4 to PyQt5.
  - This should be way easier than ripping out PyKDE4.
  - Edit: It was quite easy: 
    - Adjusted imports for the new Qt package layout (mostly importing from QtWidgets instead of QtGui).
    - Replaced some deprecated function calls with the recommended ones
    - Fixed Icons, because PyQt5 did not like my static icon assignments…
- Use the DBus implementation built into Qt instead of `python-dbus`. autokey-qt does no longer depend on `dbus.mainloop.pyqt5` (the Qt5 equivalent to `dbus.mainloop.qt), which had caused problems in the past.
  - The new implementation does export a bit more than the `python-dbus` implementation still in use in the GTK GUI, but, as long as user code does not depend on the new, fancy introspection exports, both implementations are compatible.
# Future work
- https://github.com/luziferius/autokey/issues/6
- https://github.com/luziferius/autokey/issues/8
- https://github.com/luziferius/autokey/issues/9
- https://github.com/luziferius/autokey/issues/11
- https://github.com/luziferius/autokey/issues/17
- https://github.com/luziferius/autokey/issues/18


troxor:
Wow, this is a ton of progress towards #56 . I don't personally run Qt on my system, but @xuzhao9 and others may be interested.

sporkwitch:
I _do_ use Qt (KDE FTW), so this is great :)  Not looking forward to fixing the deps in the control file, though lol

luziferius:
@sporkwitch You may want to look at the last few commits. I ripped out everything KDE related from all dependency lists I found.

sporkwitch:
@luziferius yup, i got that, it's more just resolving which packages pull what from the ubuntu repos.  Probably will be easier than the initial bionic setup, since the changes are recent, not historic to dig through, so less trial-and-error heh

Thanks for the hard work, mate

luziferius:
Ok. I did the last steps and ported the whole thing to PyQt5.
This should fix
- #56 
- Maybe #90, This looks like it is a weird Qt UI issue (probably caused by global menus).
- Maybe #77

Edit: I successfully built Debian/Ubuntu packages from the current branch HEAD, which look like working flawlessly (except for known, not UI related bugs, ofc ;) ).

luziferius:
@troxor I think this can be merged. I ironed out all breaking bugs I found, autokey-qt now works smoothly on PyQt5.

PR #211, opened 2018-11-06
Document how to launch Autokey directly from the repo

dogweather:
I wrote up the method for launching Autokey without needing to install anything - a very handy trick, from @luziferius .


luziferius:
Looks good, merged

Issue #197, opened 2018-10-10
Abbreviation case folding (ignore case) not working with abbreviations defined as upper case


luziferius:
## Reproducibility:

Always

## Summary

Abbreviation setting "ignore case" does not work, if the abbreviation *stored in the item* is UPPER CASE.

## Steps to Reproduce (if applicable)

- Add an upper case abbreviation (e.g. `AB`) to a Phrase.
- Tick the "ignore case" check box in the abbreviation dialogue and accept the dialog.
- Save
- Try to trigger the phrase. Neither `ab`, `Ab`, `aB`, nor `AB` triggers the phrase.

## Expected Results

- Phrase expansion should happen.
- If also checked, case matching should be performed

## Actual Results

- Instead, nothing happens.
- There are no exceptions thrown. The case folding logic seems to rely on abbreviations entered as lower-case for case-folding to work.

## Version

AutoKey version: 0.95.3; git develop HEAD

Used GUI (Gtk, Qt, or both): Qt

If the problem is known to be present in more than one version, please list all of those.

Installed via: (PPA, pip3, …).

Distro:

## Notes

The bug is in `Phrase.check_input(buffer: str) -> bool` or functions used by it.

PR #214, opened 2018-11-08
WIP: Restore unit tests

dogweather:
(Created this temp PR just to discuss it.)

Thanks @luziferius -

EDIT: I set up my dev environment with `pipenv` and ran the tests with `pytest`. I'm getting just four failures - looking good.

```
4 failed, 44 passed, 37 xfailed, 1 warnings
```


luziferius:
I looked around and it seemed to be good practice to incorporate the test launch into `setup.py`
So i integrated it and launch tests using `python3 setup.py test`
You get the same results as I.
The 37 xfailing tests are caused by #197. I just marked them as xfail to reduce output verbosity for things I know are broken at the moment.

dogweather:
:+1: I'll give `setup.py test` a shot.

I totally agree with just marking them as expected-to-fail, getting things to a known state.

PR #282, opened 2019-05-12
Change timeouts for wait_for_exist() and wait_for_focus()

Drugoy:
1. Allow timeout to be 0 (which means no timeout, i.e. wait indefinitely).
2. Make default timeout be 0.

### DIFF
--- a/doc/scripting/lib.scripting.Window-class.html
+++ b/doc/scripting/lib.scripting.Window-class.html
@@ -106,7 +106,7 @@
         <tr>
           <td><span class="summary-sig"><a href="lib.scripting.Window-class.html#wait_for_focus" class="summary-sig-name">wait_for_focus</a>(<span class="summary-sig-arg">self</span>,
         <span class="summary-sig-arg">title</span>,
-        <span class="summary-sig-arg">timeOut</span>=<span class="summary-sig-default">5</span>)</span><br />
+        <span class="summary-sig-arg">timeOut</span>=<span class="summary-sig-default">0</span>)</span><br />
       Wait for window with the given title to have focus</td>
           <td align="right" valign="top">
             <span class="codelink"><a href="lib.scripting-pysrc.html#Window.wait_for_focus">source&nbsp;code</a></span>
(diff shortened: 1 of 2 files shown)


luziferius:
Thank you for your time to improve the API.

I’d like to not include "surprises" in the current git master and thus in the next bugfix release(s).
So including the feature to be able to wait indefinitely is fine, but changing the default behaviour to that is not. Because this change alters the behaviour of existing user scripts (maybe hanging them indefinitely!), I’d like to postpone the latter to the next major release, which will be 0.96.0.

If you wish to continue to contribute improvements for the scripting API, please take a look at the current [`develop`](https://github.com/autokey/autokey/tree/develop) branch, which will be released as 0.96 once it accumulated enough features and improvements. [This is how the API modules look like, currently.](https://github.com/autokey/autokey/tree/develop/lib/autokey/scripting) I’ve split the modules and re-arranged them in a separate package, so working on a single API class now involves way less scrolling around and potential interference from other classes.

If you work against `develop` for larger things, it’ll make including the work much easier.

Drugoy:
Thomas, I don't think I get you.
Will you accept the patch if I merge it against `develop` branch?
Are you against changing the defaults at all? If yes - this is a road to hell, as you'll end up being a hostage of legacy.
In my opinion, AutoKey is quite at the start of it's way and needs quite drastic changes, I think it's not a good time yet to stabilize the project's defaults/APIs/etc.
I would even go further and rename a lot of things, as reading the code I've felt that current naming is just wrong for some methods.
For example AutoHotkey's commands WinWait/WinWaitActive/WinWaitClose by default wait indefinitely and it makes way more sense to me than a weird value of 5 seconds: why 5? why not 10 or 1 or 0.5 or anything else?

luziferius:
Edit: For your first question: Yes.


I'm currently against changing the behaviour of defaults in the stable master branch. Because AK was almost a zombie project for several years when the original developer left, the API got 'stable' (read as 'no changes') from the user point of view due to lack of development. I'm planning on possibly releasing additional bugfix releases for 0.95 (see the Github milestone), which should not change behaviour, unless that change fixes bugs.

The next feature release (0.96), currently accumulating commits in the `develop` branch, will introduce quite some changes to the API, and that's fine. Changing defaults, adding, extending or removing functions is OK, because that'll be clearly communicated on release.

I just want to go the route of least surprise for the (possibly few) users AK has.

PR #331, opened 2019-12-06
Scripting API: Allow creation of 'temporary' phrases and folders

BlueDrink9:
* Adds API function to create folders from scripts (documented with docstrings, tests written)
* Adds 'temporary' arg to phrase and folder creation
* Adds function for removing all temporary phrases/folders (tests written in wrong test file (since configmanager has no test file and I don't know how to set one up))

This optional arg 'temporary' means the phrase will not be persisted across restarts of the program, and is easily removed.

This is for phrases created by scripts, which may be run or changed frequently.

The use-case I have for this is best outlined in the following example:

```python

# Allows switching wasd between sending arrow keys and sending regular keys.
#
# Uses pseudocode that assumes the ability to call script functions from
# phrases.
folder = engine.create_folder("temporary_demo", temporary=True)

mode = "normal"

def switch_mode(mode):
    configManager.remove_all_temporary()
    if mode == "normal":
        normal()
    elif mode == "movement"
        movement()
    else:
        print("Invalid mode specified")
        normal()


def normal():
    engine.create_phrase(folder, "w", "w", temporary=True)
    engine.create_phrase(folder, "a", "a", temporary=True)
    engine.create_phrase(folder, "s", "s", temporary=True)
    engine.create_phrase(folder, "d", "d", temporary=True)
    engine.create_phrase(folder, "F10", "<function name=switch_mode args='movement'>", temporary=True)

def movement():
    engine.create_phrase(folder, "w", "<up>", temporary=True)
    engine.create_phrase(folder, "a", "<left>", temporary=True)
    engine.create_phrase(folder, "s", "<down>", temporary=True)
    engine.create_phrase(folder, "d", "<right>", temporary=True)
    engine.create_phrase(folder, "F10", "<function name=switch_mode args='normal'>", temporary=True)
```


tests/scripting_api/test_engine.py
```diff
@@ -135,9 +135,78 @@ def test_engine_create_folder():
...
-    "Doesn't add parent folder as parent of subfolder")
+        "Doesn't add parent folder as parent of subfolder")
+    assert_that(folder.folders,
+        has_item(test_folder),
+            "Doesn't create subfolder in correct folder")
+
+
+# These tests shouldn't really be here since they're for configmanager
+# class, not engine. But I'm not confident enough with settings up
+# tests to move them.
```
luziferius: Tests are autodetected by pytest based on the file name, which must be `test_whatever.py` or `whatever_test.py`.
Just create a new python file called `test_configmanager.py` or create a whole package named `configmanager` under `tests` with a `test_<some_name>.py` inside. I’ll leave that choice to you, as you see fit.

Move the tests to the new test module, import things from test libraries and autokey as needed and copy any setup boilerplate code you currently use there to decouple them. After you decoupled them, you can alter the copy to suit the specific needs.
Any non-test functions and structures I defined at the beginning of test files are written to reduce test code length and write concise code, which is tailored to the specific needs of the test module. In my opinion, those should not be shared using imports as that introduces dependencies.

tests/scripting_api/test_engine.py
```diff
@@ -135,9 +135,78 @@ def test_engine_create_folder():
...
+    assert_that(folder.folders,
+            has_item(test_subfolder_nontemp),
+                "Removes non-temp subfolders")
+    assert_that(folder.folders,
+            not_(has_item(test_subfolder)),
+                "doesn't remove temp subfolders")
+    assert_that(test_subfolder.items,
+            not_(is_(equal_to(test_phrase))),
+                "doesn't remove temp phrases from temp subfolders")
+    # I'm less certain that this should be the defined behavior.
```
luziferius: Temporary folders with non-temporary content seems fishy. So yeah, I’d say that this should not be permissible.
Something like `temp_folder/permanent_subdir/temporary_sub_subdir/again_permanent_dir` sounds like it creates hundreds of edge cases and weird behaviour no-one expects. And I can’t come up with any valid use case.

And how should one save a permanent if the temporary parent disappears?
So there are two possible solutions I see:
- Disallow setting `temporary=False`, if `parent.temporary == True`
  - Throwing an error in this case
  - Silently set `temporary=True` (This does not sound like a good design)
- Automatically set `temporary=False` recursively on parents, if a non-temporary item is created inside a temporary one. Again, this might not be what users expect.

lib/autokey/scripting/engine.py
```diff
@@ -99,7 +99,8 @@ def create_phrase(self, folder, name: str, contents: str,
                       temporary=False):
         """
         Create a new text phrase inside the given folder. Use C{engine.get_folder(folder_name)} to retrieve the folder
-        you wish to create the Phrase in.
+        you wish to create the Phrase in. If the folder is a temporary
+        one, the phrase will be created as temporary.
```
luziferius, line 136: Sounds reasonable.

luziferius:
Thank you for taking your time and contributing to AutoKey.

The feature sounds interesting. It can be used to automate specific tasks that have to be done multiple times, but only once during a single session. One thing that needs to be fleshed out is accessibility of this function from outside the API. For example, should the GUI be able to create temporary items? If so, how? (checkbox in creation dialogues or designated items in the context menu)

Currently, the GUI does not update when calls to the API create items. The GTK GUI can be closed and re-opened to show them. But this happens only because closing the window destroys the whole GUI including all data hold by it. Clicking the tray icon re-creates everything from scratch, which implicitly updates the tree view. The Qt5-GUI is only hidden, thus doesn’t update at all by closing and re-opening the GUI. Updating the GUI on API calls that create stuff is unrelated to your PR, but should be addressed later. Otherwise, the temporary scripts can’t be seen from the GUI at all.

About your example: You can use the `store` to permanently store python objects. It is a persistent, per-script local dict that can hold anything. (Everything that is not json-serializable will be discarded on AutoKey exit.) The global store (`store.GLOBALS`) is shared between all scripts. So you can use the store to signal a flag and use it inside scripts to alter behaviour.

lib/autokey/scripting/engine.py
```diff
@@ -155,13 +202,20 @@ def create_phrase(self, folder, name: str, contents: str,
                 p.set_hotkey(*hotkey)
             if window_filter:
                 p.set_window_titles(window_filter)
+            # XXX: Could these next 3 be refactored to just set p.<val> to the
```
luziferius: The function is annotated to accept booleans. But it does not type check the input, so the user may pass anything truth-y or false-y, like `[1]` and `[]`. This if implicitly calls bool().

I’d say yes to your question, it can be refactored the way you proposed. It should be even better, because that is more explicit.
But then wrap the user passed value in bool(), like `p.prompt = bool(always_prompt)`. Otherwise a list or string (or whatever else user provided cruft) ends up in the data structure and later in the json serializer. I already discovered issues with this kind of issue when I tried to store function objects in the `store` API.

tests/scripting_api/test_engine.py
BlueDrink9: I'll throw an error.

lib/autokey/scripting/engine.py
BlueDrink9: Good catch. I'll do that.

BlueDrink9:
I'd like to argue that the GUI doesn't need to be able to create these. The use-case I see for temporary hotkeys is for scripts. Two reasons:

* Unless you are only making one or two temporary hotkeys (in which case, may as well just make and delete them) doing them by hand in the GUI will be tedious. Hence, script them.
*  You may want temporary hotkeys because they are constantly changed, for example bindings for a window manager. Therefore, you might set the script to autostart with autokey and be confident that any changes are reflected in the hotkeys, instead of adding new keys.

I agree that the GUI should be able to *see* them. Is there a function I can call to allow that to happen? Or would it be a whole lot of hassle?

Thanks for the tip with the store, I'll keep that in mind. Were you suggesting that as storing a flag for which mode (movement vs normal) the user is in?

luziferius:
> I agree that the GUI should be able to _see_ them. Is there a function I can call to allow that to happen? Or would it be a whole lot of hassle?

I don’t know how much hassle it is. Both GUIs need a new callback function that is called by the scripting API after creating items. That callback updates the tree view. It should add the created item without changing the selected item(s). This is needed regardless of this PR, since the API is already able to create permanent Phrases.

> Thanks for the tip with the store, I'll keep that in mind. Were you suggesting that as storing a flag for which mode (movement vs normal) the user is in?


Basically yes. For example, alternate between typing `y` and `n`:

```python
state = store.get("some_flag", True)
keyboard.send_keys("y" if state else "n")
store.set_value("some_flag", not state)
```

For something more advanced to hint at the possibilities, look at https://github.com/luziferius/autokey_scripts/tree/master/NumpadIME. This was something someone requested on IRC, but got turned down as being impossible by others before I got to interfere.
It basically allows typing text using the keypad. You just have to assign the keypad number keys as hotkeys for the individual scripts after cloning it into your AutoKey scripts.

But back to the PR: Do you consider your PR as complete and ready to merge?

luziferius:
I pushed some stale work to develop I had already finished. I think the merge conflict was caused by your IDE removing a whole lot of whitespace characters.
But resolving it should be easy. You can either rebase on top of `upstream develop` or resolve the conflict by accepting the sections that add code and discard the empty ones.

BlueDrink9:
Trivial fix. I'll avoid editing whitespace too much in future.

Will be ready to merge with one more commit. Nontemp within temporary now throws an error. I'll do the refactoring too.

BlueDrink9:
Did the refactor, and as part of that I ended up adding type validation to all of the inputs into `create_phrase`.

Now ready to merge

lib/autokey/scripting/engine.py
```diff
@@ -177,11 +178,80 @@ def create_phrase(self, folder, name: str, contents: str,
...
+                    # Then check second element is a key or str
+                    if not isinstance(hotkey[1], str) and not \
+                            isinstance(hotkey[1], model.Key):
+                        fail=True
+                else:
+                    fail=True
+            else:
+                fail=True
+            if fail:
+                raise ValueError("Expected hotkey to be a tuple of modifiers then keys, as lists of model.Key or stsr, not {}".format(
```
luziferius: typo: `stsr`
But i can fix that on my side, no problem.

lib/autokey/scripting/engine.py
```diff
@@ -122,12 +168,90 @@ def create_phrase(self, folder, name: str, contents: str,
...
+            fail=False
+            if isinstance(hotkey, tuple):
+                if len(hotkey) == 2:
+                    # First check modifiers is list of str or keys
+                    if isinstance(hotkey[0], list):
+                        for item in hotkey[0]:
+                            if not isinstance(item, str) and not \
+                            isinstance(item, model.Key):
+                                fail=True
+                    elif not isinstance(hotkey[0], str) and not \
```
luziferius: Just about the readability:
Personally, these instance checks are incredibly hard to parse for me. I don’t know if it’s just me, but it took me over two minutes to think about these lines and check if this is actually correct. I solved it by applying De-Morgan rule which made it immediately obvious.
I think `not(isinstance(item, str) or isinstance(item, model.Key))` is way easier to read, as "is not one of two possible types".

So I think I’ll let my IDE apply De-Morgan rule here.

lib/autokey/scripting/engine.py
```diff
@@ -52,14 +52,60 @@ def get_folder(self, title: str):
                 return folder
         return None
 
+    def create_folder(self, title: str, parent_folder=None, temporary=False):
```
luziferius, line 58: I just noticed:
You can create AutoKey folders at arbitrary paths using the GUI. So you should somehow be able to set the actual file system path. Maybe allowing `parentFolder` to be a PathLike. If it is a `str` or `pathlib.Path`, check the file system existance, try to create the directory path (recursively! `mkdir(parent=True)`) and set the path explicitly.

Not setting the path is safe though, as it will use the default location in this case.
https://github.com/autokey/autokey/blob/cd5746e7337af4895b9c7b032aa5c15473cd5f9b/lib/autokey/model.py#L438-L445

josephj11: I use a folder (sub-tree) outside of the AutoKey config for all my phrases and scripts. It makes them immune to any package modifications. This works even better now that you added code to prevent deletion of existing directories when adding a folder inside AutoKey. This has been a feature for as long as I can remember.

luziferius: I noticed that specifying the storage location is the last thing missing to make the new  `create_folder() `API function in the PR feature-complete with the GUI.



@josephj11 And unrelated to this PR: The package manager must not interfere with stuff in $HOME, as it can never know if you use self-installed applications in `~/.local` or similar. So it is always immune to the package manager regardless of location.

lib/autokey/scripting/engine.py
BlueDrink9: Yeah, I struggled with this one. Also would be happy turning these into functions, eg for validating parent folders for create_folder.
There's a test made, but the hotkey one isn't totally comprehensive I don't think

BlueDrink9: Yeah, I struggled with this one. Also would be happy turning these into functions, eg for validating parent folders for create_folder.
There's a test made, but the hotkey one isn't totally comprehensive I don't think

lib/autokey/scripting/engine.py
BlueDrink9: Oh yeah! I realised that at some point, then forgot. Will fix

BlueDrink9: Done

BlueDrink9:
Could use another review of the folder from path changes. Should be ready to merge now, I think.

lib/autokey/scripting/engine.py
```diff
@@ -47,19 +49,91 @@ def get_folder(self, title: str):
...
+        The 'temporary' property is not touched to avoid deleting an existing
+        folder.
+        Note that if more than one folder has the same title, only the first match will be
+        returned.
+        """
+        validateType(title, "title", str)
+        validateType(parent_folder, "parent_folder",
+                [model.Folder, pathlib.Path])
+        validateType(temporary, "temporary", bool)
+        # XXX Doesn't check if a folder already exists at this path in autokey.
```
luziferius, line 86: I thought a bit about this existence check (This is not a request to actually implement this, I’m just brainstorming a bit):

A simple approach (not 100% foolproof) is checking if the path is a descendant of any existing top level path. To also catch symbolic links, both the given path and all of the autokey top level folder paths have to be resolved and made absolute. If the given path is a descendant of any AK folder, add all directory levels in between in case they are missing, then add the requested folder.

But the above still does not catch hard links (It’s not possible on Linux to create hard directory links) and `bind` mounts. The latter either requires parsing `/etc/mtab` (which may contain wrong data…) or using another approach:

Take the absolute, given path and iterate to the root. At each step, check:
- Does the current path contain an AutoKey json file. If so, stop the iteration.
- Then check if the found path is actually registered in AutoKey (you may have found a backup or smth). See above. This step still has issues with bind mounts, unless you start being tricky: Iterate further up until you end finding AK json files (so you found the top level), saving the Folder names on the way in a list. Use that list to look up the path inside the AutoKey data structure. Beware the "duplicate names permittet" issue. To solve that, also examine the numerical suffixes.
- If the given path is multiple levels below the found path, add all in-between as descendants to AutoKey, Then add the given path.

lib/autokey/scripting/engine.py
```diff
@@ -122,12 +196,30 @@ def create_phrase(self, folder, name: str, contents: str,
...
           internals of AutoKey. Ignore it, if you don’t need it or don’t know what to do with it.
           It can be used for _really_ advanced use cases, where further customizations are desired. Use at your own
           risk. No guarantees are made about the object’s structure. Read the AutoKey source code for details.
         """
+        # Start with input type-checking.
+        validateType(folder, "folder", model.Folder)
+        # For when we allow pathlib.Path
+        # validateType(folder, "folder",
+        #         [model.Folder, pathlib.Path])
+        validateType(name, "name", str)
```
luziferius, line 214: looks nicer this way!

BlueDrink9:
I made a mistake with the hotkey validation, which I fixed in my latest commit on my `develop` branch (cb16d61485642724e33533e240eb0d965c215943). Feel like cherry-picking it?

BlueDrink9:
Anyway, thanks for the merge!

BlueDrink9:
Note to self: Change validation errors to TypeError. [ValueError is for something different](https://airbrake.io/blog/python-exception-handling/class-hierarchy)

PR #335, opened 2019-12-10
Fix #212

### LINKED ISSUE #212: Fatal error if user enters an invalid regex
## Classification:

Crash / locks up

## Reproducibility:

Always

## Summary

We don't seem to do any validation on the user's input for regex matching. The app then doesn't gracefully handle a regex string which Python `re` doesn't parse.

1. Enter a malformed regex for a phrase, like the single character, `*`.
2. Attempt to save this change.
3. Try to trigger the phrase.

## Expected Results

**Good:** The app shouldn't crash. 
**Better:** It should trap and output `re` errors somewhere. 
**Best:** the app should validate the user input by attempting an `re` compile when a user provides a regex. If it fails, then the regex won't be saved.

## Actual Results

The app progressively locks up, and the UI needs to be killed. The server then crashes.

## Version

AutoKey version: **0.95.4**

Used GUI (Gtk, Qt, or both): **Gtk** 

Distro: **Fedora**

## Notes

I think this would be a good first PR for me to work on.

BlueDrink9:
Absolutely still a work in progress, but for the time being I've silenced the crash. It just won't save the regex if it's invalid.

I don't know enough about the GUI to handle warning the user about the error.


BlueDrink9:
Closed because it was based off wrong branch

PR #336, opened 2019-12-10
Fix #212 crash on invalid regex

### LINKED ISSUE #212: Fatal error if user enters an invalid regex

BlueDrink9:
Absolutely still a work in progress, but for the time being I've silenced the crash. It just won't save the regex if it's invalid.

I don't know enough about the GUI to handle warning the user about the error.

### DIFF
--- a/lib/autokey/gtkui/dialogs.py
+++ b/lib/autokey/gtkui/dialogs.py
@@ -36,6 +36,8 @@ from ..iomediator.key import Key
 # from . import configwindow
 from .configwindow0 import get_ui
 
+logger = __import__("autokey.logger").logger.get_logger(__name__)
+
 WORD_CHAR_OPTIONS = {
                      "All non-word": model.DEFAULT_WORDCHAR_REGEX,
                      "Space and Enter": r"[^ \n]",
@@ -532,13 +534,16 @@ class WindowFilterSettingsDialog(DialogBase):
             self.triggerRegexEntry.set_text(item.get_filter_regex())
             self.recursiveButton.set_active(item.isRecursive)
 
+    # It seriously bugs me how much duplication there is between the two GUIs.
     def save(self, item):
+        regex = self.get_filter_text()
         try:
-            item.set_window_titles(self.get_filter_text())
-        except re.error as e:
-            # TODO warn the user somehow. Currently just ignores the regex
-            # instead of saving it.
-            pass
+            item.set_window_titles(regex)
+        except re.error:
+            logger.error(
+                    "Invalid window filter regex: '{}'. \
+                            Discarding without saving.".format(regex)
+                            )
         item.set_filter_recursive(self.get_is_recursive())
 
     def reset(self):
--- a/lib/autokey/gtkui/dialogs.py
+++ b/lib/autokey/gtkui/dialogs.py
@@ -64,28 +64,28 @@ def validate(expression, message, widget, parent):
 
 class DialogBase:
 
-    def __init__(self):    
+    def __init__(self):
         self.connect("close", self.on_close)
-        self.connect("delete_event", self.on_close)        
+        self.connect("delete_event", self.on_close)
 
     def on_close(self, widget, data=None):
         self.hide()
-        return True    
+        return True
 
     def on_cancel(self, widget, data=None):
         self.load(self.targetItem)
         self.ui.response(Gtk.ResponseType.CANCEL)
         self.hide()
-        
+
     def on_ok(self, widget, data=None):
         if self.valid():
             self.response(Gtk.ResponseType.OK)
             self.hide()
-        
+
     def __getattr__(self, attr):
         # Magic fudge to allow us to pretend to be the ui class we encapsulate
         return getattr(self.ui, attr)
-    
+
     def on_response(self, widget, responseId):
         self.closure(responseId)
         if responseId < 0:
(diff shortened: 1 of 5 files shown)


lib/autokey/model.py
```diff
@@ -288,19 +288,25 @@ def get_serializable(self):
...
-            self.isRecursive = data["isRecursive"]
-        else:
-            self.set_window_titles(data)
+        try:
+            if isinstance(data, dict): # check needed for data from versions < 0.80.4
+                self.set_window_titles(data["regex"])
+                self.isRecursive = data["isRecursive"]
+            else:
+                self.set_window_titles(data)
+        except re.error as e:
```
luziferius, line 295: This try block looks like it does nothing?
```python
except re.error as e:
    raise e
```

I think it should silence the error for now like in the dialogue cases. We can add a warning dialog later explaining to the user that it was reset because the saved data was invalid.

lib/autokey/qtui/dialogs/windowfiltersettings.py
```diff
@@ -48,7 +48,11 @@ def load(self, item: model.Item):
             self.apply_recursive_check_box.setChecked(item.isRecursive)
 
     def save(self, item):
-        item.set_window_titles(self.get_filter_text())
+        try:
+            item.set_window_titles(self.get_filter_text())
+        except re.error:
+            # TODO: Warn user. Currently just doesn't save regex on error.
+            pass
```
luziferius: I thought a bit about it. For now, just logging this as a warning in all occurrences should be sufficient.

BlueDrink9:
What's our logging system?

lib/autokey/model.py
BlueDrink9: Yeah, I figure it can't deserialise an erroneous regex if it can't make them in the first place

luziferius: But the user can quit AutoKey and then modify the expression in the settings file and break it.
But AutoKey is generally not hardened against this kind of data manipulation. I just mentioned it because you added the try block to the deserialiser.

(Hardening against erroneous save file manipulation can be done in the future, as that user behaviour looks like not being the main source of crashes and issues. I don’t consider it to be a top priority at the moment.)

luziferius:
AutoKey uses the standard library `logging` module. But I forgot that it is not that well used in AutoKey. 
For the Qt GUI, you can grab a `logger` instance from `autokey.qtui.common`. I looked around and it seems that the GTK GUI does no logging at all. So you can disregard my previous comment about logging it.
Just a `TODO: Add logging` should be good enough. I’ll refactor the logging system used by AutoKey in the future, as that’s something on my roadmap anyways.

To track the logging, I opened #338

lib/autokey/model.py
BlueDrink9: I recognised that, but wrote it off since if you're manipulating files like that you're prepared for the consequences. It's 'unsupported' behavior.

luziferius:
Can we get this in? I think catching the exception, so that it does not crash is ok for now. I can then cherry-pick it into `master` and wrap the lasted fixes up into `0.95.10`

BlueDrink9:
I'll take a look. Does the new logging system have any automated integration with the guis?

luziferius:
Nope. It currently works per module.

So just create a logger instance at the top (https://github.com/autokey/autokey/issues/338#issuecomment-568531697), if there isn't already one and you are good to go.

BlueDrink9:
Ok. Might it be worth looking at adding a handler that alerts via gui?

luziferius:
Definitely. In the Qt abbreviation configuration dialog, the OK button only enables, if the content is actually valid. When the content is invalid, only Cancel is available. This removed the need for an error message box that was shown there previously.

Something like disabling OK plus marking the invalid RE in red plus showing a description label  should be good.

BlueDrink9:
Added logger.

Also, I have a local commit that moves the contents of the 'save' functions (which are 100% shared between the two UI classes) to a separate file. I haven't pushed it to make it easier for you to merge this, but afterwards I'll need help writing the tests for them.

lib/autokey/gtkui/dialogs.py
```diff
@@ -532,13 +534,16 @@ def load(self, item):
...
     def save(self, item):
+        regex = self.get_filter_text()
         try:
-            item.set_window_titles(self.get_filter_text())
-        except re.error as e:
-            # TODO warn the user somehow. Currently just ignores the regex
-            # instead of saving it.
-            pass
+            item.set_window_titles(regex)
+        except re.error:
```
luziferius, line 542: Imho sufficient for now. :+1:

BlueDrink9: Cool, so can be merged and cherried?

luziferius:
Merged and cherry-picked the crash mitigation into the master branch.

Issue #330, opened 2019-12-05
Document documentation


BlueDrink9:
Could you please add info on how to do things like generating documentation, and any other relevant contributor information, to `CONTRIBUTORS.rst`?

luziferius:
The documentation at https://autokey.github.io/? It was generated using `Epydoc` before I joined and was never really updated (At least that’s what I think and the site tells me).

I did some "search&replace" style fixes in the copy that is included in the main repository when I renamed stuff, but I think it should be re-generated completely.

I’m not quite sure how it was done in the past, but I think regenerating it using that tool and then doing a PR against https://github.com/autokey/autokey.github.io should get the online version updated.

BlueDrink9:
Cool. I see there is `extractDocs.py`, but running it doesn't work. I assume the html in `doc/scripting` is the generated doc.

luziferius:
> Cool. I see there is `extractDocs.py`, but running it doesn't work.

I think `extractDocs.py` has to be removed or deleted. It is 8 years old and did not get ported to Python 3 alongside the main application. It was used to create the autocompletion file for the Qt GUI.

> I assume the html in `doc/scripting` is the generated doc.

Yes it is. But it was not generated by `extractDocs.py`. Honestly, I’m not sure how the original dev generated it using `Epydoc`.

BlueDrink9:
Epydoc doesn't support python 3. Looking at `engine.py`, javadoc style is used. So I think I'll just pick a replacement, and document how to run that. The resultant website will look different, sure, but should have the same content.

The alternative is keeping a py2 installation around to run epydoc.

Thoughts?

BlueDrink9:
Epydoc doesn't support python 3. Looking at `engine.py`, javadoc style is used. So I think I'll just pick a replacement, and document how to run that. The resultant website will look different, sure, but should have the same content.
Note, this may require converting the docstrings to some other format, like one [here, which are suggested as more recommended for python](https://stackoverflow.com/questions/5334531/using-javadoc-for-python-documentation)

The alternative is keeping a py2 installation around to run epydoc.

Thoughts?

BlueDrink9:
Epydoc doesn't support python 3. Looking at `engine.py`, javadoc style is used. So I think I'll just pick a replacement, and document how to run that. The resultant website will look different, sure, but should have the same content.
[...]

PR #339, opened 2019-12-12
Allow running arbitrary scripts via `autokey-run` and phrase macros

BlueDrink9:
If the script argument given to `autokey-run` or a phrase macro like `<script name>` is an absolute path (with `~` expanded to `$HOME`), the script at that path will be run as if an autokey script, instead of searching for a script with that description.

This allows scripts to be run without their containing folder being manually added in the gui.

Also updates macro argument parsing, to allow quoted arguments with spaces in them (something which the wiki currently notes as a known limitation.)

Also makes a number of fixes to validation of hotkey creation. Sorry about not extracting this, it was a bit on-the-fly when I was testing a script.

### DIFF
--- a/lib/autokey/macro.py
+++ b/lib/autokey/macro.py
@@ -31,20 +31,20 @@ else:
 
 
 class MacroManager:
-    
+
     def __init__(self, engine):
         self.macros = []
-        
+
         self.macros.append(ScriptMacro(engine))
         self.macros.append(DateMacro())
         self.macros.append(FileContentsMacro())
         self.macros.append(CursorMacro())
-        
+
     def get_menu(self, callback, menu=None):
         if common.USING_QT:
             for macro in self.macros:
                 menu.addAction(MacroAction(menu, macro, callback))
-        
+
         else:
             menu = Gtk.Menu()
 
@@ -56,17 +56,30 @@ class MacroManager:
             menu.show_all()
 
         return menu
-        
+
     def process_expansion(self, expansion):
         parts = KEY_SPLIT_RE.split(expansion.string)
-        
-        for macro in self.macros:        
+
+        for macro in self.macros:
             macro.process(parts)
-        
+
         expansion.string = ''.join(parts)
-        
+
 
 class AbstractMacro:
+
+    @property
+    @abstractmethod
+    def ID(self):
+        pass
+    @property
+    @abstractmethod
+    def TITLE(self):
+        pass
+    @property
+    @abstractmethod
+    def ARGS(self):
+        pass
 
     def get_token(self):
         ret = "<%s" % self.ID
@@ -74,17 +87,18 @@ class AbstractMacro:
         ret += "".join((" " + k + "=" for k, v in self.ARGS))
         ret += ">"
         return ret
-            
+
     def _can_process(self, token):
         if KEY_SPLIT_RE.match(token):
             return token[1:-1].split(' ', 1)[0] == self.ID
         else:
             return False
-        
+
     def _get_args(self, token):
+        print(token)
         l = token[:-1].split(' ')
         ret = {}
-                
+
         if len(l) > 1:
             for arg in l[1:]:
                 key, val = arg.split('=', 1)
@@ -93,10 +107,11 @@ class AbstractMacro:
         for k, v in self.ARGS:
             if k not in ret:
                 raise Exception("Missing mandatory argument '{}' for macro '{}'".format(k, self.ID))
-        
+
         return ret
-        
+
     def process(self, parts):
+        print("parts ", parts)
         for i in range(len(parts)):
             if self._can_process(parts[i]):
                 self.do_process(parts, i)
@@ -111,7 +126,7 @@ class CursorMacro(AbstractMacro):
     ID = "cursor"
     TITLE = _("Position cursor")
     ARGS = []
-    
+
     def do_process(self, parts, i):
         try:
             lefts = len(''.join(parts[i+1:]))
@@ -119,18 +134,18 @@ class CursorMacro(AbstractMacro):
             parts[i] = ''
         except IndexError:
             pass
-                        
-    
+
+
 class ScriptMacro(AbstractMacro):
 
     ID = "script"
     TITLE = _("Run script")
     ARGS = [("name", _("Name")),
             ("args", _("Arguments (comma separated)"))]
-    
+
     def __init__(self, engine):
         self.engine = engine
-    
+
     def do_process(self, parts, i):
         args = self._get_args(parts[i])
         self.engine.run_script_from_macro(args)
@@ -142,7 +157,7 @@ class DateMacro(AbstractMacro):
     ID = "date"
     TITLE = _("Insert date")
     ARGS = [("format", _("Format"))]
-    
+
     def do_process(self, parts, i):
         format_ = self._get_args(parts[i])["format"]
         date = datetime.datetime.now().strftime(format_)
@@ -154,9 +169,9 @@ class FileContentsMacro(AbstractMacro):
     ID = "file"
     TITLE = _("Insert file contents")
     ARGS = [("name", _("File name"))]
-    
+
     def do_process(self, parts, i):
         name = self._get_args(parts[i])["name"]
-        
+
         with open(name, "r") as inputFile:
             parts[i] = inputFile.read()
--- a/lib/autokey/macro.py
+++ b/lib/autokey/macro.py
@@ -1,5 +1,6 @@
 import datetime
 from abc import abstractmethod
+import shlex
 
 from autokey.iomediator.constants import KEY_SPLIT_RE
 from autokey.iomediator.key import Key
@@ -29,6 +30,21 @@ if common.USING_QT:
 else:
     from gi.repository import Gtk
 
+
+def extract_tag(s):
+    if not isinstance(s, str):
+        raise TypeError
+    extracted = [p.split('>')[0] for p in s.split('<') if '>' in p]
+    if len(extracted) == 0:
+        return s
+    else:
+        return ''.join(extracted)
+
+def split_key_val(s):
+    # Split as if a shell argument.
+    # Splits at spaces, but preserves spaces within quotes.
+    pairs = shlex.split(s)
+    return dict(pair.split('=', 1) for pair in pairs)
 
 class MacroManager:
 
@@ -95,23 +111,23 @@ class AbstractMacro:
             return False
 
     def _get_args(self, token):
-        print(token)
-        l = token[:-1].split(' ')
-        ret = {}
+        macro = extract_tag(token)
+        macro_type, macro = macro.split(' ', 1)
+        args = split_key_val(macro)
+        expected_args = [arg[0] for arg in self.ARGS]
+        expected_argnum = len(self.ARGS)
 
-        if len(l) > 1:
-            for arg in l[1:]:
-                key, val = arg.split('=', 1)
-                ret[key] = val
-
-        for k, v in self.ARGS:
-            if k not in ret:
-                raise Exception("Missing mandatory argument '{}' for macro '{}'".format(k, self.ID))
+        for arg in expected_args:
+            if arg not in args:
+                raise ValueError("Missing mandatory argument '{}' for macro '{}'".format(arg, self.ID))
+        for arg in args:
+            if arg not in expected_args:
+                raise ValueError("Unexpected argument '{}' for macro '{}'".format(k, self.ID))
+        return args
 
         return ret
 
     def process(self, parts):
-        print("parts ", parts)
         for i in range(len(parts)):
             if self._can_process(parts[i]):
                 self.do_process(parts, i)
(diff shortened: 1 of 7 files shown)


lib/autokey/macro.py
```diff
@@ -95,23 +111,23 @@ def _can_process(self, token):
...
-        ret = {}
-
-        if len(l) > 1:
-            for arg in l[1:]:
-                key, val = arg.split('=', 1)
-                ret[key] = val
-
-        for k, v in self.ARGS:
-            if k not in ret:
-                raise Exception("Missing mandatory argument '{}' for macro '{}'".format(k, self.ID))
```
luziferius, line 95: Since this PR is against `develop`, changing the Exception type to ValueError is ok. This is something I wanted to do anyways, so it’s nice to see it done.

lib/autokey/macro.py
```diff
@@ -30,6 +31,21 @@ def on_triggered(self):
...
+    extracted = [p.split('>')[0] for p in s.split('<') if '>' in p]
+    if len(extracted) == 0:
+        return s
+    else:
+        return ''.join(extracted)
+
+def split_key_val(s):
+    # Split as if a shell argument.
+    # Splits at spaces, but preserves spaces within quotes.
+    pairs = shlex.split(s)
```
luziferius, line 46: Using shlex is an interesting idea.
When I started to work on this, I threw the biggest hammer on the problem I knew by the time, by building a full-fledged RE-based tokenizer and parser. My goal was to have a fully bug- and corner-case free parser. Unfortunately, there was no real end result. Just for reference, that’s the diff of my incomplete and stale attempt: https://github.com/autokey/autokey/compare/develop...luziferius:improve_phrase_macro_parsing

So let’s see how your much shorter solution copes with various corner cases.

tests/test_macro.py
```diff
@@ -13,13 +13,53 @@
...
+    # first-run logic.
+    with patch("autokey.model.Phrase.persist"), patch("autokey.model.Folder.persist"),\
+         patch("autokey.configmanager.configmanager.ConfigManager.load_global_config",
+               new=(lambda self: self.folders.append(test_folder))):
+        engine = Engine(ConfigManager(MagicMock()), MagicMock(spec=PhraseRunner))
+        engine.configManager.config_altered(False)
+
+    return engine, test_folder
+
+def test_arg_parse():
```
luziferius, line 48: With this kind of structural regularity in the test inputs, you can parametrize the test and save the typing/copy&pasting the same assert statement over and over again.
just add 3 arguments to the test case function like `test_input, expected, error_msg` and put
the `@pytest.mark.parametrize` decorator in the front of the definition.
Like
```python
@pytest.mark.parametrize("test_input, expected, error_msg", [
    ("<script name='test name' args='long arg with spaces and ='>", {"name": "test name", "args": "long arg with spaces and ="}, "Macro arg can't contain equals"),
…
])
def test_arg_parse(test_input, expected, error_msg):
    engine, folder = create_engine()
    macro = ScriptMacro(engine)
    assert_that(macro._get_args(test_input), is_(equal_to(expected)),
                error_msg)
```
Then adding a new case is just adding a new tuple with test data.

lib/autokey/service.py
```diff
@@ -532,4 +532,5 @@ def run_subscript(self, script):
 
     def run_subscript_path(self, path):
         scope = self.scope.copy()
+        scope["__file__"] = script.path
```
luziferius: Oh yeah, Seems I forgot that location. Thanks for fixing it

lib/autokey/service.py
```diff
@@ -491,6 +498,20 @@ def execute(self, script: model.Script, buffer=''):
...
 
+    @threaded
+    def execute_path(self, path: pathlib.Path, buffer=''):
+        logger.debug("Script runner executing: {}".format(path))
+        scope = self.scope.copy()
+        # scope["store"] = script.store
+        # Overwrite __file__ to contain the path to the user script instead of the path to this service.py file.
+        scope["__file__"] = path.resolve()
+        try:
+            exec(open(path).read(), scope)
```
luziferius: This `open(path).read()` leaks an opened file handle, which is never closed.
`path.read_text()` automatically closes the file after reading the content, as indicated by
https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text

lib/autokey/macro.py
BlueDrink9: Can you tell me what sorts of corner cases you were referring to? Then I can add tests, at least.

BlueDrink9: Sorry, I hadn't realised you had already put so much work in. I just came across needing spaces in args, and assumed there'd be a library to parse strings for arguments.

BlueDrink9: Yours looks like it will probably be more sophisticated. What stage is it at? May be worth replacing my quick change with yours when it is done. Mine doesn't handle nested macros, and possibly never can.

tests/test_macro.py
BlueDrink9: Thank you! I saw that in one of the other test files, but didn't understand how it was used.

lib/autokey/service.py
BlueDrink9: Interesting! I note that it was only added in 3.5. I don't know what you consider a minimal supported python 3 to be.

I'll change it. Teaches me to blindly copy the SO solution.

BlueDrink9: Oh, actually it wasn't SO, it was an equivalent function somewhere else in autokey. So I'll change it twice.

lib/autokey/macro.py
luziferius: > Sorry, I hadn't realised you had already put so much work in. I just came across needing spaces in args, and assumed there'd be a library to parse strings for arguments.

No Problem! That branch lives inside my personal fork (see the diff link) and I consider it stale at the moment. It is incomplete. There’s no way I’d assume you look around my stuff to look for bits and pieces. So don’t worry about it.


> Can you tell me what sorts of corner cases you were referring to? Then I can add tests, at least.

Let’s see if I can still pull some from under my hat. I had mostly nesting broken macro tokens and similar stuff on my mind.


- `<script name="<script name="foo", args="None">", args="None">` This nests two valid macro tokens, if you consider `<script name="foo", args="None">` to be a valid script name (which it is!). Which one to execute is to decide…
- `<script <script name="myscript" args="valid_token">>`
- `<script name=<file <date format="%Y \"> some text <script (broken) <script name="myscript" args="valid_token">> args="<script name=`

BlueDrink9: Ofc I wouldn't look at your bits, but neovim for example requires PRs discussing what you plan to do, to prevent duplication of work/unwanted work. At the least, we could set up some sort of communication system for what we are working on.

BlueDrink9: Ah, they look quite involved. The second one currently isn't valid because the outer macro requires 'args', still.

TBH my method still doesn't handle nested macros anyway, let alone broken ones. I can look at implementing it, but it's lower priority unless you make it a requirement for merge.

luziferius: > Ah, they look quite involved. The second one currently isn't valid because the outer macro requires 'args', still.

In the second point, note that the outer `<script>`-thing has no `name` parameter. Here, the inner token is valid.

> TBH my method still doesn't handle nested macros anyway, let alone broken ones. I can look at implementing it, but it's lower priority unless you make it a requirement for merge.

That won’t be a requirement, so it’s fine as it is.

tests/test_macro.py
luziferius: In those, I went a step further. You can pass any iterable, like a Generator function instead of a list into `@pytest.mark.parametrize()`.

So I wrote functions named like `generate_test_cases_for_<insert_test_name_here>()` and placed them in front of the test function. Then I used some nested helper functions inside the generator functions to cut down typing redundant bits and used those to `yield` one test case per line.

This is quite advanced but makes it super easy to add additional test cases to existing tests.

lib/autokey/service.py
luziferius: > Oh, actually it wasn't SO, it was an equivalent function somewhere else in autokey. So I'll change it twice.

Huh? OK then, TIL. Go ahead ;)

> Interesting! I note that it was only added in 3.5. I don't know what you consider a minimal supported python 3 to be.

Huh? (The second…) The Readme used to be stating that Python 3.5 is the minimal supported Python version.

BlueDrink9: Oops, sorry. Should have checked the readme.

I'll see if I can find that function, I might me misremembering.

luziferius: No no no. Sorry from my side. It seems that this bit vanished when the dependency information was de-duplicated. There used to be a slightly outdated section in the README and in the Wiki. Both got merged into the Wiki. But it seems that the minimal Python version was not properly carried to the Wiki page.

Edit: I’ve updated the section, here: https://github.com/autokey/autokey/wiki/Installing#Dependencies

BlueDrink9: I take it back, it wasn't already there. Normally there was no need to read the code in a script, I added it as part of arbitrary scripts. But I think there are two instances, I'll change both.

Statement about SO stands.

BlueDrink9: Also, regarding the wiki: As I make changes, they change relevancy bits of wiki documentation. But it seems the wiki can only apply to the current latest release. Any suggestion for storing up notes on what to update in the wiki? Maybe a file somewhere with notes

luziferius: Oh, And about:

> Teaches me to blindly copy the SO solution.

You are sure that the license of the copied SO solution was GPLv3 compatible? :smiling_imp: 

(This, plus possible issues in the actual code are the two major obstacles when looking up stuff there…)



> Also, regarding the wiki: As I make changes, they change relevancy bits of wiki documentation. […]

I think, handling this with some additional notes at the relevant bits may be sufficient. For example, noting at which version functions were added and changed. For example, somewhat like this: https://documen.tician.de/pyopencl/runtime_queue.html#pyopencl.Event.from_int_ptr

That documentation has statements like these, to indicate changes:

> New in version 2013.2.
> 
> Changed in version 2016.1: retain added.

BlueDrink9: Ok. So if I make a change in my branch, I hope it will eventually become a PR, which will eventually be merged into dev, which will become `0.96`. What stage do I say in the wiki "Changed in `0.96`"?

tests/test_macro.py
BlueDrink9: I'll push this change to `macro` branch, pull #340

lib/autokey/service.py
luziferius: I’d say some time prior to releasing the next version. The changelog has to be written anyways, so that time frame looks like it may be a good idea. I’m not totally sure either.
The best way may be actually placing it into the docstrings and automatically generate the relevant pages. (Related to #330)

BlueDrink9: I wholeheartedly agree with the last part, but for macros that could be problematic because they aren't associated with a particular function to create a docstring. Unless we
just chuck the docstring in `macro.py`

tests/test_macro.py
BlueDrink9: Any suggestions for how to parametrize `test_engine.py`'s `create_phrase`? Since it needs a `Folder` object created, but doesn't always use it?

lib/autokey/service.py
BlueDrink9: This has been fixed.

tests/test_macro.py
luziferius: Partition the test cases by the used parameters and the expected result (positive/success, negative/raisedException). Each partition of the tests gets it’s own function. Then parameters can be used to raise the number of tests by testing multiple valid/invalid inputs.

BlueDrink9: But how do you pass an object arg instead of a string without making it a global arg?

BlueDrink9: But how do you pass an object arg instead of a string without making it a global arg?

BlueDrink9:
I think there is still a bug with running arbitrary scripts from phrase macros. Once I fix it, this will be ready to merge.

BlueDrink9:
I've done some manual checks, and I think this is ready to be merged unless there's more you want to do. Merge before #340.

josephj11:
We're going to need some documentation on how to use this when it is released. Sounds pretty powerful.

BlueDrink9:
I agree Joseph. The problem is the wiki isn't version controlled, which is annoying. But I've in-script documented the api function, at least.

tests/test_macro.py
luziferius: > But how do you pass an object arg instead of a string without making it a global arg?

Sorry, it seems that I missed your last question.

Normally, you can pass anything as an argument to the test function.
But If you create class objects of any AutoKey class inside the test parameter list, any Exceptions caused during the object creation won’t be associated to the failing test case. 

Instead, those Exceptions will be risen during the test collection phase (because the test case generators are run beforehand). And Exceptions during the collection phase cause more things to fail.

I’m not quite sure, but it may even kill the whole test suite with an "Errors occurred during the collection phase." error. At least, it fails the whole test.

If we start to use `pyfakefs` to test file system interactions, the `Path` instances must be created inside the `test_function()`, otherwise it will point to the outside world. There may be similar issues with other packages.

That’s why I’m restricting myself to use built-in types or simple structs of built-in types only.

PR #340, opened 2019-12-15
Add unit tests for macros

BlueDrink9:
One test is currently failing, but it requires a non-mocked phraserunner to pass - which seems to need a non-mocked app.

I re-factored a lot of the code for macro processing to make the macro code easier to understand and test.

Branched off my previous PR, and I'm not sure quite how to handle that.

### DIFF
--- a/lib/autokey/service.py
+++ b/lib/autokey/service.py
@@ -507,7 +507,7 @@ class ScriptRunner:
         # Overwrite __file__ to contain the path to the user script instead of the path to this service.py file.
         scope["__file__"] = path.resolve()
         try:
-            exec(pathlib.Path(path).read_text(), scope)
+            exec(path.read_text(), scope)
         except Exception as e:
             logger.exception("Script error")
             self.error = "Script name: '{}'\n{}".format(path, traceback.format_exc())
@@ -533,5 +533,5 @@ class ScriptRunner:
 
     def run_subscript_path(self, path):
         scope = self.scope.copy()
-        scope["__file__"] = script.path
+        scope["__file__"] = path
         exec(pathlib.Path(path).read_text(), scope)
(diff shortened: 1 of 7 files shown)


tests/test_macro.py
```diff
@@ -0,0 +1,153 @@
...
+    expected="test result macro expansion\n two macros this time test result macro expansion\n".format(path, path)
+    assert_that(expandMacro(engine, contents), is_(equal_to(expected)),
+                "Two macros per phrase don't expand properly")
+    contents="<file name={}> mixed macro <cursor> types".format(path)
+    expected="test result macro expansion\n mixed macro  types<left><left><left><left><left><left>"
+    assert_that(expandMacro(engine, contents), is_(equal_to(expected)),
+                "mixed macros don't expand properly")
+
+# def test_nested_macro_raises_error():
+#     contents="<date format=<cursor>>"
```
luziferius, line 224: That may be a valid use case, though. Either having a literal `<cursor>` in your formatted date or expecting the cursor to be placed at the indicated position inside the formatted string.
Either way, this can be an error for now.

lib/autokey/service.py
```diff
@@ -507,7 +507,7 @@ def execute_path(self, path: pathlib.Path, buffer=''):
         # Overwrite __file__ to contain the path to the user script instead of the path to this service.py file.
         scope["__file__"] = path.resolve()
         try:
-            exec(open(path).read(), scope)
+            exec(pathlib.Path(path).read_text(), scope)
```
luziferius: Isn’t `path` already a Path instance? According to line 508 (two above), it must be. So `exec(path.read_text(), scope)` should be sufficient here.
This isn’t a direct error, it just seems unnecessary.

BlueDrink9: Oops, thanks

BlueDrink9: Fixed

BlueDrink9:
I've done some manual checks, and I think this is ready to be merged after #339.

PR #344, opened 2019-12-22
Close #246: Allow engine.run_script to accept keyword arguments

### LINKED ISSUE #246: Add API functions to call AutoKey scripts as functions
## Classification:

Enhancement to the scripting API

## Version

AutoKey version: All < 0.96.0

## Summary

Add functionality to call scripts like functions.

## Proposal:

- Add further optional arguments (`*args, **kwargs`) to [`engine.run_script(script_name)`](https://autokey.github.io/lib.scripting.Engine-class.html#run_script)
- Arguments can be accessed using a getter function. Something like [`engine.get_macro_arguments()`](https://autokey.github.io/lib.scripting.Engine-class.html#get_macro_arguments)?
- Return values can be set using the existing `engine.set_return_value()`
- engine.run_script() should return the value set by `engine.set_return_value()`

## Notes
There is a possible workaround that involves using the global scripting store. It takes quite some effort to use, because it is persistent and needs to be cleared after usage to not carry old data.

BlueDrink9:
I wrote a test, but it won't work until #339 is merged, because it uses an absolute path to the test script to make things easier.

### DIFF
--- a/lib/autokey/scripting/engine.py
+++ b/lib/autokey/scripting/engine.py
@@ -378,7 +378,7 @@ Phrases created within temporary folders must themselves be explicitly set tempo
         Usage: C{engine.get_script_arguments()}
 
         @return: the arguments
-        @rtype: C{list(str())}
+        @rtype: C{list[Any]}
         """
         return self._script_args
 
@@ -390,7 +390,7 @@ Phrases created within temporary folders must themselves be explicitly set tempo
         Usage: C{engine.get_script_keyword_arguments()}
 
         @return: the arguments
-        @rtype: C{dict(str():str())}
+        @rtype: C{Dict[str, Any]}
         """
         return self._script_kwargs
 
--- a/lib/autokey/scripting/engine.py
+++ b/lib/autokey/scripting/engine.py
@@ -37,6 +37,8 @@ class Engine:
         self.runner = runner
         self.monitor = config_manager.app.monitor
         self._macro_args = []
+        self._script_args = []
+        self._script_kwargs = {}
         self._return_value = ''
         self._triggered_abbreviation = None  # type: Optional[str]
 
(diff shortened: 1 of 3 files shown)


lib/autokey/scripting/engine.py
```diff
@@ -365,6 +371,29 @@ def run_script_from_macro(self, args):
...
             self.set_return_value("{ERROR: %s}" % str(e))
 
+    def get_script_arguments(self):
+        """
+        Get the arguments supplied to the current script via the scripting api
+
+        Usage: C{engine.get_script_arguments()}
+
+        @return: the arguments
+        @rtype: C{list(str())}
```
luziferius: This is of type `List[Any]`. The user can and should be able to pass anything.

lib/autokey/scripting/engine.py
```diff
@@ -365,6 +371,29 @@ def run_script_from_macro(self, args):
...
+
+    def get_script_keyword_arguments(self):
+        """
+        Get the arguments supplied to the current script via the scripting api
+        as keyword args.
+
+        Usage: C{engine.get_script_keyword_arguments()}
+
+        @return: the arguments
+        @rtype: C{dict(str():str())}
```
luziferius: Same here: `Dict[str, Any]`

josephj11:
This sounds like it may be end user accessible. If it is, we'll need documentation on how to use it.

luziferius:
> This sounds like it may be end user accessible. If it is, we'll need documentation on how to use it.

It is, see #246. This extends the `engine.run_script()` command to pass additional arguments to the executed script. It obsoletes workarounds using the global store to transfer data between scripts.

#339 and #340 are merged, so the blockers are resolved. But I have no idea, why this introduced merge conflicts, as this should merge fine.

One point is missing to implement everything in #246: `engine.run_script()` should actually _return_ the data set via `engine.set_return_value()`. This is a simple one-liner at the end of `run_script`.

lib/autokey/scripting/engine.py
BlueDrink9: Thanks, I misunderstood how the doc worked.

BlueDrink9:
Thomas, one question I have is whether we keep `get_macro_args` as well as `get_script_args`, or whether we depricate the former and just have it call the latter (my
preference)

luziferius:
> Thomas, one question I have is whether we keep `get_macro_args` as well as `get_script_args`, or whether we depricate the former and just have it call the latter (my
> preference)

Since the arguments passed to a script from a Phrase macro are technically still arguments passed to a script, `get_script_args` is more technically accurate than `get_macro_args`. Thus I’m with you, `get_macro_args` can be deprecated.

PR #355, opened 2020-01-15
Bugfix/issue 354

BlueDrink9:


tests/scripting_api/test_system.py
```diff
@@ -0,0 +1,29 @@
+
+# Copyright (C) 2019 Thomas Hess <thomas.hess@udo.edu>
```
luziferius: That was clearly not written by me! You should attribute yourself properly ;)
You can use your nickname, if you don’t want to do it under your real name, I’ve already seen that in many projects.

BlueDrink9: I `:r` from another test file. There's probably another one somewhere in one of the other PRs, but I don't really mind.

luziferius: No problem, I just don’t want to claim ownership for things you did.

PR #356, opened 2020-01-17
Scripting API: Add option to override duplicate hotkeys

BlueDrink9:
Also a whole bunch of refactoring sorry, but mostly related to the changed code.

Reason for wanting this is (mainly) for temporary hotkeys, creating big changes and re-defining what a hotkey does within a script.


lib/autokey/configmanager/configmanager.py
```diff
@@ -669,22 +669,42 @@ def check_hotkey_unique(self, modifiers, hotKey, newFilterPattern, targetItem):
...
+        """
+        Gets first item with the specified hotkey. Also checks the
+        special hotkeys configured from the advanced settings dialog.
+        Checks folders first, then phrases, then special hotkeys.
+
+        @param modifiers: modifiers for the hotkey
+        @param hotKey: the hotkey to check
+        @param newFilterPattern:
+        """
+        for searchSpace in [self.allFolders, self.allItems]:
```
luziferius: You _could_ use `for item in itertools.chain(self.allFolders, self.allItems)` to remove one line of nesting here, if you wish. The chain will simply iterate over the items in both lists.

But this way will work, too.

BlueDrink9: Neat python, thanks!

BlueDrink9: Resolved. Just got one failing test to fix, then I'll push.

BlueDrink9:
@luziferius bump?

lib/autokey/scripting/engine.py
```diff
@@ -132,7 +132,7 @@ def create_phrase(self, folder, name: str, contents: str,
                       hotkey: Tuple[List[Union[model.Key, str]], Union[model.Key, str]]=None,
                       send_mode: model.SendMode=model.SendMode.KEYBOARD, window_filter: str=None,
                       show_in_system_tray: bool=False, always_prompt: bool=False,
-                      temporary=False):
+                      temporary=False, replaceExistingHotkey=False):
```
luziferius: Python PEP8 (code style recommendation) says to not use camelCase for variables and parameters. instead use lower_case_with_underscores. Adding CamelCase exactly _here_ even mixes both styles in a public API function call.

luziferius:
This is quite a complex diff. I’ll continue to review this tomorrow.

lib/autokey/scripting/engine.py
BlueDrink9: Oops, missed that. Sorry

lib/autokey/scripting/engine.py
```diff
@@ -203,45 +203,32 @@ def create_phrase(self, folder, name: str, contents: str,
                                     not persisted as .jsons, and are replaced if the description is not
                                     unique within the folder.
                                      Used for single-source rc-style scripts.
+        @param replaceExistingHotkey: If true, instead of warning if the hotkey
```
luziferius: Just a reminder: Don’t forget to update the doc-string, too

lib/autokey/configmanager/configmanager.py
```diff
@@ -722,6 +742,7 @@ def remove_all_temporary(self, folder=None, in_temp_parent=False):
                 pass
             self.remove_all_temporary(subfolder, in_temp_parent)
 
+    # TODO: merge this function with model.Phrase.unset_hotkey
```
luziferius: This is actually member of `AbstractHotkey`, where you added it.

I think I remember a somewhat similar situation with comments that were dead-on-arrival ;-)

lib/autokey/configmanager/configmanager.py
```diff
@@ -669,22 +670,41 @@ def check_hotkey_unique(self, modifiers, hotKey, newFilterPattern, targetItem):
...
-                    return item is targetItem, item
+                if self.item_has_same_hotkey(item,
+                                             modifiers,
+                                             hotKey,
+                                             newFilterPattern):
+                    return item
+        return None
 
-        return True, None
+    def item_has_same_hotkey(self, item, modifiers, hotKey, newFilterPattern):
```
luziferius: can be a `@staticmethod`, as `self` is unused.

lib/autokey/configmanager/configmanager.py
```diff
@@ -669,22 +670,41 @@ def check_hotkey_unique(self, modifiers, hotKey, newFilterPattern, targetItem):
...
+                                              modifiers,
+                                              hotKey,
+                                              newFilterPattern):
+                return item
 
         for item in self.globalHotkeys:
             if item.enabled:
-                if item.modifiers == modifiers and item.hotKey == hotKey and item.filter_matches(newFilterPattern):
-                    return item is targetItem, item
+                if self.item_has_same_hotkey(item,
```
luziferius: You can just `and` both conditions here and reduce a nesting depth, like you did in the for loop above (line 690, https://github.com/autokey/autokey/pull/356/files#diff-d8b02a3684c92866db39ed969414c71dR689-R690)

lib/autokey/scripting/engine.py
```diff
@@ -483,44 +470,68 @@ def remove_all_temporary(self, folder=None,
         self.configManager.remove_all_temporary(folder,
                 in_temp_parent)
 
+    def validateArguments(self, folder, name, contents,
```
luziferius: Please move `validateArguments` out of the class. It can be static method, so it is safe to move out. That way, it will be inaccessible to user scripts.

lib/autokey/scripting/engine.py
```diff
@@ -483,44 +470,68 @@ def remove_all_temporary(self, folder=None,
...
+        validateType(temporary, "temporary", bool)
+        validateType(replaceExistingHotkey, "replaceExistingHotkey", bool)
+        # TODO: The validation should be done by some controller functions in the model base classes.
+
+        if folder.temporary and not temporary:
+            raise ValueError("Parameter 'temporary' is False, but parent folder is a temporary one. \
+    Phrases created within temporary folders must themselves be explicitly set temporary")
+
+
+    def check_abbreviation_unique(self, abbreviations):
```
luziferius: You have to add the window filter as an argument here. Multiple model items may have the same abbreviation, if both have distinct * window filters. 

(Side note: I’m using this functionality myself for those GPL headers, which I paste using AutoKey Phrases. The Phrases have the same abbreviation, but different window filters. I use it to paste language-specific headers based on the IDE in use.)

https://github.com/autokey/autokey/blob/f0c69ba81fc082030ce35564e1a52102e20d7be8/lib/autokey/configmanager/configmanager.py#L611-L629

So the abbreviation may be considered "unique" by AutoKey, if the window filters are distinct.

### [*] About distinct window filters, (which is unrelated to the PR):

To determine if two window filters, which are a Python regular expression are distinct, one has to check if the intersection of the regular languages given by the REs is empty. I.e. there exists no string that is matched by both REs. (¬∃x.x∈L(RE1)∧x∈L(RE2))

This test is infeasible on REs. But they can be converted to DFAs, for which such a test is trivial. 
(This conversion can be done by converting first both REs to [ε-NFAs](https://en.wikipedia.org/wiki/Nondeterministic_finite_automaton#NFA_with_%CE%B5-moves) using https://en.wikipedia.org/wiki/Thompson%27s_construction and then to [DFAs](https://en.wikipedia.org/wiki/Deterministic_finite_automaton) using https://en.wikipedia.org/wiki/Powerset_construction.)

But the last step ε-NFA → DFA  may grow in O(2^n) size. So this test is equally infeasible. But I’ve heard that the intersection test can be implemented directly on ε-NFAs. So if there is a library to convert Python REs to ε-NFAs (pro-tip: there currently isn’t), the test "are window filters distinct" can actually be implemented.

lib/autokey/scripting/engine.py
```diff
@@ -483,44 +470,68 @@ def remove_all_temporary(self, folder=None,
...
+        if not abbreviations:
+            return
+        for abbr in abbreviations:
+            if not self.configManager.check_abbreviation_unique(abbr, None, None)[0]:
+                raise ValueError("The specified abbreviation '{}' is already in use.".format(abbr))
+
+    def check_hotkey_unique(self, hotkey):
```
luziferius: Same as above.

tests/scripting_api/test_engine.py
```diff
@@ -29,9 +31,14 @@
...
 from autokey.scripting import Engine
 
+# For use in paramerizations.
+# Call replace_folder_in_args(folder, args) to use.
+folder_param="create_new_folder"
+
 def get_autokey_dir():
     return os.path.dirname(os.path.realpath(sys.argv[0]))
 
+@pytest.fixture
```
luziferius, line 41: Ah. _That’s how_ you use the decorator. I’ve never figured out myself. TIL.

lib/autokey/scripting/engine.py
BlueDrink9: Good point. If my refactoring tool wasn't broken (meaning I had to find/replace) I would have missed this

lib/autokey/configmanager/configmanager.py
BlueDrink9: I hadn't even picked that python could have those, d'oh

lib/autokey/configmanager/configmanager.py
BlueDrink9: hahahaha, had to `blame` to check myself.

tests/scripting_api/test_engine.py
BlueDrink9: I fumbled around for a bit until things passed

lib/autokey/scripting/engine.py
BlueDrink9: Wait, so we can't tell if window filters are distinct? Then how was it done before? (I mostly copied the `unique` code from elsewhere in engine, iirc). Was before incorrect? In which case, do you still need me to correct it?

luziferius: > Wait, so we can't tell if window filters are distinct? 

Yes, in the general case.

> Then how was it done before? (I mostly copied the `unique` code from elsewhere in engine, iirc). 

This check, performed on all model class instances:
https://github.com/autokey/autokey/blob/14cbb12acef326b77a4c3964c4ec7172cc2118c5/lib/autokey/model.py#L327-L331

It currently simply checks if the RE patterns are unique. This approach seems to be kind-of sufficient, because normally the REs are written to match literals (WM_CLASS) or simple prefixes or suffixes (window titles). At least, there was never a bug report about it.
You have to try to break it to actually trigger the corner cases here.

For example, the window title filter for the Firefox web browser can be written as `.*Mozilla Firefox`. Testing the uniqueness of such RE patterns are quite successful. It breaks, if you deliberately game the implementation: For example by using a second filter like `.*.*Mozilla Firefox`, which matches the same regular language, but is syntactically different.

> Was before incorrect?

Which “before” do you refer to? To me, that question is ambiguous, so I’ll answer both interpretations I can come up with.
- The uniqueness-check performed by AutoKey is technically incorrect, but works during normal use. Solving this is out of scope for this PR.
- Your code should perform the filter-uniqueness check like the GUI does. 

> In which case, do you still need me to correct it?

Yes, please include the window filter in the uniqueness check. The functionality should match the GUI, even if there are some known-bad corner cases in the validation logic used currently.



Further notes:
The method above that checks the REs using finite automata will solve it for the general case. But this is more like an academic research topic at the moment. Maybe at some point in time, this can be implemented as a standalone library, which is then used by AutoKey. A function that solves the question "Are these 2 Python REs distinct?" may be useful for other projects, too. 
Another caveat: Python REs have some advanced features that cause them to be actually context-free if used. So this problem may be actually unsolvable, in the mathematical sense, as context-free grammars can’t be checked for being intersection-free.

BlueDrink9: Thanks for that explanation. The explanations about DFAs and the theory are interesting too, is that a specific interest of yours?

luziferius: That’s something i learned in university. I happened to start to like the theoretical bits, because I started to grasp that the reasoning the professors gave were true. So that stuff just stuck. And yeah, I kind of like formal languages. It took a bit to understand, but then it stuck ^^

If you know the theoretical bits, you can argue about the feasibility of features in software based on hard maths. Here, window filters are regular languages, so you can take the toolbox for those and use already done proofs to argue about what can be done, how that can be done and what can not be done.

And when I wrote the initial comment, I was in the mood to write down my thoughts, even if it was a bit unrelated. I should put the notes into a separate issue, so that it won’t get lost.

BlueDrink9: Yeah, I get you. That was the main reason I pushed through learning formal language theory. The famous cartoon about proving NP-equivalence comes to mind.

BlueDrink9:
Done. Ready to merge

luziferius:
Will do a last round of review tomorrow and merge then.

lib/autokey/model.py
```diff
@@ -343,11 +343,17 @@ def get_filter_regex(self):
             return ""
 
     def filter_matches(self, otherFilter):
+        # XXX Should this be and?
```
luziferius, line 346: I think this is used to check if the trigger of another item, given that it has this `otherFilter` filter, will collide with the item’s own trigger.
So the triggers do collide if the other filter is None(matching everything) _or_ the own filter is None (matching everything). So `or` should be correct.

or, in formula: L(any_filter) ⊆ L(Σ) = L(".*")

- L(X) == Regular Language of RE X, everything matched by X
- Σ == The alphabet / all characters, L(Σ) == all possible strings.

If the filter is `None`, it is equivalent to matching L(Σ) or setting the filter to `.*`

As long as proper filter matching is not implemented, this has one edge case: If one filter is actually not-None, but matching L(∅), e.g. RE `$^` (matches "string begin _after_ string end") or `$.` ("any one character after the string end"), it won’t allow the other to be None. But this edge case can be discarded for now.

BlueDrink9: Ah, I see. It assumes the item has a filter - which makes sense given the class it's in. But possibly this should be made explicit?

PR #358, opened 2020-01-25
Better script error handling

luziferius:
Rewritten how the `ScriptRunner` class handles errors in user scripts.
Instead of simply storing a single string, the class now creates a simple container object and pushes it into a list.
The user can then
- view all errors in order
- remove them individually (middle button with trash can icon)
- clear the list (leftmost button with reset icon)
- close the window without removing the error messages from the list. The plain Close button (red x) won’t delete data from the error list.

GUI screenshot (ignore the localized buttons in the button box, Qt automatically translates them):
Script error

TODO:
- [ ] Implement the GTK GUI dialogue
- [ ] Store the trigger mode, and abbreviation or hotkey, if applicable. Display that information, as it may be of use for debugging purposes.
- [ ] Some sort of rate limiting (store only last 1000 exceptions or so)
- [x] Maybe some further improvements, like a integer spin box to show the current error number. Can be used for easier navigation when the exception list is large.
- [ ] Add an API call? Maybe `engine.clear_error_records()`?

Edit: Getting the source of a triggered Script seems to be a quite challenging task. Some sources, like the DBus interface and the GUI buttons are easy to identify, but getting the actual triggered abbreviation is quite a task. :-1:

### DIFF
--- a/lib/autokey/qtapp.py
+++ b/lib/autokey/qtapp.py
@@ -301,18 +301,6 @@ class Application(QApplication):
             message_box.setDetailedText(details)
         message_box.exec_()
 
-    def show_script_error(self):
-        """
-        Show the last script error (if any)
-        """
-        # TODO: i18n
-        if self.service.scriptRunner.error:
-            details = self.service.scriptRunner.error
-            self.service.scriptRunner.error = ''
-        else:
-            details = "No error information available"
-        QMessageBox.information(None, "View Script Error Details", details)
-
     def show_popup_menu(self, folders: list=None, items: list=None, onDesktop=True, title=None):
         if items is None:
             items = []
--- a/lib/autokey/qtapp.py
+++ b/lib/autokey/qtapp.py
@@ -106,9 +106,6 @@ class Application(QApplication):
         logger.info("Initialising application")
         self.setWindowIcon(QIcon.fromTheme(common.ICON_FILE, ui_common.load_icon(ui_common.AutoKeyIcon.AUTOKEY)))
         try:
-
-            # Initialise logger
-
             if self._verify_not_running():
                 self._create_lock_file()
 
@@ -119,6 +116,11 @@ class Application(QApplication):
             self._try_start_service()
             self.notifier = Notifier(self)
             self.configWindow = ConfigWindow(self)
+            # Connect the mutual connections between the tray icon and the main window
+            self.configWindow.action_show_last_script_error.triggered.connect(self.notifier.reset_tray_icon)
+            self.notifier.action_view_script_error.triggered.connect(
+                self.configWindow.show_script_errors_dialog.update_and_show)
+
             self.monitor.start()
             # Initialise user code dir
             if self.configManager.userCodeDir is not None:
--- a/lib/autokey/qtapp.py
+++ b/lib/autokey/qtapp.py
@@ -32,7 +32,7 @@ from PyQt5.QtWidgets import QMessageBox, QApplication
 from autokey import common
 common.USING_QT = True
 
-from autokey import service, monitor
+from autokey import service, monitor, model
 
 import autokey.argument_parser
 import autokey.configmanager.configmanager as cm
@@ -266,13 +266,15 @@ class Application(QApplication):
         os.remove(common.LOCK_FILE)  # TODO: maybe use atexit to remove the lock/pid file?
         logger.debug("All shutdown tasks complete... quitting")
 
-    def notify_error(self, message):
+    def notify_error(self, error: model.ScriptErrorRecord):
         """
         Show an error notification popup.
 
-        @param message: Message to show in the popup
+        @param error: The error that occurred in a Script
         """
+        message = "The script '{}' encountered an error".format(error.script_name)
         self.exec_in_main(self.notifier.notify_error, message)
+        self.configWindow.script_errors_available.emit(True)
 
     def update_notifier_visibility(self):
         self.notifier.update_visible_status()
--- a/lib/autokey/qtapp.py
+++ b/lib/autokey/qtapp.py
@@ -117,7 +117,7 @@ class Application(QApplication):
             self.notifier = Notifier(self)
             self.configWindow = ConfigWindow(self)
             # Connect the mutual connections between the tray icon and the main window
-            self.configWindow.action_show_last_script_error.triggered.connect(self.notifier.reset_tray_icon)
+            self.configWindow.action_show_last_script_errors.triggered.connect(self.notifier.reset_tray_icon)
             self.notifier.action_view_script_error.triggered.connect(
                 self.configWindow.show_script_errors_dialog.update_and_show)
 
--- a/lib/autokey/qtui/configwindow.py
+++ b/lib/autokey/qtui/configwindow.py
@@ -42,6 +42,7 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
         super().__init__()
         self.setupUi(self)
         self.about_dialog = dialogs.AboutAutokeyDialog(self)
+        self.show_script_errors_dialog = dialogs.ShowRecentScriptErrorsDialog(self)
         self.app = app
         self.action_create = self._create_action_create()
         self.toolbar.insertAction(self.action_save, self.action_create)  # Insert before action_save, i.e. at index 0
@@ -102,7 +103,7 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
 
     def _connect_all_tools_menu_signals(self):
         self.action_show_last_script_error.triggered.connect(self.app.notifier.reset_tray_icon)
-        self.action_show_last_script_error.triggered.connect(self.app.show_script_error)
+        self.action_show_last_script_error.triggered.connect(self.show_script_errors_dialog.update_and_show)
         self.action_record_script.triggered.connect(self.on_record)
         self.action_run_script.triggered.connect(self.on_run_script)
         # Add all defined macros to the »Insert Macros« menu
--- a/lib/autokey/qtui/configwindow.py
+++ b/lib/autokey/qtui/configwindow.py
@@ -39,11 +39,16 @@ PROBLEM_MSG_SECONDARY = "%1\n\nYour changes have not been saved."
 
 class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwindow")):
 
+    script_errors_available = pyqtSignal(bool, name="script_errors_available")
+
     def __init__(self, app: QApplication):
         super().__init__()
         self.setupUi(self)
         self.about_dialog = dialogs.AboutAutokeyDialog(self)
         self.show_script_errors_dialog = dialogs.ShowRecentScriptErrorsDialog(self)
+        # Just forward the signal
+        self.show_script_errors_dialog.script_errors_available.connect(self.script_errors_available)
+
         self.app = app
         self.action_create = self._create_action_create()
         self.toolbar.insertAction(self.action_save, self.action_create)  # Insert before action_save, i.e. at index 0
@@ -104,6 +109,8 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
 
     def _connect_all_tools_menu_signals(self):
         self.action_show_last_script_error.triggered.connect(self.show_script_errors_dialog.update_and_show)
+        # Only enable the action if script errors are recorded. Disable the action if no errors are viewable.
+        self.script_errors_available.connect(self.action_show_last_script_error.setDisabled)
         self.action_record_script.triggered.connect(self.on_record)
         self.action_run_script.triggered.connect(self.on_run_script)
         # Add all defined macros to the »Insert Macros« menu
--- a/lib/autokey/qtui/configwindow.py
+++ b/lib/autokey/qtui/configwindow.py
@@ -18,6 +18,7 @@ import threading
 import time
 import webbrowser
 
+from PyQt5.QtCore import pyqtSignal
 from PyQt5.QtGui import QIcon, QKeySequence, QCloseEvent
 from PyQt5.QtWidgets import QApplication, QAction, QMenu
 
@@ -102,7 +103,6 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
         self.action_rename_item.triggered.connect(self.central_widget.on_rename)
 
     def _connect_all_tools_menu_signals(self):
-        self.action_show_last_script_error.triggered.connect(self.app.notifier.reset_tray_icon)
         self.action_show_last_script_error.triggered.connect(self.show_script_errors_dialog.update_and_show)
         self.action_record_script.triggered.connect(self.on_record)
         self.action_run_script.triggered.connect(self.on_run_script)
--- a/lib/autokey/qtui/configwindow.py
+++ b/lib/autokey/qtui/configwindow.py
@@ -108,9 +108,9 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
         self.action_rename_item.triggered.connect(self.central_widget.on_rename)
 
     def _connect_all_tools_menu_signals(self):
-        self.action_show_last_script_error.triggered.connect(self.show_script_errors_dialog.update_and_show)
+        self.action_show_last_script_errors.triggered.connect(self.show_script_errors_dialog.update_and_show)
         # Only enable the action if script errors are recorded. Disable the action if no errors are viewable.
-        self.script_errors_available.connect(self.action_show_last_script_error.setDisabled)
+        self.script_errors_available.connect(self.action_show_last_script_errors.setEnabled)
         self.action_record_script.triggered.connect(self.on_record)
         self.action_run_script.triggered.connect(self.on_run_script)
         # Add all defined macros to the »Insert Macros« menu
--- a/lib/autokey/qtui/configwindow.py
+++ b/lib/autokey/qtui/configwindow.py
@@ -18,7 +18,7 @@ import threading
 import time
 import webbrowser
 
-from PyQt5.QtCore import pyqtSignal
+from PyQt5.QtCore import pyqtSignal, QTimer
 from PyQt5.QtGui import QIcon, QKeySequence, QCloseEvent
 from PyQt5.QtWidgets import QApplication, QAction, QMenu
 
@@ -286,14 +286,14 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
             self.action_record_script.setChecked(False)
 
     def on_run_script(self):
-        t = threading.Thread(target=self._run_script)
-        t.start()
-
-    def _run_script(self):
         script = self.central_widget.get_selected_item()[0]
-        time.sleep(2)  # Fix the GUI tooltip for action_run_script when changing this!
-        self.app.service.scriptRunner.execute(script)
-    
+        QTimer.singleShot(
+            2000,  # Fix the GUI tooltip for action_run_script when changing this!
+            (lambda: self.app.service.scriptRunner.execute(
+                script
+            ))
+        )
+
     # Settings Menu
             
     def on_advanced_settings(self):
--- a/lib/autokey/qtui/configwindow.py
+++ b/lib/autokey/qtui/configwindow.py
@@ -45,10 +45,7 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
         super().__init__()
         self.setupUi(self)
         self.about_dialog = dialogs.AboutAutokeyDialog(self)
-        self.show_script_errors_dialog = dialogs.ShowRecentScriptErrorsDialog(self)
-        # Just forward the signal
-        self.show_script_errors_dialog.script_errors_available.connect(self.script_errors_available)
-
+        self.show_script_errors_dialog = self._create_show_recent_script_errors_dialog()
         self.app = app
         self.action_create = self._create_action_create()
         self.toolbar.insertAction(self.action_save, self.action_create)  # Insert before action_save, i.e. at index 0
@@ -80,6 +77,12 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
         action_create.setMenu(create_menu)
         return action_create
 
+    def _create_show_recent_script_errors_dialog(self) -> dialogs.ShowRecentScriptErrorsDialog:
+        show_script_errors_dialog = dialogs.ShowRecentScriptErrorsDialog(self)
+        # Forward the signal from the dialog instance to the own signal
+        show_script_errors_dialog.script_errors_available.connect(self.script_errors_available)
+        return show_script_errors_dialog
+
     def _connect_all_file_menu_signals(self):
         # Show the action_create popup menu regardless where the user places the click.
         # The Action is displayed as "[<icon>]v". Clicking on the downwards arrow opens the popup menu as
@@ -109,8 +112,11 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
 
     def _connect_all_tools_menu_signals(self):
         self.action_show_last_script_errors.triggered.connect(self.show_script_errors_dialog.update_and_show)
-        # Only enable the action if script errors are recorded. Disable the action if no errors are viewable.
+        # Only enable action_show_last_script_errors if script errors are recorded.
+        # Automatically disable the action if no errors are viewable to prevent the user from seeing a dialogue window
+        # in some undefined state.
         self.script_errors_available.connect(self.action_show_last_script_errors.setEnabled)
+
         self.action_record_script.triggered.connect(self.on_record)
         self.action_run_script.triggered.connect(self.on_run_script)
         # Add all defined macros to the »Insert Macros« menu
@@ -289,7 +295,7 @@ class ConfigWindow(*autokey.qtui.common.inherits_from_ui_file_with_name("mainwin
         script = self.central_widget.get_selected_item()[0]
         QTimer.singleShot(
             2000,  # Fix the GUI tooltip for action_run_script when changing this!
-            (lambda: self.app.service.scriptRunner.execute(
+            (lambda: self.app.service.scriptRunner.execute_script(
                 script
             ))
         )
--- a/lib/autokey/qtui/dialogs/show_recent_script_errors.py
+++ b/lib/autokey/qtui/dialogs/show_recent_script_errors.py
@@ -39,6 +39,8 @@ class ShowRecentScriptErrorsDialog(*ui_common.inherits_from_ui_file_with_name("s
     # this and enables/disables itself based on the boolean value. The connection is defined in the .ui file.
     has_next_error = pyqtSignal(bool, name="has_next_error")
 
+    script_errors_available = pyqtSignal(bool, name="script_errors_available")
+
     def __init__(self, parent):
         super(ShowRecentScriptErrorsDialog, self).__init__(parent)
         self.setupUi(self)
@@ -50,6 +52,10 @@ class ShowRecentScriptErrorsDialog(*ui_common.inherits_from_ui_file_with_name("s
         self.recent_script_errors = QApplication.instance().\
             service.scriptRunner.error_records  # type: typing.List[ScriptErrorRecord]
         self.currently_viewed_error_index = 0
+
+    def hide(self):
+        self.script_errors_available.emit(bool(self.recent_script_errors))
+        super(ShowRecentScriptErrorsDialog, self).hide()
 
     @pyqtSlot(QAbstractButton)
     def handle_button_box_buttons(self, clicked_button: QAbstractButton):
--- a/lib/autokey/qtui/dialogs/show_recent_script_errors.py
+++ b/lib/autokey/qtui/dialogs/show_recent_script_errors.py
@@ -30,7 +30,10 @@ class ShowRecentScriptErrorsDialog(*ui_common.inherits_from_ui_file_with_name("s
     """
     This dialogue is used to show errors that caused user Scripts to abort. The ScriptRunner class holds a list
     with errors, which can be viewed and cleared using this dialogue window.
+
     """
+    # TODO: When the minimal python version is raised to >= 3.6, add millisecond display to both the script start
+    # timestamp and the error timestamp.
 
     # When switching errors, emit a boolean indicating if previous errors are available. The Previous button reacts on
     # this and enables/disables itself based on the boolean value. The connection is defined in the .ui file.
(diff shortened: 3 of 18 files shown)


BlueDrink9:
Want this reviewed?

luziferius:
> Want this reviewed?

If you wish, you can do so.

Edit: I’ve added more comments to the GUI window logic, so it should be understandable now.

You can definitely help with the user-facing GUI strings. I’m not a native English speaker, and based on your location on your GitHub profile you are, so you know that stuff better than me.
If you see awkwardly constructed sentences or other things to point out, please do so.

lib/autokey/qtapp.py
```diff
@@ -119,6 +116,11 @@ def __init__(self, argv: list=sys.argv):
             self._try_start_service()
             self.notifier = Notifier(self)
             self.configWindow = ConfigWindow(self)
+            # Connect the mutual connections between the tray icon and the main window
+            self.configWindow.action_show_last_script_errors.triggered.connect(self.notifier.reset_tray_icon)
```
BlueDrink9, line 120: I feel like this should be extracted into a new function. May as well make new code tidy, even if the old stuff isn't.

lib/autokey/qtui/configwindow.py
```diff
@@ -38,10 +39,16 @@
...
 
+    script_errors_available = pyqtSignal(bool, name="script_errors_available")
+
     def __init__(self, app: QApplication):
         super().__init__()
         self.setupUi(self)
         self.about_dialog = dialogs.AboutAutokeyDialog(self)
+        self.show_script_errors_dialog = dialogs.ShowRecentScriptErrorsDialog(self)
+        # Just forward the signal
+        self.show_script_errors_dialog.script_errors_available.connect(self.script_errors_available)
```
BlueDrink9: Again, move to separate func?

lib/autokey/qtui/configwindow.py
```diff
@@ -101,8 +108,9 @@ def _connect_all_edit_menu_signals(self):
         self.action_rename_item.triggered.connect(self.central_widget.on_rename)
 
     def _connect_all_tools_menu_signals(self):
-        self.action_show_last_script_error.triggered.connect(self.app.notifier.reset_tray_icon)
-        self.action_show_last_script_error.triggered.connect(self.app.show_script_error)
+        self.action_show_last_script_errors.triggered.connect(self.show_script_errors_dialog.update_and_show)
+        # Only enable the action if script errors are recorded. Disable the action if no errors are viewable.
+        self.script_errors_available.connect(self.action_show_last_script_errors.setEnabled)
```
BlueDrink9, line 118: Separate func. Also, these lines look similar to the others where I've suggested this, is there any duplication?

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+    def handle_button_box_buttons(self, clicked_button: QAbstractButton):
+        """
+        Used to connect the button presses with logic for 'Reset error list' and 'Discard current error' buttons in
+        the button box at the bottom of the dialogue window.
+        """
+        button_role = self.buttonBox.buttonRole(clicked_button)
+        # The Close is handled internally and simply hides the dialogue, therefore nothing is implemented here.
+        if button_role == QDialogButtonBox.DestructiveRole:  # Discard current error
+            self.remove_currently_shown_error_from_error_list()
+        elif button_role == QDialogButtonBox.ResetRole:  # Clear the error list
```
BlueDrink9, line 92: (I'm nitpicking here, but) Your function names are good here, you don't need the comments. They risk becoming out of date.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+            self.clear_error_list()
+
+    @pyqtSlot()
+    def update_and_show(self):
+        error_count = len(self.recent_script_errors)
+        if error_count:
+            if self.currently_viewed_error_index >= error_count:
```
BlueDrink9, line 99: This is all reasonable. I'd suggest extracting out funcs:

```python
if error_count:
    reset_currently_viewed_error()
    update_has_next_and_prev_error()  # I'm less sure about this func name, I might be misinterpreting the purpose of lines 80-81.
```

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+                self.currently_viewed_error_index = error_count-1
+
+            self.has_next_error.emit(self.currently_viewed_error_index < error_count-1)
```
BlueDrink9: This line is duplicated at line 102, and it's a little hard to tell what the condition is. Consider making either the whole line, just the condition, or both a function.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+    def show_next_error(self):
+        """
+        Switch to the next error in the error list.
+        The connection from the Next button to this slot function is defined in the .ui file.
+        """
+        logger.debug("User views the next error.")
+        # Out of bounds handling is not needed, as the Next button gets disables when the last error is reached.
+        # See has_next_error Signal.
+        self.currently_viewed_error_index += 1
+        self.has_next_error.emit(self.currently_viewed_error_index < len(self.recent_script_errors)-1)
```
BlueDrink9: See line 80

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @pyqtSlot()
+    def update_and_show(self):
+        error_count = len(self.recent_script_errors)
+        if error_count:
+            if self.currently_viewed_error_index >= error_count:
+                self.currently_viewed_error_index = error_count-1
+
+            self.has_next_error.emit(self.currently_viewed_error_index < error_count-1)
+            self.has_previous_error.emit(self.currently_viewed_error_index > 0)
```
BlueDrink9: Also duplicated at 115. Create a function.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+    def show_previous_error(self):
+        """
+        Switch to the previous error in the error list.
+        The connection from the Previous button to this slot function is defined in the .ui file.
+        """
+        logger.debug("User views the previous error.")
+        # Out of bounds handling is not needed, as the Previous button gets disables when the first error is reached.
+        # See has_previous_error Signal.
+        self.currently_viewed_error_index -= 1
+        self.has_previous_error.emit(self.currently_viewed_error_index > 0)
```
BlueDrink9: line 81

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+        self._show_error(self.currently_viewed_error_index)
```
BlueDrink9: This line is duplicated a few times, may be worth its own wrapper `self._show_currently_viewed_error()`

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        error_count = len(self.recent_script_errors)
+        if error_count == 1:
+            self.clear_error_list()
+        else:
+            del self.recent_script_errors[self.currently_viewed_error_index]
+            if self.currently_viewed_error_index == error_count-1:
+                self.show_previous_error()
+            else:
+                self.has_next_error.emit(self.currently_viewed_error_index < len(self.recent_script_errors)-1)
```
BlueDrink9: Looks familiar

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+logger = __import__("autokey.logger").logger.get_logger(__name__)
+
+
+class ShowRecentScriptErrorsDialog(*ui_common.inherits_from_ui_file_with_name("show_recent_script_errors_dialog")):
+    """
+    This dialogue is used to show errors that caused user Scripts to abort. The ScriptRunner class holds a list
+    with errors, which can be viewed and cleared using this dialogue window.
+    """
+
+    # When switching errors, emit a boolean indicating if further errors are available. The Next button reacts on
```
BlueDrink9: `has_previous` is the boolean for further errors? Or are the comments here swapped around?

lib/autokey/qtui/dialogs/show_recent_script_errors.py
```diff
@@ -0,0 +1,152 @@
...
+            self.clear_error_list()
+        else:
+            del self.recent_script_errors[self.currently_viewed_error_index]
+            if self.currently_viewed_error_index == error_count-1:
+                self.show_previous_error()
+            else:
+                self.has_next_error.emit(self.currently_viewed_error_index < len(self.recent_script_errors)-1)
+                self._show_error(self.currently_viewed_error_index)
+
+    def clear_error_list(self):
```
BlueDrink9: `clear_error_list_and_hide`?

lib/autokey/qtui/resources/ui/mainwindow.ui
```diff
@@ -352,16 +352,19 @@
...
+   <property name="enabled">
+    <bool>false</bool>
+   </property>
    <property name="icon">
     <iconset theme="dialog-error">
      <normaloff>../../../../../../../.designer/backup</normaloff>../../../../../../../.designer/backup</iconset>
    </property>
    <property name="text">
-    <string>&amp;Show last script error</string>
+    <string>&amp;Show last script errors</string>
```
BlueDrink9: Reads better as 'Show errors from the last script', if this is what I think it is (a button label or similar)

lib/autokey/qtui/resources/ui/mainwindow.ui
```diff
@@ -352,16 +352,19 @@
...
    </property>
    <property name="toolTip">
-    <string>Show the last occured script error, if any.</string>
+    <string>Show the recently occured script errors.</string>
```
BlueDrink9: 'Show recently-occurred errors from run scripts' if this only displays *errors* that rose recently.

'Show occurred errors from recently-run scripts' if this only displays errors from recently-run *scripts*.

lib/autokey/qtui/resources/ui/show_recent_script_errors_dialog.ui
```diff
@@ -0,0 +1,350 @@
...
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>762</width>
+    <height>565</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>View most recent Script errors</string>
```
BlueDrink9: See review in mainwindow.ui

lib/autokey/qtui/resources/ui/show_recent_script_errors_dialog.ui
```diff
@@ -0,0 +1,350 @@
...
+    <widget class="QLabel" name="script_trigger_source_label">
+     <property name="text">
+      <string/>
+     </property>
+    </widget>
+   </item>
+   <item row="0" column="0" colspan="2">
+    <widget class="QLabel" name="dialog_header_label">
+     <property name="text">
+      <string>Shown below are the most recent errors that occured in Scripts run by AutoKey during this session.</string>
```
BlueDrink9: How are 'most recent errors' defined? Most recent 10? All errors in the last 10 mins?

lib/autokey/service.py
```diff
@@ -489,28 +497,37 @@ def execute(self, script: model.Script, buffer=''):
...
+
+    def _execute(self, scope, script: typing.Union[model.Script, pathlib.Path]):
+        script_code = script.read_text() if isinstance(script, pathlib.Path) else script.code
+        start_time = datetime.datetime.now().time()
+        # noinspection PyBroadException
         try:
-            exec(path.read_text(), scope)
+            exec(script_code, scope)
         except Exception as e:
+            error_time = datetime.datetime.now().time()
```
BlueDrink9: The exception block could possibly be its own function?

lib/autokey/service.py
```diff
@@ -475,6 +480,9 @@ def __init__(self, mediator: IoMediator, app):
 
         self.engine = self.scope["engine"]
 
+    def clear_error_records(self):
+        self.error_records.clear()
+
     @threaded
     def execute(self, script: model.Script, buffer=''):
```
BlueDrink9: Buffer is unused, right? I assume you don't want to remove it for deprecation reasons?

lib/autokey/service.py
```diff
@@ -489,28 +497,37 @@ def execute(self, script: model.Script, buffer=''):
...
         # Overwrite __file__ to contain the path to the user script instead of the path to this service.py file.
         scope["__file__"] = path.resolve()
+        self._execute(scope, path)
+
+    def _record_error(self, error: model.ScriptErrorRecord):
+        self.error_records.append(error)
+        self.app.notify_error(error)
+
+    def _execute(self, scope, script: typing.Union[model.Script, pathlib.Path]):
+        script_code = script.read_text() if isinstance(script, pathlib.Path) else script.code
```
BlueDrink9: This is a nice refactor of the two exec methods.

BlueDrink9: Would be easier to read as `script_code = self._get_script_code(script)` (new func)

lib/autokey/service.py
```diff
@@ -489,28 +497,37 @@ def execute(self, script: model.Script, buffer=''):
         if script.path is not None:
             # Overwrite __file__ to contain the path to the user script instead of the path to this service.py file.
             scope["__file__"] = script.path
-        try:
-            exec(script.code, scope)
-        except Exception as e:
-            logger.exception("Script error")
-            self.error = "Script name: '{}'\n{}".format(script.description, traceback.format_exc())
-            self.app.notify_error("The script '{}' encountered an error".format(script.description))
+        self._execute(scope, script)
```
BlueDrink9, line 498: It's confusing having such similar func names (`execute` and `_execute`), and probably easy to make a mistake and use the wrong one. I'm having trouble thinking up alternatives though... `_execute_and_record_errors`?

BlueDrink9 (changes requested):
I've reviewed. English all looks ganz gut. Only a couple of suggestions to help it read better, and they're definitely more because you're a programmer than because you're a non-native speaker ;) 

I've been reading Robert Martin's *Clean Code*, which is why a lot of the comments are related to refactoring into nicely-named functions. Sorry if it comes across as pedantic.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
luziferius: I moved that hide() into the function after writing the header.
Will do, as that is semantically totally relevant.

lib/autokey/qtui/configwindow.py
luziferius: I’ll put it into a function that returns the `ShowRecentScriptErrorsDialog(self)` instance.
That way, the member variable is assigned in `__init__`, which conforms to the style guides of not creating members outside of `__init__`

lib/autokey/qtui/configwindow.py
luziferius: Here, I’m opposing that suggestion. This is branch-free logic that does one single thing, namely connecting the logic for the entries in the named GUI menu.

Those are definitely not duplicates.
The first one in `__init__` connects the `ShowRecentScriptErrorsDialog.script_errors_available` Qt _Signal_
with the  `self.script_errors_available` Qt _Signal_ (basically just forwarding the signal),
whilst this line connects the Signal with the appropriate Qt _Slot_ function of the menu QAction. (Which causes the Qt event loop to call the slot function whenever the signal is `emit`ted.)

lib/autokey/qtui/resources/ui/mainwindow.ui
luziferius: It is a menu action label, found in the `Tools` menu. With this PR, it shows all recorded errors from all scripts.

lib/autokey/qtui/resources/ui/show_recent_script_errors_dialog.ui
luziferius: Currently, it is everything that the user hasn’t already discarded/removed. Maybe changing `most recent` to `recorded`.
How does `Shown below are all recorded errors that occured in Scripts run by AutoKey during this session.` sound?

lib/autokey/qtui/dialogs/show_recent_script_errors.py
luziferius: I checked the XML definition and you are right. How did _that_ happen? :see_no_evil:

That’s why people say that writing comments is bad, as they get outdated or are plainly wrong…

lib/autokey/qtui/dialogs/show_recent_script_errors.py
luziferius: Closing and re-opening the dialogue keeps the currently viewed error (stored as a list index in `self.currently_viewed_error_index`).
`if self.currently_viewed_error_index >= error_count` clamps the stored value into the valid range [0, `error_count`). This is currently dead code, as nothing else modifies the list at the moment.

But if there is an API call that clears the list, it will become handy and handle that case.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
luziferius: I can just name the boolean by extracting a variable or extract the condition into a function or property. it computes "is the current error index below the end of the error list", which is equivalent to "has next error".
This line, plus the next cause both "Next" and "Previous" buttons to be disabled if there is exactly one error in the error list.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
luziferius: Hmmm. I wrote it as a generic function initially. But it is only ever called using the current index, as you pointed out.

So renaming it to `self._show_currently_viewed_error()` and removing the parameter is a good idea. The function parameter is currently always a member variable, so the value is already available inside the function.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
luziferius: It has a similar structure to the update_and_show() function.

lib/autokey/qtui/resources/ui/mainwindow.ui
luziferius: I think, I’ll take the latter.

lib/autokey/service.py
luziferius: You checked based on the diff shown here? Then GitHub’s diff rendering is at fault. It is used here in line 493:
https://github.com/autokey/autokey/blob/5fcbe73d35f4c41718c57dfda37a79a8441e29cb/lib/autokey/service.py#L486-L493

lib/autokey/service.py
luziferius: I can rename `execute()` to `execute_script()`, which will have a similar naming structure to `execute_path()`. This isn’t part of any user-facing API, so a global rename is safe to do.

lib/autokey/service.py
luziferius: I’m quick to write ternary operators…
And putting my one-liner into a separate function feels odd.

How about this?
```python
if isinstance(script, pathlib.Path):
    script_code = script.read_text()
else:
    script_code = script.code
```
or even 
```python
if isinstance(script, pathlib.Path):
    script_code = script.read_text()
elif isinstance(script, model.Script):
    script_code = script.code
else:
    raise TypeError(
        "Unknown script type passed in, expected one of "
        "[autokey.model.Script, pathlib.Path], got {}".format(
              type(script))
    )
```
Especially the second one warrants a separate function.

lib/autokey/service.py
luziferius: I can move everything into `_record_error()`, which is called in the except block and currently doesn’t do much:
https://github.com/autokey/autokey/blob/5fcbe73d35f4c41718c57dfda37a79a8441e29cb/lib/autokey/service.py#L512-L514

BlueDrink9: Good

lib/autokey/qtui/configwindow.py
BlueDrink9: Fair. Leave a comment? or, better, put each line in its own well-named func?

lib/autokey/qtui/dialogs/show_recent_script_errors.py
BlueDrink9: This snippet isn't the one I meant to refer to with this comment. I'll correct later

lib/autokey/qtui/dialogs/show_recent_script_errors.py
BlueDrink9: Sorry, this is my first code review. I was referring only to line 132, which is same as 80/81. I didn't expect the whole block to come with it

lib/autokey/qtui/resources/ui/mainwindow.ui
BlueDrink9: Ok. then change it, + *scripts

lib/autokey/service.py
BlueDrink9: Seems we may as well do the second one.

I can't stand ternaries except in python, but when the clauses are long it hinders readability

lib/autokey/service.py
BlueDrink9: Perfect. `execute_Script` or better `execute_stored_script`? to distinguish that it takes an autokey object?

lib/autokey/service.py
BlueDrink9: Ah, will check again later

lib/autokey/qtui/resources/ui/show_recent_script_errors_dialog.ui
BlueDrink9: Good

lib/autokey/qtui/resources/ui/mainwindow.ui
BlueDrink9: And maybe change "occurred" to "recorded"

lib/autokey/service.py
luziferius: I’m opposing `execute_stored_script`. A script stored in a file pointed to by a _Path_ is technically still _stored_ (as in stored on disk). And _temporary_ AutoKey scripts (that you added support for) are technically _not_ stored (as they are in memory only). So the function `execute_stored_script` can’t execute _temporary_ Scripts?
Since that’s not true, I find that more confusing than the plain `execute_script`.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
luziferius: > Sorry, this is my first code review. 

It’s my second :)


> I was referring only to line 132, which is same as 80/81. I didn't expect the whole block to come with it

The GitHub interface seems to be a bit ambiguous here. So I interpreted that two-words comment together with the full block quoted by GitHub as actually referring to the full block quoted. And your comment is technically correct for the full block, too, as the logic flow is kind of similar. So that didn’t clear the ambiguity, either.

luziferius:
I’ve addressed most of your points in your comments. Then I wanted to address one thing that bothered me: The file name shown in Tracebacks is set to `<string>`, which is kind of _helpful_, so to speak. So I wanted to change it. Based on https://stackoverflow.com/questions/12977578/how-do-i-set-the-filename-of-a-module-in-python-so-that-exceptions-that-occur-in, it should be possible by replacing `exec()` with a combo of `compile()` and `exec()`.

This works as intended, still showing `<string>` as the file name:
```python
script_name = "<string>"
start_time = datetime.datetime.now().time()
try:
    compiled_code = compile(script_code, script_name, 'exec')
    exec(compiled_code, scope)
except Exception:
    self._record_error(script, start_time)
```
The resulting error message:
```
  File "/mnt/Daten/Projekte/autokey/lib/autokey/service.py", line 530, in _execute
    exec(compiled_code, scope)
  File "<string>", line 16, in <module>
  File "<string>", line 8, in foo
  File "<string>", line 8, in foo
  File "<string>", line 8, in foo
  [Previous line repeated 497 more times]
  File "<string>", line 10, in foo
  File "<string>", line 14, in bar
ZeroDivisionError: division by zero
```
Now I change that name to something else. Using a constant here for testing and demonstration purposes:

```python
script_name = "whatever"
start_time = datetime.datetime.now().time()
try:
    compiled_code = compile(script_code, script_name, 'exec')
    exec(compiled_code, scope)
except Exception:
    self._record_error(script, start_time)
```
Then _this_ abomination is printed: 
```
  File "/mnt/Daten/Projekte/autokey/lib/autokey/service.py", line 530, in _execute
    exec(compiled_code, scope)
  File "whatever", line 16, in <module>
    # You should have received a copy of the GNU General Public License
  File "whatever", line 8, in foo
    # the Free Software Foundation, either version 3 of the License, or
  File "whatever", line 8, in foo
    # the Free Software Foundation, either version 3 of the License, or
  File "whatever", line 8, in foo
    # the Free Software Foundation, either version 3 of the License, or
  [Previous line repeated 497 more times]
  File "whatever", line 10, in foo
    #
  File "whatever", line 14, in bar
    # GNU General Public License for more details.
ZeroDivisionError: division by zero
```
WTF? Where do those spurious GPL lines come from?

Edit: If set, the file name must be an absolute path pointing to the file on disk. It is then loaded by the Traceback walker to load the code lines. Otherwise it’ll load the current source file and prints what it finds there, which is the GPL header of `service.py`.
This means, I’ll have to find a solution for temporary scripts. Edit 2: I’ll use `<string>` for now, if the Script has no file system path.

I’ve put some additional work in it, now it works nicely with permanent Scripts and Subscripts run by `engine.run_script()`:
```
  File "/mnt/Daten/Projekte/autokey/lib/autokey/service.py", line 529, in _execute
    exec(compiled_code, scope)
  File "/home/thomas/.config/autokey/data/test/error2.py", line 5, in <module>
    bar2()
  File "/home/thomas/.config/autokey/data/test/error2.py", line 3, in bar2
    engine.run_script("error")
  File "/mnt/Daten/Projekte/autokey/lib/autokey/scripting/engine.py", line 365, in run_script
    self.runner.run_subscript(target_script)
  File "/mnt/Daten/Projekte/autokey/lib/autokey/service.py", line 573, in run_subscript
    exec(compiled_code, scope)
  File "/home/thomas/.config/autokey/data/test/error.py", line 16, in <module>
    foo()
  File "/home/thomas/.config/autokey/data/test/error.py", line 8, in foo
    foo()
  File "/home/thomas/.config/autokey/data/test/error.py", line 8, in foo
    foo()
  File "/home/thomas/.config/autokey/data/test/error.py", line 8, in foo
    foo()
  [Previous line repeated 497 more times]
  File "/home/thomas/.config/autokey/data/test/error.py", line 10, in foo
    bar()
  File "/home/thomas/.config/autokey/data/test/error.py", line 14, in bar
    return 1/0
ZeroDivisionError: division by zero
```

BlueDrink9:
Glorious,  except for the line duplications. Great work!
 I wonder what variable python is reading… what's printed by a script that manually sets `__file__`?

luziferius:
> I wonder what variable python is reading… what's printed by a script that manually sets `__file__`?

That shouldn’t matter. It seems that the file name is baked into the compiled bytecode object by `compile()`. So setting `__file__` inside compiled code shouldn’t do anything regarding the Traceback printing.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
BlueDrink9: Actually yes it is. I think you can remove the comments from the end of the `if` clauses

lib/autokey/service.py
BlueDrink9: Yup, my mistake

lib/autokey/qtui/dialogs/show_recent_script_errors.py
luziferius: I’m using the assigned ButtonRole to identify the buttons, as the buttons are automatically generated using a bitmask. It is not immediately clear that the button assigned to the  `DestructiveRole` deletes the _current_ item and not _all_ items.

luziferius:
I tested it: This is the Script I used to generate the errors. I added setting `__file__` (and `__name__` for completeness) and nothing happened. The Traceback still uses the correct values. It is indeed hardcoded into the compiled bytecode.

```python
i = 0
__file__ = "testtext123"
__name__ = "bla"
def foo():
    global i
    if i < 500:
        i += 1
        foo()
    else:
        bar()
    
def bar():
    time.sleep(2)
    return 1/0
    
foo()
```
Result:
```
  File "/mnt/Daten/Projekte/autokey/lib/autokey/service.py", line 529, in _execute
    exec(compiled_code, scope)
  File "/home/thomas/.config/autokey/data/test/error.py", line 17, in <module>
    foo()
  File "/home/thomas/.config/autokey/data/test/error.py", line 9, in foo
    foo()
  File "/home/thomas/.config/autokey/data/test/error.py", line 9, in foo
    foo()
  File "/home/thomas/.config/autokey/data/test/error.py", line 9, in foo
    foo()
  [Previous line repeated 497 more times]
  File "/home/thomas/.config/autokey/data/test/error.py", line 11, in foo
    bar()
  File "/home/thomas/.config/autokey/data/test/error.py", line 15, in bar
    return 1/0
ZeroDivisionError: division by zero
```

lib/autokey/service.py
```diff
@@ -519,17 +519,20 @@ def _record_error(self, script: typing.Union[model.Script, pathlib.Path], start_
...
     def _execute(self, scope, script: typing.Union[model.Script, pathlib.Path]):
-        script_code, script_name = self._get_script_source_code_and_name(script)
-        logger.debug("About to execute script {}".format(script_name))
         start_time = datetime.datetime.now().time()
         # noinspection PyBroadException
-
         try:
-            compiled_code = compile(script_code, script_name, 'exec')
+            compiled_code = self._compile_script(script)
             exec(compiled_code, scope)
```
luziferius, line 525: The alternative, naming it `_compile_script_and_execute()` and moving the exec() inside the function, reduces one additional line.
But doing so will make the user-facing Traceback unnecessarily verbose, by adding another step. I’d like to avoid exposing too many internals in those Tracebacks, as those only clutter the data the user is actually interested in.

BlueDrink9: That's a very good point.

lib/autokey/qtui/dialogs/show_recent_script_errors.py
BlueDrink9: Yes it is. It calls `self.remove_*currently*_shown_error_from_error_list()`. You've named that very unambiguously

lib/autokey/service.py
BlueDrink9: I guess ideally we'd modify the stack trace to only include user script errors here. I don't know how feasible that is, and even typing out "modify the stack trace" makes me feel uneasy…

josephj11: I don't understand the fine points here, but since stack traces need to be useful to developers as well as for end users, maybe we could use some sort of log level parameter to vary what is reported (to show more for a developer than what an end user would want to see.)

BlueDrink9: I think that would be ideal. But considering this is about running user scripts, I'm wondering how many users run other people's scripts?

luziferius: > I guess ideally we'd modify the stack trace to only include user script errors here. I don't know how feasible that is, and even typing out "modify the stack trace" makes me feel uneasy…

Hmmm. You’d have to dynamically analyse the traceback and remove all line tuples where the file name matches `__file__`. To determine is how that works if AK is ZIP archive packaged.

I’d like to avoid going that route, as that starts to obstruct the shown logic flow, in case someone does advanced stuff.

It’ll trade one line duplication and one saved line of code with a whole bunch of carefully crafted black magic.

josephj11: Developers might run other people's scripts fairly often when they are submitted to recreate issues. And the "additional" detail (that users wouldn't normally see) would presumably be useful to developers running their own test scripts and phrases during the normal development/testing process. 

Also, as Thomas mentions, an advanced user might need the extra detail - but as an exception - not the rule.

I'm not advocating removing anything, just hiding some of it unless the user requests to see the full/more details.

PR #392, opened 2020-03-23
[RFC]Refactor model classes

### LINKED ISSUE #359: Refactor model.py
Could we extract all the data structure classes from model? Maybe create a new folder 'structures' for things like then new error class, phrases, folders, macros etc so they are all in their own files?

Currently each class is not very discoverable, and it's a hassle to find them within the single 'model' file.

### LINKED ISSUE #372: Circular import trying to test dialogs
## Classification:

Bug

## Reproducibility:

Always

## Version

AutoKey version: Develop

Used GUI (Gtk, Qt, or both): both

## Description

If you go to https://github.com/BlueDrink9/autokey/tree/regex and run the tests, you'll see that one test fails because of a circular import. This makes it very difficult to mock and test UI code, because importing it fails.

Can you please take a look and see if it is totally necessary to have that order of imports there?

luziferius:
Goals:

- - [x] Split `model.py`. (First three commits.) Implements most of #359
- - [x] Fixed #372

Edit: Postponed the point below for a later PR. Keep this one simple, as it is already large enough. The class structure can be reworked later.
- - [ ] Rework the class hierarchy. Detailed plan: https://github.com/autokey/autokey/issues/372#issuecomment-590298910


BlueDrink9:
Want a review?

luziferius:
It's tagged RFC ;-)

It is mostly just a large diff with boring code movement and import adjustments. Most of the diff reads really tiresome.

And I wasn't able to split it into smaller parts, because I didn't want to commit broken in-between steps.

The review:
You can checkout the branch and judge if the end result looks satisfactory. Just don't read the diff line by line :-)

lib/autokey/iomediator/__init__.py
```diff
@@ -1,23 +0,0 @@
...
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
-"""
-IoMediator public interface
-"""
-
-from ._windowgrabber import WindowGrabber
```
luziferius, line 20: Writing this file at all was a mistake. These imports caused even more tight coupling and circular imports. Now, the code just imports directly from the submodules.

Issue #410, opened 2020-04-27
Change default to Paste using clipboard (ctrl+v) for new phrases


josephj11:
## Classification:

Enhancement

## Reproducibility:

N/A

## Version

AutoKey version: All - longstanding problem

Used GUI (Gtk, Qt, or both): Both

Installed via: All

Linux Distribution: All

## Summary

The current default - Paste using keyboard - has a number of problems

1) Fails on long phrases - especially in slow target windows
2) Fails to handle some keys which are not on the physical keyboard
3) Fails with multibyte character sets (everything except EN-US)
4) Fails for some characters on remapped keyboards

## Steps to Reproduce (if applicable)

Define a phrase with any of the above problem characteristics
Define a long phrase (> 20 characters) and use it with a slow application such as a word processor

## Expected Results

Phrase contents should be accurately reproduced

## Actual Results

Phrases are emitted with missing, out of order, and/or wrong characters

We get unnecessary support questions and issues filed, including a couple in the last few days.

## Notes

This has been discussed at length on our forum, Gitter, and in other issues.

Selecting Paste using clipboard (ctrl+v) fixes most issues

I am not suggesting that we get rid of Paste using keyboard or that we do anything unexpected to change the method used automatically. It should still be an option and continue to work as is for the great many preexisting phrases which use it.

josephj11:
See also: #29 which also highlights why Paste using keyboard is still needed for `<cursor>` ...
#241 Emoji
#263 - closed

josephj11:
We seem to be getting a lot of reports of problems like this. I just answered one in #101  .

It might even make sense to add a new Paste using option which works like my type_slow() function to handle extreme cases. It would be better than forcing a user to convert a lot of their phrases to scripts when all they need is for them to paste slower.

BlueDrink9:
Changed in latest beta

vvasuki:
Is this live in 0.96.0?  @josephj11 @BlueDrink9  ??

josephj11:
Paste using clipboard (Ctrl+V) is the new default in 0.96.0. That's why this issue is closed.

 If you're still using 0.95.10, the option is there, but it's not the default.

If you'd like to upgrade and it's not yet available in your distribution's repositories, see these [instructions](https://github.com/autokey/autokey/wiki/Installing). The current release is [here](https://github.com/autokey/autokey/releases/).

0.96.0 contains a lot of bug fixes and several new features. Be sure to read the change log.

josephj11:
[...]

Issue #241, opened 2019-01-27
Emoji are not printed 0.95.1


Zulgrib:
## Classification:

Bug

## Reproducibility:

Always

## Version

AutoKey version: 0.95.1

Used GUI (Gtk, Qt, or both): GTK

If the problem is known to be present in more than one version, please list all of those.

Installed via: PPA

Linux Distribution: Mint19

## Summary

When expanding text from a txt file that contain emojis, the emojis are ignored and not typed.

## Steps to Reproduce (if applicable)

- Add "🔧" to an example phrase
- Set an expension shortcut (I used dollar tmain)
- Write the trigger string

## Expected Results

- My txt is printed as is without alterations

## Actual Results

- Instead, all text is printed excluding the emojis

## Notes

The emojis are properly displayed in the GTK interface and in xed/gedit.

Paste using keyboard

luziferius:
Typing symbols is a bit brittle. If you check the log file, you’ll probably notice some warnings about AutoKey failing to map the symbol. It means that AutoKey failed to register a new X11 key code for the symbol, i.e. failed to create a virtual key that creates the requested symbol.

Switching to clipboard pasting should make the symbols appear. It will also work better with long phrases.

Zulgrib:
Indeed it works.

Looks like keyboard mode is to be kept strictly for key-presses and not writing text.

Fantastic work.

luziferius:
Writing text works, but is a bit slow. Anything below about 100 characters should work within reasonable time. But it only works with characters that you can type using your physical keyboard.

Using the clipboard for longer texts is the way to go. It will also work better with slower applications, like Firefox or LibreOffice, which choke when confronted with the AutoKey typing speed, like in #151.

But you need the keyboard typing if you want to use navigation keys or other special keys inside your phrase.

rodrigograca31:
I'm here to show support for emojis

I'm trying to do something like:
```
keyboard.send_keys("👍")
keyboard.send_keys("<enter>")
```
Also tried:
```
keyboard.send_keys("\U+1F44D")
```
And doesnt type the emoji.

Codepoint at the bottom: https://emojipedia.org/thumbs-up/

josephj11:
@rodrigograca31 Try using

```
clipboard.fill_clipboard("👍")
keyboard.send_keys("<ctrl>+v")
```

That should work in most applications. If you're in a terminal, you'd need

```
keyboard.send_keys("<ctrl>+<shift>+v")
```

[...]

PR #413, opened 2020-05-04
Expand unicode characters using code points

jspaetzel:
What do ya'll think of doing something like this to provide keyboard support for emojis/other symbols?

Related to https://github.com/autokey/autokey/issues/241
and somewhat to https://github.com/autokey/autokey/issues/410

This works pretty well on my machine taking advantage of https://help.ubuntu.com/stable/ubuntu-help/tips-specialchars.html.en#ctrlshiftu seems like this is also fairly standard across distros. Could also always make the code-point trigger configurable if this doesn't work in some case.


jspaetzel:
Cool, thanks!

josephj11:
I hope someone will document this so we know how to use it. Maybe add an example to our API examples on the wiki.

PR #418, opened 2020-05-11
Add dbus handlers for pause_service and unpause_service messages

trxcllnt:
This PR adds dbus handlers to pause and unpause autokey. This allows someone to pause/unpause autokey when using software KVMs like [Barrier](https://github.com/debauchee/barrier):

```bash
# Pause autokey when switching to Mac
dbus-send --session --type=method_call --dest="org.autokey.Service" \
    "/AppService" "org.autokey.Service.pause_service"

# Unpause autokey when switching back to Linux
dbus-send --session --type=method_call --dest="org.autokey.Service" \
    "/AppService" "org.autokey.Service.unpause_service"
```


josephj11:
I have no idea how to do anything with dbus to make use of such a feature, but it sounds like a good idea.

PR #429, opened 2020-07-12
Add wait_for_keyevent scripting function

xpusostomos:
I've set myself a project to get autokey to support emacs style keybinds.
One of the difficulties of that is supporting the C-u <number> prefix to commands:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Prefix-Command-Arguments.html

To do that I need a scripting function that can wait for an arbitrary number of keystrokes, accumulate the result, and store it. To accomplish that I made a new scripting API wait_for_keyevent that takes a lambda/function that says when we are done, and potentially accumulates a result.

I also note that the existing wait_for_keypress API modifies the iomediator.listeners list in the middle of iterating over it. In most computer languages this is illegal and cause all sorts of issues, and non consistent behavior. In any case, it caused a problem for me because it immediately creates a call to the function with the same value. This would presumably cause a problem currently if you say had a script for C-x and then waited for keypress C-x again, it would immediately exit. Anyway, I fixed that by copying the list before iterating, which is the normal way one avoids these multi threaded iterator issues.

Documentation in the code comes with an example of how to use it to implement C-u


xpusostomos:
Closed this because it was based on master not develop branch. Created a new pull request.

Issue #337, opened 2019-12-10
Stop hiding the sidecar files for phrases and scripts


josephj11:
## Reproducibility: N/A

AutoKey version: N/A

Linux Distribution: N/A

## Summary

All phrases and scripts have sidecar .json files which hold all the metadata that AutoKey needs to make them work. The original/current design is to name these files the same name as the script/phrase, but with a leading period to make them hidden files.

As far as I can see, there is no advantage to doing this and it causes a lot of support issues and extra explanations.

Please eliminate the leading period in the names of these sidecar files.

We obviously need to maintain compatibility with existing phrases and scripts. Or, as Thomas has suggested, we could automatically convert existing phrases and scripts to the new format.

luziferius:
No objection from my side.

By default everything is kept below `.config/autokey`, so it shouldn’t make any major difference in the overall appearance for the average user who does not add external directories.

Unhiding the files should indeed ease the maintenance for users and supporters.

Writing a migration task should not be that hard. There are already a bunch in the code.
The only issue that can create an arbitrary high amount of complexity is trying to solve hypothetical issues, like: What if users have AutoKey folders on disconnected external media? Burnt on CD-R or placed in read-only file systems? (But those are corner cases anyways.)

BlueDrink9:
OOoooh, that's where those files are! That should be documented somewhere! I spent ages looking in .config/autokey trying to figure out why I needed to manually add new folders.

I take it back, all that's in .local/share/autokey is the log file.

Regarding the backporting issue... We're making a lot of new changes from the legacy. At what point should we consider saying 'this is a totally new version, old stuff won't work/will need explicit conversion'?

luziferius:
@BlueDrink9 Sorry. I quoted the wrong path. The correct one is `~/.config/autokey/`.
It contains `autokey.json` and a `data` directory that is used as the default location when creating new top-level folders in AutoKey.

herrdeh:
I'm the culprit... (;
I tried to share my autokey files via owncloud on various machines and wondered why that does not work...  Being an autokey addict, it's frustrating having to set up phrases freshly on each machine you're working.

Special thanks to @ josephj11 to have opened this issue!

xpusostomos:
I added a pull request:

https://github.com/autokey/autokey/pull/430

BlueDrink9:
@josephj11 this can be closed I think

josephj11:
[...]

PR #431, opened 2020-07-13
issue337 don't use dot files

xpusostomos:
A fix for issue 337: Stop hiding the sidecar files for phrases and scripts #337

#337

json config files no longer hidden by "." in filename.

### DIFF
--- a/lib/autokey/common.py
+++ b/lib/autokey/common.py
@@ -33,7 +33,7 @@ LOCK_FILE = os.path.join(RUN_DIR, "autokey.pid")
 
 APP_NAME = "autokey"
 CATALOG = ""
-VERSION = "0.95.10"
+VERSION = "0.95.11"
 HOMEPAGE = "https://github.com/autokey/autokey"
 AUTHOR = 'Chris Dekter'
 AUTHOR_EMAIL = 'cdekter@gmail.com'
--- a/lib/autokey/configmanager/configmanager.py
+++ b/lib/autokey/configmanager/configmanager.py
@@ -348,7 +348,7 @@ class ConfigManager:
 
                 # --- handle changes to folder settings
 
-                if baseName == ".folder.json":
+                if baseName == "folder.json":
                     folder = self.__checkExistingFolder(directory)
                     if folder is not None:
                         folder.load_from_serialized()
--- a/lib/autokey/configmanager/version_upgrading.py
+++ b/lib/autokey/configmanager/version_upgrading.py
@@ -54,6 +54,8 @@ def upgrade_configuration(configuration_manager, config_data: dict):
         configuration_manager.config_altered(True)
     if version < "0.95.3":
         convert_autostart_entries_for_v0_95_3()
+    if version < "0.95.11":
+        convertDotFiles_v95_11(config_data)
     # Put additional conversion steps here.
 
 
@@ -125,3 +127,16 @@ def convert_autostart_entries_for_v0_95_3():
             old_autostart_file, new_file_name)
         )
         old_autostart_file.rename(new_file_name)
+
+def convertDotFiles_v95_11_folder(p: Path):
+    for name in p.glob('.*.json'):
+        new_json = p / name.name[1:]
+        name.rename(new_json)
+
+    for name in p.iterdir():
+        if name.is_dir():
+            convertDotFiles_v95_11_folder(name)
+
+def convertDotFiles_v95_11(configData):
+    for name in configData["folders"]:
+        convertDotFiles_v95_11_folder(Path(name))
--- a/lib/autokey/model/folder.py
+++ b/lib/autokey/model/folder.py
@@ -68,7 +68,7 @@ class Folder(AbstractAbbreviation, AbstractHotkey, AbstractWindowFilter):
         if not os.path.exists(self.path):
             os.mkdir(self.path)
 
-        with open(self.path + "/.folder.json", 'w') as outFile:
+        with open(self.path + "/folder.json", 'w') as outFile:
             json.dump(self.get_serializable(), outFile, indent=4)
 
     def get_serializable(self):
@@ -119,7 +119,7 @@ class Folder(AbstractAbbreviation, AbstractHotkey, AbstractWindowFilter):
 
     def load_from_serialized(self):
         try:
-            with open(self.path + "/.folder.json", 'r') as inFile:
+            with open(self.path + "/folder.json", 'r') as inFile:
                 data = json.load(inFile)
                 self.inject_json_data(data)
         except Exception:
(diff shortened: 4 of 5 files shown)


BlueDrink9:
Github tip for the future, if you explicitly state "close #337" or "fix #337" in your PR text, when it gets merged the issue will automatically be closed.

josephj11:
I thought this required some conversion, etc. to keep current sidecar files working. Are there any legacy issues?

BlueDrink9:
This patch includes conversion, at least from a cursory review. It probably shouldn't have been merged unless troxor reviewed it; but then we go back to our original problem.

josephj11:
Someone please review this to make sure it was ready for prime time. As mentioned elsewhere, this will change things in user files, so we need to alert them that this is intentional. Also, several people have external scripts to generate new phrases, etc. and they'll need to know that they have to change them.

BlueDrink9:
Any suggestions for how to alert people? Particularly for the scripting issue?

0.96 in general is going to be a huge update if/when it gets released, so it might just be a case of having a big changelog with a heading for incompatible/breaking changes.

Normally with semantic versioning we would change the major version number for such changes, but we haven't got an official 1.0 release so we can't do that.

xpusostomos:
Autokey already has popup messages in certain circumstances. Maybe any orphan sidecars, not just these ones, should popup a message on startup.

josephj11:
@xpusostomos Not sure what you're proposing. If someone has a few hundred
phrases and scripts, we'd certainly want only one pop-up. That's probably
what you intend. What's an orphan sidecar? One which has no corresponding
phrase or script? You might just want to ignore files like that (and
anything else that doesn't look like AutoKey files.) If I edit a script
with kate and it leaves a script.py~ file lying around, AutoKey doesn't
seem to notice. In KDE, there could also be a .desktop file lying around.

xpusostomos:
That's a good point. I kinda think it makes sense for the files to have their own .autokey extension rather than a generic .json so that one can be sure what the files are purporting to be.

And also, if I remember right, the code organises itself according to the .py and .txt scripts, and then looks for the sidecar. But to my way of thinking, it should look for the sidecar, and if necessary look for the .py and .txt. But I guess that doesn't help the immediate problem here.

josephj11:
Having a distinct extension and looking at them first makes sense, but there won't be anything else in those directories unless the user explicitly puts it there, so it kinda comes under the rubric of "You can't fix stupid."

The main point of however the code works is that it shouldn't produce any false negatives or false positives. If you find a sidecar with no corresponding main file, there's nothing you can do about it. It probably should just be ignored. If AutoKey starts deleting phrases and scripts by itself somehow, I'm sure we'll hear about it!

Apocrypha: Long ago and far away, I used WordMaster on my CP/M system. It was possible to insert an end of file character in the middle of your document (which would truncate it at that point - probably not what you wanted!) When you did that, the word "Turkey" would be displayed on your screen in reverse video for  second, but it let you do it. (You were OK until you saved the document.) The best you can do sometimes is laugh when people do stupid things. There are usually very few correct actions and nearly infinite incorrect ones.

xpusostomos:
I wonder if it wouldn't be nicer to have everything in the one file. Maybe all python, the first section would be what's now the sidecar variables, then some separator, then the action python.
Something like:
```
hotkey = Hotkey(modifiers=["<ctrl>"], key="c")
filter = Filter("... etc
#SCRIPT
keyboard.send_keys("blah")
```

Would make the code simpler too... don't even have to parse json, just execute it. From the user's point of view it would look the same.

josephj11:
Interesting idea. I think it would cause more problems than it solves.

I edit my scripts in kate. Putting the metadata in with the script would open it up to accidental corruption. The code that reads and runs it would probably have to be significantly enhanced to defend itself from relatively random possibilities.

At the very least, we'd have to do something like adding a metadata hash to the file. If it didn't match the computed hash (checked every time) when loading the file, then something other than AutoKey changed it and that would cause it to be rejected.

Now, if a user screws things up by manually editing the metadata, it's their problem,  not ours. I like that!

Most users never need to look at the metadata directly; putting it in front of them might be confusing.

It would also cause the script file to be marked as changed when only metadata has been changed. That would be confusing when debugging a script failure by making it look like something was changed in a script that might be a couple of years old. This would be even worse if we modified a field in the metadata across all scripts. Then, the modification dates on all of them would change.

If we wanted to now, we could add binary data of any type to the metadata. That's probably not the best idea, but it wouldn't even be an option if we combined the script and the metadata in one file.

xpusostomos:
I often thought resource fork file systems were a good idea, but they never took off...
* Binary data.. I presume with the sidecar you'd uuencode it or something and put it in the JSON. Under this idea you'd 
uuencode it into a string, same thing.
* Putting the meta data in front of the user.... it would only be that for the few that use external editors.
* Random corruption.. You mean like... using an external editor and corrupting the meta data? I mean, if you're going to edit stuff manually, you have to have some care anyway. You can edit the json and corrupt it too. But being Python, if you're using a python editor it would alert you to syntax errors.

But yeah, there would be some downsides definitely.

One half way proposition would be to store the config file as a separate Python file rather than a json file. I'm a big fan of using single languages for projects, and using a javascript notation in a python project, kinda seems weird to me, even though I know this format is incredibly popular. If it was Python you could even declare and share variables and stuff, though how the gui would cope with that is an open question. Maybe it would show the final evaluation and if you changed it, then you'd lose any hand-written stuff. And then  maybe an autokey collection of directories is an actual, regular python module like any other module, except that it follows a certain pattern that the gui can edit. But maybe some people don't even want the gui, they just want to code it up.

BlueDrink9:
@xpusostomos how thoroughly did you test the conversion behaviour before opening this? An issue on the gitter with the new beta makes it look like the conversion isn't happening with any of the files.

BlueDrink9:
Might be related to the version comparison, because we're on a 0.96.0 beta rather than the expected 0.95.11

BlueDrink9:
Fixed now. The folder you had been looking through wasn't populated at that point. I updated it to instead look through all the loaded folders and their paths. The only side effect I can think of is that any dotfiles associated with those folders might not get loaded until the next run.

BlueDrink9:
Actually this is an issue. (These are notes to self btw). The folder's json is loaded when the folder list is populated. If we wait for populating folders before upgrading the dotfiles, the old jsons for folders won't have been loaded.

A custom iterator over all folders would be elegant, but passing it might be another story.

The issue is just with phrases in the default folders I think, because those folders don't get added to autokey's data.

xpusostomos:
I'm not quite sure what you are referring to, but it sounds like you have it under control. I did test it, but apparently you think I missed something.

BlueDrink9:
Did you have anything in the default folder that you tested?

PR #432, opened 2020-07-13
Add wait_for_keyevent scripting function

xpusostomos:
I've set myself a project to get autokey to support emacs style keybinds.
One of the difficulties of that is supporting the C-u prefix to commands:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Prefix-Command-Arguments.html

To do that I need a scripting function that can wait for an arbitrary number of keystrokes, accumulate the result, and store it. To accomplish that I made a new scripting API wait_for_keyevent that takes a lambda/function that says when we are done, and potentially accumulates a result.

I also note that the existing wait_for_keypress API modifies the iomediator.listeners list in the middle of iterating over it. In most computer languages this is illegal and cause all sorts of issues, and non consistent behavior. In any case, it caused a problem for me because it immediately creates a call to the function with the same value. This would presumably cause a problem currently if you say had a script for C-x and then waited for keypress C-x again, it would immediately exit. Anyway, I fixed that by copying the list before iterating, which is the normal way one avoids these multi threaded iterator issues.

Documentation in the code comes with an example of how to use it to implement C-u

Regarding the implementation, I note that the develop branch has broken the pre-existing wait_for_keypress function. I've fixed it at the same time, although the code has undergone quite a bit of restructuring since master, and I'm not fully sure of the intention. It seems like the intention is that all the functionality is supposed to go via the keyboard.mediator proxy. That makes accessing the Waiter class difficult. I've added a waiter member to the iomediator to solve this, but that means Waiter can't access iomediator (circular dependency). So that means wait_for_keypress now adds itself to listeners. Feel free to give advice if there's a better way, but this works.

        Wait for a key event, potentially accumulating the intervening characters
        Usage: C{keyboard.wait_for_keypress(self, check, name=None, timeOut=10.0)}
        @param check: a function that returns True or False to signify we've finished waiting
        @param name: only one waiter can have this name. Used to prevent more threads waiting on this.
        @param timeOut: maximum time, in seconds, to wait for the keypress to occur
        Example:
        # Accumulate the traditional emacs C-u prefix arguments
        # See https://www.gnu.org/software/emacs/manual/html_node/elisp/Prefix-Command-Arguments.html
        def check(waiter,rawKey,modifiers,key,*args):
            isCtrlU = (key == 'u' and len(modifiers) == 1 and modifiers[0] == '<ctrl>')
            if isCtrlU: # If we get here, they've already pressed C-u at least 2x
                try:
                    val = int(waiter.result) * 4
                    waiter.result = str(val)
                except ValueError:
                    waiter.result = "16"
                return False
            elif any(m == "<ctrl>" or m == "<alt>" or m == "<meta>" or m == "<super>" or m == "<hyper>" for m in modifiers):
                # Some other control character is an indication we're done.
                if waiter.result is None or waiter.result == "":
                    waiter.result = "4"
                store.set_global_value("emacs-prefix-arg", waiter.result)
                return True
            else: # accumulate as a string
                waiter.result = waiter.result + key
                return False

        keyboard.wait_for_keyevent(check, "emacs-prefix")

### DIFF
--- a/lib/autokey/iomediator/iomediator.py
+++ b/lib/autokey/iomediator/iomediator.py
@@ -23,6 +23,7 @@ from autokey.model.phrase import SendMode
 
 from autokey.model.key import Key, KEY_SPLIT_RE, MODIFIERS, HELD_MODIFIERS
 from .constants import X_RECORD_INTERFACE
+from .waiter import Waiter
 
 CURRENT_INTERFACE = None
 
@@ -47,6 +48,7 @@ class IoMediator(threading.Thread):
         self.queue = queue.Queue()
         self.listeners.append(service)
         self.interfaceType = ConfigManager.SETTINGS[INTERFACE_TYPE]
+        self.waiter = Waiter
         
         # Modifier tracking
         self.modifiers = {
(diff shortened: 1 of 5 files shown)


BlueDrink9:
This is pretty neat. If you can and it makes sense, it would be great to get some tests added for this. Copy one of the existing test files, doesn't matter too much at this stage if they contain unnecessary imports etc.

PR #451, opened 2020-10-03
Fix for TriggerMode.HOTKEY (3) showing up in modes in the save file t…

sebastiansam55:
…wice.

Previously saving would add 3, and only 3, twice to the modes portion of the save file.

And fix for the preferences window not saving changes

### DIFF
--- a/lib/autokey/model/abstract_hotkey.py
+++ b/lib/autokey/model/abstract_hotkey.py
@@ -45,7 +45,7 @@ class AbstractHotkey(AbstractWindowFilter):
         modifiers.sort()
         self.modifiers = modifiers
         self.hotKey = key
-        if key is not None:
+        if key is not None and TriggerMode.HOTKEY not in self.modes:
             self.modes.append(TriggerMode.HOTKEY)
 
     def unset_hotkey(self):
--- a/lib/autokey/gtkui/settingsdialog.py
+++ b/lib/autokey/gtkui/settingsdialog.py
@@ -111,7 +111,9 @@ class SettingsDialog:
         else:
             autokey.configmanager.autostart.delete_autostart_entry()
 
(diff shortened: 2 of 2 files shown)


lib/autokey/model/abstract_hotkey.py
```diff
@@ -45,7 +45,7 @@ def set_hotkey(self, modifiers, key):
         modifiers.sort()
         self.modifiers = modifiers
         self.hotKey = key
-        if key is not None:
+        if key is not None and TriggerMode.HOTKEY not in self.modes:
```
BlueDrink9, line 48: This change specifically breaks some tests relating to creating phrases with duplicate hotkeys (which is forbidden). Normally creating duplicates raises an error, but with this change it does not. Could you please review this again @sebastiansam55

BlueDrink9 (commented):
@sebastiansam55 I forgot to be explicit when you started making PRs, but you should be running the tests please

sebastiansam55:
Noted will do so going forward!

BlueDrink9:
Thank you. Once I've fixed this bug, I'll push a note to my branch on how to add automatically running the tests as a pre-push git hook. That's how I caught these failures.

BlueDrink9:
What was the actual issue here? I think I have solved it a different way, because the way it was solved here was causing problems elsewhere and I have reverted it on my branch. I'd like to test/confirm my fix, but I'm not totally sure what the issue description was @sebastiansam55

sebastiansam55:
From what I remember;
My `.json` files started showing up with two 3's in the `modes` list. This fix made it stop for me.

I probably should have waited longer before starting to open commits; this is my first return to any form of programming for about 5 years so I probably misunderstood some stuff.

BlueDrink9:
I think the only issue was how little personpower for checking PRs we have for this repo specifically. Commits in general are welcome.

PR #452, opened 2020-10-03
Theme selection (GTKSourceView GUI)

sebastiansam55:
Add GTKSourceView theme selection to the settings file and to the gtk GUI


josephj11 (commented):
I don't know the source code, but this change confuses me. I didn't know we had a problem with changes not being saved. I thought that was fixed quite awhile ago and that the default behavior was to save changes when the window was closed - or something like that. Are we adding back a checkbox that we don't need?

sebastiansam55:
In the develop branch the preferences window wasn't saving (and throwing an error) because it was trying to access an object that didn't exist, I commented it out to finish testing the changes that I was making.

I think what was going on is that `self.promptToSaveCheckbox` was changed to `self.autosaveCheckbox` at some point? But the old variable name wasn't changed in the on_save method. I'll check the history of the file sometime or other.

But based on reading the source code, that auto save checkbox in the prefernces probably doesn't do anything in the develop branch right now because it isn't being saved.

The only change that should be needed is swapping self.promptToSaveCheckbox on line 128 to self.autosaveCheckbox (or vice versa, I'm not sure why those changes were made or which variable name would be considered "correct")

sebastiansam55:
Looking at the history of the file it looks like it is supposed to be autosaveCheckbox but got reverted from some merge idk much about reading git history logs. 

https://github.com/sebastiansam55/autokey/commit/eb8d5f35a0e98f8a7b0e5cfd73d580bf27cb3457

This commit fixes it idk how github works, should I open a new pull request?

josephj11:
OK. Ask @BlueDrink9 - maybe he will know the best way to organize the PRs.

BlueDrink9:
You can push new changes to a PR, so if you need to update it just push it to the same branch again.

PR #459, opened 2020-11-03
Change "Help/Scripting Help" link to new Github documentation link.

sebastiansam55:
Update link from older documentation to new wiki documentation


josephj11:
This is an improvement, but it is a perfect example of why our Wiki needs refactoring. Ideally, this should point to something that explains basic scripting and references our API, but we don't have anything like that now.

sebastiansam55:
This may or may not be necessary with the updated documentation. 

I would also suggest adding two different buttons, one for "API documentation" and one for "Scripting help"

josephj11:
Great idea once we have somewhere to point that new button to.

PR #461, opened 2020-11-10
Extend mouse library functionality (move cursor, press/release buttons)

sebastiansam55:
Add functions for getting mouse location, mouse button manipulation (up and down signals), ability to move cursor on screen via warp or "tracking" (+/- x +/- y every 0.001 second), method of "drag and select".


sebastiansam55:
Oh well... I see that some of this has been done in highlevel.py ... using `subprocess` and `xte`. Might still be valuable?  

`highlevel.py` also uses different documentation conventions in a quick look at it. Will probably update the documentation there and put in a reference to those functions existing in the `mouse.py` documentation at a minimum.

sebastiansam55:
```
mouse.move_cursor(0,0)
mouse.press_button(1)
mouse.move_cursor(300,300)
mouse.release_button(1)
```
Works to highlight the area it indicates.
`mouse.select_area(0,0,300,300,1)` is equivalent to the above

`mouse.get_location()` has some caveats, it returns the location of the cursor instantly, so like I poorly explain in the documentation string if your script does something like;
```
mouse.move_cursor(0,0)
x,y = mouse.get_location()
```
The values of x,y will not be 0,0 and instead be where ever your cursor currently is. If you put a `time.sleep()` of almost any time you'll get the "correct" return of 0,0.

This is because of the way that the action queuing works, someone might be able to write it so that `get_location` function is added to the queue but that might also end up in unexpected behavior.

josephj11:
What would happen if mouse.get_location() just incorporated a slight delay at the beginning?

Note that, in some situations, the more concise form is not the equivalent of the detailed version because the lower level approach allows other code to be interspersed in the sequence of events and the underlying application window may respond to the individual events as they occur if it is given a little time (delays...) to do so.

sebastiansam55:
Added a minimal delay (0.05) which seems to fix the issue on my end to `mouse.get_location`. 

Added `mouse.scroll_down` and `mouse.scroll_up` along with matching `interface.py` methods. 

Optional flags added to `mouse.select_area`:
* `scrollNumber` - number of times to trigger the scroll button signal. By default is set to 0.
* `down` - True by default, set to false to scroll up.
* `warp` - True by default, set to False if you don't want the cursor to warp to cursor position at start of method execution

josephj11:
This keeps getting better! I'll merge it when you feel it's stable.

If warp is false and the mouse cursor is not already at the initial coordinates, what does it do? I assume it starts at the "start", ends at the "end" and just doesn't go back to where it started. So I could move over a button and then do nothing if I wanted to "trigger" the display of a tooltip.

sebastiansam55:
Added methods:
* `mouse.move_relative` - Moves (warps instantly for all methods) the mouse cursor to an x,y coordinate location relative to the upper left hand corner of the currently active window (that has input focus)
* `mouse.move_relative_self` - Moves the mouse cursor to an x,y coordinate relative to the current location of the mouse cursor itself
* `mouse.get_relative_location` - Returns the x,y coordinate location relative to the upper left hand corner of the currently active window.

Added debugging line for `handle_mouseclick` in `service.py`. 

Added file `model/button.py` which defines the mouse buttons recognized by X11, which can be accessed via `mouse`:

```
mouse.move_cursor(100,100)
mouse.press_button(mouse.Button.LEFT)
time.sleep(1)
mouse.release_button(mouse.Button.LEFT)
```

sebastiansam55:
I'm good with these changes being merged with the addition of the `-m`/`--mouse` flag, as otherwise verbose log output becomes harder to scroll through (as scroll button events are logged whenever scrolling).

PR #463, opened 2020-11-10
Updated documentation for develop branch, focusing on the scripting m…

sebastiansam55:
…odule. 

Simple documentation additions to fill out the auto generated pydoctor pages.

### DIFF
--- a/lib/autokey/scripting/clipboard_gtk.py
+++ b/lib/autokey/scripting/clipboard_gtk.py
@@ -12,6 +12,9 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+GtkClipboard Functions
+"""
 
 from gi.repository import Gtk, Gdk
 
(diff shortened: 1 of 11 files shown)


BlueDrink9:
This looks great, thanks for this work.

One thing that feels funny: IIRC, functions beginning with `___` are generally considered 'internal' methods, so shouldn't need documentation. It isn't enforced by Python, just a standard. Docs generators shouldn't generate docs for them. If the docs are adding something useful to the code itself, great, but otherwise can and should probably strip those out IMO.

sebastiansam55:
I think that in the new documentation generator I proposed adds them hidden behind a click to reveal internal methods. I was erring on the side of more documentation instead of less.

https://sebastiansam55.github.io/autokey-develop/scripting.clipboard_gtk.GtkClipboard.html

See above, there is a "show private API button" I don't think that we ever reached a final decision on changing documentation generation.

I'll remove if you still want, just wanted to throw my reasoning out there

BlueDrink9:
Oh nice, may as well leave it in then. My gut is that internal stuff only needs in-source docs, but since it's there we may as well keep it.

@josephj11 I see no problem with merging this now.

PR #467, opened 2020-11-12
Fix for hotkeys still being applied when cleared (my fault `not` wher…

sebastiansam55:
…e there shouldn't have been one)

### DIFF
--- a/lib/autokey/model/abstract_hotkey.py
+++ b/lib/autokey/model/abstract_hotkey.py
@@ -45,7 +45,7 @@ class AbstractHotkey(AbstractWindowFilter):
         modifiers.sort()
         self.modifiers = modifiers
         self.hotKey = key
-        if key is not None and TriggerMode.HOTKEY not in self.modes:
+        if key is not None and TriggerMode.HOTKEY in self.modes:
             self.modes.append(TriggerMode.HOTKEY)
 
     def unset_hotkey(self):


BlueDrink9:
@josephj11 good to merge

BlueDrink9:
Probably on my head for not cloning and checking more thoroughly, but I was expecting @sebastiansam55 to have run tests before pushing. This commit breaks 4 tests. I'm looking into why now.

I've added a note to the CONTRIBUTERS.rst file on my `develop` branch suggesting creating a `pre-push` git hook to run them all. On my machine it takes about 10 seconds to run the full suite, so it's worth doing.

josephj11:
I could revert it if desired, but you'd have to tell me how to do that.

BlueDrink9:
I'm sorting out the underlying reason why these two PRs are causing issues. This isn't the only one; #451 is also causing issues. I'm not sure we need to just yet, since it's develop and the tests are related to new features no-one seems to be using. I'll fix it in another PR.

I see github is offering a 'revert' option beside the merge commit on this PR, if we need it.

lib/autokey/model/abstract_hotkey.py
```diff
@@ -45,7 +45,7 @@ def set_hotkey(self, modifiers, key):
         modifiers.sort()
         self.modifiers = modifiers
         self.hotKey = key
-        if key is not None and TriggerMode.HOTKEY not in self.modes:
+        if key is not None and TriggerMode.HOTKEY in self.modes:
```
BlueDrink9, line 48: What is the intended behavior here? Currently your 'fix' duplicates the HOTKEY triggermode by adding it only if it already exists, is that intentional?

BlueDrink9: @sebastiansam55

Issue #306, opened 2019-08-29
window.get_active…()—provide unique window ID to accurately target specifc window via resize_move()


mmhere:
## Classification:

Feature (New) / Enhancement

## Reproducibility:

Always

## Version

AutoKey version: 0.95.7

Used GUI (Gtk, Qt, or both): GTK (likely same in Qt)

Installed via:
dpkg --install autokey-common_0.95.7-0_all.deb autokey-gtk_0.95.7-0_all.deb

Linux Distribution:  Ubuntu 16.04

## Summary

window.get_active…()—provide unique window ID to accurately target specifc window via resize_move()

### API tweak request

This script API method...

    thisWin = window.get_active_title() # returns string

...returns only a window title as a string.

This makes it impossible to control specific window(s) with this...

    window.resize_move(thisWin, ...) # assumes thisWin is a string

...when window titles are not unique across the set of all open windows.

As wmctrl is used as the underlying method of window control (according to docs), and that tool accepts a specific window ID (hexadecimal or integer), it would be very nice to access said number that is unique to each window.  Example:

    wmctrl -i -a 0x0a000003
    # alternate:
    wmctrl -i -a 167772163

In autokey Python scripts, you really need to be able to do this:

    thisWinID = window.get_active_id() # give me a unique number!
    window.resize_move(thisWinID, ...) # then I can move that specific window!

Why?

In my situation, there are *usually* multiple windows with the same title!  My style is to lean on the X11 window manager to manage many unique windows, as opposed to leaning on lots of tabs cluttered within windows (I know this fad is popular right now).  In an application such as the Vivaldi browser, I often have several windows with the same title "Vivaldi".

When window.get_active_title() returns "Vivaldi", but there are, say, three such windows, you may not actually control the *active* (top) window with a subsequent call on window.resize(...).  You get whichever one wmctrl decides to select first out of the group of windows with identical titles.

### Suggested Tags

As I do not have write access, I am unable to apply Labels to this issue.  I suggest these:

- enhancement
- Scripting

josephj11:
While this is a valid request, as a workaround, it's not particularly difficult to call `xdotool` or `wmctrl` directly from your script to access any capabilities they provide which AutoKey does not.

`xdotool` returns stacks of window IDs matching various search conditions. I've played with it a bit, but really don't understand it very well. Restricting it to only find visible windows helps a bit.

[...]

PR #471, opened 2020-11-17
Add more functions to window API

sebastiansam55:
Adds `get_window_geom`, `get_window_hex` and `get_window_list` to the window API. Allows for use of other commands like `import -window <hexid>` for getting screenshots of windows (for upcoming visgrep changes).

### DIFF
--- a/lib/autokey/scripting/window.py
+++ b/lib/autokey/scripting/window.py
@@ -27,6 +27,7 @@ class Window:
     Note: in all cases where a window title is required (with the exception of wait_for_focus()),
     two special values of window title are permitted:
 
+    #TODO should these be broken out into enums like keys/buttons
     :ACTIVE: - select the currently active window
     :SELECT: - select the desired window by clicking on it
     """
@@ -180,7 +181,9 @@ class Window:
 
         Usage: C{window.set_property(title, action, prop, matchClass=False)}
 
+        #TODO should these be broken out into enums like keys/buttons
         Allowable actions: C{add, remove, toggle}
+        #TODO should these be broken out into enums like keys/buttons
         Allowable properties: C{modal, sticky, maximized_vert, maximized_horz, shaded, skip_taskbar,
         skip_pager, hidden, fullscreen, above}
 
@@ -240,14 +243,15 @@ class Window:
         """
         return self.mediator.interface.get_window_class()
 
-    def center_window(self, win_width=None, win_height=None, title=":ACTIVE:", monitor=0):
+    def center_window(self, title=":ACTIVE:", win_width=None, win_height=None, monitor=0, matchClass=False):
         """
         Centers the active (or window selected by title) window. Requires xrandr for getting monitor sizes and offsets.
 
+        @param title: Title of the window to center (defaults to using the active window)
         @param win_width: Width of the centered window, defaults to screenx/3. Use -1 to center without size change.
         @param win_height: Height of the centered window, defaults to screeny/3. Use -1 to center without size change.
-        @param title: Title of the window to center (defaults to using the active window)
         @param monitor: Monitor number (0 is primary, listed via C{xrandr --listactivemonitors} etc.)
+        @param matchClass: if True, match on the window class instead of the title
         @raises ValueError: If title or desktop is not found by wmctrl
         """
         #could also use Gdk.Display.get_default().get_montiors etc.
@@ -258,13 +262,13 @@ class Window:
         # 1: +DVI-D-0 1440/408x900/255+1920+0  DVI-D-0
         # Regex gets  ____ and ____    ____ _
         # this translates to monitor x and y size, and x and y offset for each monitor
-        rcode, output = self._run_xrandr(["--listactivemonitors"])
+        returncode, output = self._run_xrandr(["--listactivemonitors"])
         regex = r" (\d{3,4}).*?x(\d{3,4})\/.*?\+(\d{1,4})\+(\d{1,4})"
         matches = re.findall(regex, output, re.MULTILINE)
-        x_offset = int(matches[monitor][2])
-        y_offset = int(matches[monitor][3])
-        width = int(matches[monitor][0])
-        height = int(matches[monitor][1])
+
+        width, height = [int(i) for i in matches[monitor][0:2]]
+        x_offset, y_offset = [int(i) for i in matches[monitor][2:4]]
+
         #work backwards from the size of the window
         if win_width is None:
             win_width = round(width/2)
@@ -272,11 +276,12 @@ class Window:
             win_height = round(height/2)
         top_x = round((width-win_width)/2)
         top_y = round((height-win_height)/2)
-        xrandr_size = ",".join(["0", str(x_offset+top_x), str(top_y), str(win_width), str(win_height)])
-        print(["-r", title, "-e", xrandr_size])
+
         #remove maximized attributes from window
-        returncode, output = self._run_wmctrl(["-r", title, "-b", "remove,maximized_vert,maximized_horz"])
-        returncode, output = self._run_wmctrl(["-r", title, "-e", xrandr_size])
+        self.set_property(title, "remove", "maximized_vert", matchClass=matchClass)
+        self.set_property(title, "remove", "maximized_horz", matchClass=matchClass)
+        #resize and move window
+        self.resize_move(title, x_offset+top_x, y_offset+top_y, win_width, win_height, matchClass=matchClass)
 
     def _run_xrandr(self, args):
         try:
@@ -309,24 +314,28 @@ class Window:
         C{desktop} is the number of which desktop (sometimes called workspaces) the item appears upon.
 
         C{hostname} is the hostname of your computer (I'm not sure why this is included as it seems to be the same for everything I have seen)
-        
+
         C{title} is the title that you would usually see in your window manager of choice.
 
         @param filter_desktop: String, (usually 0-n) to filter the windows by. Any window not on the given desktop will not be returned.
         @return: C{[[hexid1, desktop1, hostname1, title1], [hexid2,desktop2,hostname2,title2], ...etc]}
         Returns C{[]} if no windows are found.
         """
-        windowList = self._run_wmctrl(["-l"])[1].split("\n")
-        output = []
-        for window in windowList:
-            info = window.split("  ")[1] #the remaining information
-            hexid = window.split("  ")[0] #the hexid of the window
-            desktop, hostname, window_title = info.split(" ", 2)
-            if filter_desktop is not None:
+        returncode, output = self._run_wmctrl(["-l"])
+        # regex = r"^(0x[0-9a-fA-F]{8})  (\d*) (.*?) (.*?)$"
+        regex = r"^(0x[0-9a-fA-F]{8})\s*(\d*)\s*(.*?)\s(.*?)$"
+        matches = re.findall(regex, output, re.MULTILINE)
+        print(matches)
+        if filter_desktop is None:
+            return matches
+        else:
+            output = []
+            for match in matches:
+                hexid, desktop, hostname, window_title = match
                 if not filter_desktop==desktop:
                     continue
-            output.append([hexid,desktop,hostname, window_title])
-        return output
+                output.append((hexid,desktop,hostname, window_title))
+            return output
 
     def get_window_hex(self, window_title):
         """
@@ -335,7 +344,7 @@ class Window:
         @param window_title: Window title to match for returning hexid
         @return: Returns hexid of the window to be used for other functions See L{get_window_geom}, L{visgrep_by_window_id}
 
-        Returns none if no matches are found
+        Returns C{None} if no matches are found
         """
         windowList = self.get_window_list()
         for window in windowList:
@@ -355,15 +364,10 @@ class Window:
         @return: C{[offsetx, offsety, sizex, sizey]} Returns none if no matches are found
         """
 
-        windowList = self._run_wmctrl(["-lG"])[1].split("\n")
-        output = []
-        for window in windowList:
-            w_split= window.split("  ")
-            hexid = w_split[0]
-            geom1 = w_split[1].split(" ")[1:]
-            geom2 = w_split[2].split(" ")
-
-
-            if hexid==window_id:
-                return [geom1+geom2]
+        returncode, output = self._run_wmctrl(["-lG"])
+        regex = r"^(0x[0-9a-fA-F]{8})\s*(\d*)\s*(\d*)\s*(\d{1,})\s*(\d{1,})\s*(\d{1,})\s*(.*?)$"
+        matches = re.findall(regex, output, re.MULTILINE)
+        for match in matches:
+            if match[0] == window_id:
+                return [int(i) for i in match[2:-1]] # trim hexid, gravity on the front end and title on the backend
         return None
--- a/lib/autokey/scripting/window.py
+++ b/lib/autokey/scripting/window.py
@@ -249,3 +249,73 @@ class Window:
             return 1, 'ERROR: Please install wmctrl'
 
         return returncode, output
+
+    def get_window_list(self, filter_desktop=None):
+        """
+        Returns a list of windows matching an optional desktop filter, requires C{wmctrl}!
+
+        Each list item consists of: C{[hexid, desktop, hostname, title]}
+
+        Where the C{hexid} is the ID used for some other functions (like L{import -window} from ImageMagick).
+
+        C{desktop} is the number of which desktop (sometimes called workspaces) the item appears upon.
+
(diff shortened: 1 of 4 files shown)


lib/autokey/scripting/window.py
```diff
@@ -247,3 +247,73 @@ def _run_wmctrl(self, args):
...
+        C{title} is the title that you would usually see in your window manager of choice.
+
+        @param filter_desktop: String, (usually 0-n) to filter the windows by. Any window not on the given desktop will not be returned.
+        @return: C{[[hexid1, desktop1, hostname1, title1], [hexid2,desktop2,hostname2,title2], ...etc]}
+        Returns C{[]} if no windows are found.
+        """
+        windowList = self._run_wmctrl(["-l"])[1].split("\n")
+        output = []
+        for window in windowList:
+            info = window.split("  ")[1] #the remaining information
```
BlueDrink9: Tidy up the inside of this loop with a function maybe? Something that just returns the list of useful info?

You can use `@staticmethod` above the function declaration to avoid having to include API documentation, I think.

If you're really keen, getting some tests written where you can would be great. Even if you mock out the external bits to return the expected results, that would still be super useful.

BlueDrink9 (changes requested):

lib/autokey/scripting/window.py
sebastiansam55: I may have brought it up elsewhere but I think we should probably swap from using wmctrl to internal implementation because implementing reading output from the commandline is always going to be a pain to implement (and be easily broken if `wmctrl` changes anything about their output)

I say this understanding that the main thing we are lacking atm is developer time and re implementing existing features is not the best use of time

BlueDrink9: yeah, reasonable point. One stop-gap to cover wmctrl changing output is to create some unit tests that assert its interface is what we expect it to be. Have you had a go with the testing system yet? It's very liberating once you get used to it.

BlueDrink9: I've changed my mind about re-writing this loop (so ignore my comment about `@staticmethod`). Would still be nice to have some tests though.

I think the easiest way to start is to create `tests/scripting_api/test_window.py`. Copy the contents of `tests/scripting_api/test_macro.py` to get the relevant setup methods and imports, etc. (Feel free to change the copyright notice at the top to your name/username and email).

The [mock objects](https://docs.python.org/3/library/unittest.mock.html) and [patches](https://docs.python.org/3/library/unittest.mock.html#patch) that you can create are very useful. In this case, you might want a test body like 

```python
get_win_details = MagicMock(return_value=test_string)
with patch("self._run_wmctrl", get_win_details):
    output = engine.get_window_list()
    assert_equal(output, expected_output)
```

Where `test_string` is an example of what you might expect `self._run_wmctrl(["-l"])` to return in a real system, and `expected_output` is what you expect the function to return.

Create one of these for now, and when you've got the hang of it, I'll show you how to test multiple test strings in a single test. (Or play with the `@parameterize` decorator if you want to learn yourself).

Each test you write is a function in the file that starts with `test_`.

sebastiansam55:
Rebased to develop and added center_window convenience method.

BlueDrink9:
For a "convenience" method, it's a pretty long method! It's longer than most of the other top-level methods you've added so far. It would be good to come back and refactor it.

Also let me know how you are getting on with tests. Feel free to push them, either here or on a branch on your fork, if you want me to review them or help out

sebastiansam55:
True, convenience more in the sense I wanted some way to center windows and it was surprisingly hard to find modern methods to accomplish this without writing code, even this method is still 100% reliant on wmctrl and xrandr (most online solutions recommend compiz or digging into dconf) But as long as I was going to write code for it it might as well be in AutoKey so others can take advantage of it.

RE testing;
I've taken a few swings at writing some and am making more process than I have previously! Your comments on what I'm doing are more than welcome, I'm sure that I'm not following all of the coding conventions around tests...

Also I annotated a few places in the `window.py` file that seem like they might make more sense as an enum like those used in `button.py` or `key.py`. More specifically the special methods for window selection (`":ACTIVE:", ":SELECT:"`) (line 30 in window.py) and the properties and actions wmctrl offers, as seen described in the comment for the `set_property` method (line 184ish)

BlueDrink9:
I'm not sure how helpful I can be in terms of testing conventions; I'm still learning them myself. But I will try my best to at least pull you up to where I'm at!

BlueDrink9:
The Enum question is an interesting one. It seems like a good thing in general, but I think this is a special case because these are public API methods. We want them to be callable without anything extra being imported, so it seems like using strings instead of enums is reasonable here. If they were being used internally (which they shouldn't be?) then I would agree. Internal use = enum, but external/docs should use the equivalent string.

lib/autokey/scripting/window.py
```diff
@@ -258,25 +262,26 @@ def center_window(self, win_width=None, win_height=None, title=":ACTIVE:", monit
...
+        returncode, output = self._run_xrandr(["--listactivemonitors"])
         regex = r" (\d{3,4}).*?x(\d{3,4})\/.*?\+(\d{1,4})\+(\d{1,4})"
         matches = re.findall(regex, output, re.MULTILINE)
-        x_offset = int(matches[monitor][2])
-        y_offset = int(matches[monitor][3])
-        width = int(matches[monitor][0])
-        height = int(matches[monitor][1])
+
+        width, height = [int(i) for i in matches[monitor][0:2]]
+        x_offset, y_offset = [int(i) for i in matches[monitor][2:4]]
```
BlueDrink9: This is way less clear than the previous version. Don't hide details under complicated python syntax just to save a few lines, IMO. Especially in a project like this, where maintainers may not know python well.

lib/autokey/scripting/window.py
```diff
@@ -309,24 +314,28 @@ def get_window_list(self, filter_desktop=None):
...
-        windowList = self._run_wmctrl(["-l"])[1].split("\n")
-        output = []
-        for window in windowList:
-            info = window.split("  ")[1] #the remaining information
-            hexid = window.split("  ")[0] #the hexid of the window
-            desktop, hostname, window_title = info.split(" ", 2)
-            if filter_desktop is not None:
+        returncode, output = self._run_wmctrl(["-l"])
+        # regex = r"^(0x[0-9a-fA-F]{8})  (\d*) (.*?) (.*?)$"
+        regex = r"^(0x[0-9a-fA-F]{8})\s*(\d*)\s*(.*?)\s(.*?)$"
```
BlueDrink9: Can you comment this regex?

Also, for regexes especially, tests are great for two reasons:
1. Easy enough to write, just have a bunch of examples and ensure it matches them and not others. Check `test_macro` for examples of regex tests.
2. Acts as a way to self-document exactly what the regex does.

lib/autokey/scripting/window.py
```diff
@@ -309,24 +314,28 @@ def get_window_list(self, filter_desktop=None):
         C{desktop} is the number of which desktop (sometimes called workspaces) the item appears upon.
 
         C{hostname} is the hostname of your computer (I'm not sure why this is included as it seems to be the same for everything I have seen)
```
BlueDrink9: This is inappropriate in a docstring comment. Stick it outside the docstring.

As to why it is included: No idea

lib/autokey/scripting/window.py
```diff
@@ -355,15 +364,10 @@ def get_window_geom(self, window_id):
...
-            geom1 = w_split[1].split(" ")[1:]
-            geom2 = w_split[2].split(" ")
-
-
-            if hexid==window_id:
-                return [geom1+geom2]
-        return None
\ No newline at end of file
+        returncode, output = self._run_wmctrl(["-lG"])
+        regex = r"^(0x[0-9a-fA-F]{8})\s*(\d*)\s*(\d*)\s*(\d{1,})\s*(\d{1,})\s*(\d{1,})\s*(.*?)$"
```
BlueDrink9: Write a comment that breaks this regex down into its components please. Also things like examples of input it expects, and example of match/output. (Doing this in tests would also be a good idea; point people to the test in a comment here).

Also consider making your regexes specific constants, [like is done elsewhere in the codebase](https://github.com/autokey/autokey/blob/4b87b32452cd41d2b98341db220e99cf4f9e34d6/lib/autokey/model.py#L35)

lib/autokey/scripting/window.py
```diff
@@ -355,15 +364,10 @@ def get_window_geom(self, window_id):
...
+        matches = re.findall(regex, output, re.MULTILINE)
+        for match in matches:
+            if match[0] == window_id:
+                return [int(i) for i in match[2:-1]] # trim hexid, gravity on the front end and title on the backend
```
BlueDrink9: What do you mean by gravity, front end, back end? Can you give an example in this comment? (also put it the line above, not inline)

tests/scripting_api/test_window.py
```diff
@@ -0,0 +1,162 @@
...
+    (None, (0, xrandr_raw_single), False), #single monitor
+    (None, (0, xrandr_raw_double), False), #double monitor
+    (":ACTIVE:", (0, xrandr_raw_double), False),
+    ("AutoKey", (0, xrandr_raw_double), False),
+])
+def test_window_center_window(title, xrandr_raw_output, matchClass):
+    window = create_window()
+    xrandr = MagicMock(return_value = xrandr_raw_output)
+    with patch("autokey.scripting.window.Window.set_property") as set_prop:
+        with patch("autokey.scripting.window.Window.resize_move") as resize_move:
```
BlueDrink9: [Don't need multiple `with`s](https://github.com/autokey/autokey/blob/29a8b86ecc0ed3e210b1845e2599b94a305aaf03/tests/engine_helpers.py#L43)

BlueDrink9 (changes requested):

tests/scripting_api/test_window.py
```diff
@@ -0,0 +1,162 @@
...
+    with patch("autokey.scripting.window.Window._run_wmctrl", wmctrl):
+        assert func_output == window.get_window_list()
+
+
+@pytest.mark.parametrize("title, func_output, get_window_list_output",[
+    ("AutoKey", "0x07200007", [("0x07200007", "0", "samdesktop", "AutoKey"),("0x02000007", "0", "samdesktop", "Desktop")]),
+])
+def test_window_get_window_hex(title, func_output, get_window_list_output):
+    #TODO there is probably some way to not have to write this at the top of every function
+    window = create_window()
```
BlueDrink9, line 177: There is a way, but it isn't much difference because you still have to pass it in as an argument. ([Example here](https://github.com/autokey/autokey/blob/29a8b86ecc0ed3e210b1845e2599b94a305aaf03/tests/scripting_api/test_engine.py#L82) [fixture defined here](https://github.com/autokey/autokey/blob/29a8b86ecc0ed3e210b1845e2599b94a305aaf03/tests/engine_helpers.py#L35)) This is fine as a one-liner.

That being said, I'm not an expert so I might have missed a better way to do it. Pytest's docs on fixtures are a little confusing.

BlueDrink9 (commented):
The tests you have there are looking really good! I'm impressed and heartened, they're definitely better than my early attempts.

BlueDrink9:
Rebased atop develop on your behalf and force-pushed after merging of #494

lib/autokey/scripting/window.py
```diff
@@ -249,3 +302,72 @@ def _run_wmctrl(self, args):
...
+        """
+        returncode, output = self._run_wmctrl(["-l"])
+        # regex = r"^(0x[0-9a-fA-F]{8})  (\d*) (.*?) (.*?)$"
+        regex = r"^(0x[0-9a-fA-F]{8})\s*(\d*)\s*(.*?)\s(.*?)$"
+        matches = re.findall(regex, output, re.MULTILINE)
+        print(matches)
+        if filter_desktop is None:
+            return matches
+        else:
+            output = []
```
BlueDrink9: This branch (`filter_desktop is not None`) doesn't currently have test coverage. Would be nice to get something in here.

lib/autokey/scripting/window.py
```diff
@@ -249,3 +302,72 @@ def _run_wmctrl(self, args):
...
+        windowList = self.get_window_list()
+        for window in windowList:
+            if window_title in window[3]:
+                return window[0]
+        return None
+
+
+
+
+    def get_window_geom(self, window_id):
```
BlueDrink9: How does this function compare to `get_active_geometry` (line 201)? Can you please deduplicate? (Would be nice to have your tests converted for `get_active_geometry` as well)

lib/autokey/scripting/window.py
sebastiansam55: Gravity is a weird thing that is in the X11 window manager spec but appears to not be used/respected by at least the WM I use (whatever the Ubuntu 20.04 defaults to).

You can read more about it here https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#Window%20Geometry

From reading about it, gravity seems like it would be an ideal way to center windows using `wmctrl`, my understanding is that it changes the "anchor" point where the window is placed, defaulting to the top left hand corner as one might expect, and each different value choosing a different point on either the midpoints or corners of the sides or the center itself

I may be reading it wrong, there are lots of words there lol. There seems to be some difference between the "client window" and the "frame window" but I'm not sure what it is referring to. 

I would include gravity in the return values for this function, as it does seem relevant to window geometry but the pre-existing `get_active_geometry` only returns the offset x,y and width/height. So for consistency I choose to discard it.

Not aware of anyone ever asking/wondering where it is, and from my googling of it it does not come up a lot.

sebastiansam55: It should be considered that the two methods (`get_window_geom` and `get_active_geometry`) are more or less equivalent, I'd advocate that we should make `get_active_geometry` as deprecated and change up `get_window_geom` to accomplish the same thing (would get the active window by default, and take title/hex id as optional args)

The two functions are doing essentially the same thing (calling wmctrl and parsing the output), just `get_active` is limited to only getting the active window's geometry obviously.

tests/scripting_api/test_window.py
sebastiansam55: Yeah I figured that there had to be some way to do it but was not aware of the syntax, neat!

josephj11:
Just wondering: Is there a character class for hex? If there is, would it be better to use in the regexes in case someone's strange locale's collating sequence contains some surprises?

sebastiansam55:
> Just wondering: Is there a character class for hex? If there is, would it be better to use in the regexes in case someone's strange locale's collating sequence contains some surprises?

I'm not sure I did try `[[:xdigit:]]` but python didn't like it and according to the website I was using (regex101.com):
```
[[:xdigit:]]

Matches hexadecimal digits, case insensitive. Equivalent to [0-9a-fA-F].
```

lib/autokey/scripting/window.py
BlueDrink9: Rather than deprecate, how about you just convert it into a wrapper than calls your func?

BlueDrink9:
> but python didn't like it and according to the website I was using (regex101.com)

[Python uses similar regex to perl flavour](https://docs.python.org/3/library/re.html#module-re), which is likely to be relevant when it comes to choosing flavour of regex on the website.

sebastiansam55:
> > but python didn't like it and according to the website I was using (regex101.com)
> 
> I suggest

I think you may have sent this message before completing it haha

sebastiansam55:
@BlueDrink9 
* I see in the original `get_active_geometry` there are the following lines; (https://github.com/autokey/autokey/blob/develop/lib/autokey/scripting/window.py#L216)

```
# return [int(x) for x in output]
return list(map(int, output))
```

They seem to be equivalent, and I use the commented out method in other methods contained in this commit, should I convert them to use the other "newer" (presumably) method?

* Also in `_run_wmctrl` the newline is stripped via character indexing, instead of something like `.strip()` should this be updated to move away from.

* I would also like to say that I might add a `by_hex` flag to many (if not all of the existing functions) in order to address the issues raised in issue #306 . Because they are right, the way it works right now if there is umore than one window with the same name you'd only ever be able to select/act upon the "first" one in the list. 

The flag would be set by default to false in order to maintain backward compatibility of course.

BlueDrink9:
Yes, your points are all good ones. I think `return list(map(int, output))` is the better way, because it's easier to look up what `map` does than Google the name for list comprehension's syntax.

tests/scripting_api/test_window.py
BlueDrink9: Might be able to use the example I linked, but make it an [autouse fixture](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request) (`@pytest.fixture(autouse=True)`)

BlueDrink9:
Great round of changes, nice work

BlueDrink9 (approved):

PR #482, opened 2020-11-28
Develop json link

sebastiansam55:
Changes pictured below. Adds a link to the JSON config file along side the original link to the file itself. Unlike the `*.txt` or `*.py` it only displays the filename instead of the full path in order to save space.

Implemented because I was sick of navigating manually to the config files when doing debugging.

Screenshot from 2020-11-27 19-15-29
Screenshot from 2020-11-27 19-15-37
Screenshot from 2020-11-27 19-15-43

### DIFF
--- a/lib/autokey/gtkui/configwindow.py
+++ b/lib/autokey/gtkui/configwindow.py
@@ -62,7 +62,7 @@ PROBLEM_MSG_SECONDARY = _("%s\n\nYour changes have not been saved.")
 from .configwindow0 import get_ui
 
 
-def set_linkbutton(button, path):
+def set_linkbutton(button, path, filename_only=False):
     button.set_sensitive(True)
 
     if path.startswith(cm_constants.CONFIG_DEFAULT_FOLDER):
@@ -70,7 +70,10 @@ def set_linkbutton(button, path):
     else:
         text = path.replace(os.path.expanduser("~"), "~")
 
-    button.set_label(text)
+    if filename_only:
+        button.set_label(path[path.rindex("/")+1:])
+    else:
+        button.set_label(text)
     button.set_uri("file://" + path)
     label = button.get_child()
     label.set_ellipsize(Pango.EllipsizeMode.START)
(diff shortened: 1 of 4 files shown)


lib/autokey/gtkui/configwindow.py
```diff
@@ -62,15 +62,18 @@
...
     button.set_sensitive(True)
 
     if path.startswith(cm_constants.CONFIG_DEFAULT_FOLDER):
         text = path.replace(cm_constants.CONFIG_DEFAULT_FOLDER, _("(Default folder)"))
     else:
         text = path.replace(os.path.expanduser("~"), "~")
 
-    button.set_label(text)
+    if filename_only:
+        button.set_label(path[path.rindex("/")+1:])
```
BlueDrink9: I feel like we should be using actual path parsing libraries here. This *might* do for now, but it smells awful. Change it if you can.

I suspect the fact that we aren't importing them may be a hangover from them not being in python2, perhaps.

lib/autokey/gtkui/data/phrasepage.xml
```diff
@@ -12,15 +12,43 @@
...
+                <property name="relief">none</property>
+              </object>
+              <packing>
+                <property name="expand">True</property>
+                <property name="fill">True</property>
+                <property name="position">0</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkLinkButton" id="linkButton1">
```
BlueDrink9: Can we please use a name other than `linkButton1` for the id? Way too hard to know what it refers to elsewhere

BlueDrink9 (changes requested):
Seems mostly fine to me, small enough that @josephj11 could merge it. However, there are just two changes I'd like to see first, if possible. Esp renaming that id.

lib/autokey/gtkui/configwindow.py
sebastiansam55: Not sure what you mean by "actual path parsing libraries" I have switched it to use `os.path.basename`:
```
    if filename_only:
        filename = os.path.basename(path)
        button.set_label(filename)
```

Which definitely looks much better

BlueDrink9: Yup, I think that's what I had in mind; either that or create Path objects. This is indeed much better.

BlueDrink9:
@josephj11 this looks good to me. Let's merge.

PR #484, opened 2020-11-29
fix #483

### LINKED ISSUE #483: #354 fix causes IndexError: string index out of range
## Classification:

Bug

## Reproducibility:

``` python
```

## Version

AutoKey version: 0.95.10

Used GUI (Gtk, Qt, or both): 

## Summary

## Expected Results

Nothing happens.

## Actual Results

``` python
    exec(script.code, scope)
    if output[-1] == "\n":
```

fx-kirin:
https://github.com/autokey/autokey/issues/483

### DIFF
--- a/lib/autokey/scripting.py
+++ b/lib/autokey/scripting.py
@@ -90,16 +90,16 @@ class Keyboard:
     """
     Provides access to the keyboard for event generation.
     """
-    
+
     def __init__(self, mediator):
         self.mediator = mediator
-        
+
     def send_keys(self, keyString):
         """
         Send a sequence of keys via keyboard events
-        
+
         Usage: C{keyboard.send_keys(keyString)}
-        
+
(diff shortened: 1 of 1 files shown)


sebastiansam55:
We don't really have active knowledgable developers, so we are currently trying to avoid committing anything to master unless it has been approved by someone with knowledge of both autokey and python in general.

It also sort of looks like your text editor is messing with the line endings or something I don't know that we even have a defined code style aside from 4 spaces instead of tabs. So if you could reopen this on the PR without the whitespace changes the main code change looks great!

Thanks for contributing!!

fx-kirin:
Oh, sorry about it. While I was just reinstalling my OS and environment, I found this bug. So I tried to make a patch but I didn't have proper development environment. Thank you for your review! I'll reopen the PR to make it clear.

Issue #462, opened 2020-11-10
Errors started coming after updating Linux Mint Cinnamon 19.3 -> 20


mikkovedru:
## Classification:
Bug

## Reproducibility:
Always

## Version
AutoKey version: 0.95.10
Used GUI (Gtk, Qt, or both): both
Installed via: PPA
Linux Distribution: Linux Mint Cinnamon 20
Python (`$ python --version`): Python 3.6.4

## Summary

I have a simple script that opens a directory in the file manager with a hotkey. Example script:
```
# Enter script code
system.exec_command("nemo /media/mikko/Files/Download")
```

In Linux Mint Cinnamon 19.3 (and who knows how many versions before that, because I have been using Mint Cinnamon since 2014 and have been always updating) everything worked without problem. In Mint Cinnamon 20:
1. the script itself works (the directory is opened)
2. but autokey gives an error (both inside the program log as well as trashing my OS notifications list every time I invoke one of my many scripts).

OS Notifications:
![image](https://user-images.githubusercontent.com/2521942/98740227-62e19080-23b3-11eb-8296-361ff3cadafd.png)
![image](https://user-images.githubusercontent.com/2521942/98740342-8c022100-23b3-11eb-857b-3097548ed045.png)

Error Autokey gives:
![image](https://user-images.githubusercontent.com/2521942/98740501-d6839d80-23b3-11eb-97c3-d6c79a4c0bcd.png)

## Actual Results
Here is an attached log file: [2020-11-11_Autokey_error.txt](https://github.com/autokey/autokey/files/5520555/2020-11-11_Autokey_error.txt)

Three things that interesting/worrying:
Errors 1:
```
2020-11-11 00:24:44,862 DEBUG - inotify - Adding watch for /home/mikko/.config/autokey/data/Linux cursors/-button
[2020-11-11 00:24:44,862 pyinotify ERROR] add_watch: cannot watch /home/mikko/.config/autokey/data/Linux cursors/-button WD=-1, Errno=No space left on device (ENOSPC)
2020-11-11 00:24:44,862 ERROR - pyinotify - add_watch: cannot watch /home/mikko/.config/autokey/data/Linux cursors/-button WD=-1, Errno=No space left on device (ENOSPC)
```

Error 2. 
```
(autokey-gtk:305378): Gdk-CRITICAL **: 00:24:45.478: gdk_window_thaw_toplevel_updates: assertion 'window->update_and_descendants_freeze_count > 0' failed
[...]

Issue #354, opened 2020-01-15
autokey script omits last character


freddyw:
## Classification:
Bug
(Pick one of: Bug, Crash/Hang/Data Loss, Performance, UI/Usability, Feature (New), Enhancement)

## Reproducibility:
Always
(Pick one of: Always, Sometimes, Rarely, Unable, I Didn't Try)

## Version

AutoKey version: 
0.95.4, 0.95.7
Used GUI (Gtk, Qt, or both): 
Gtk

Installed via: (PPA, pip3, …).
fedora DNF, OracleLinux pip3
Linux Distribution:
Fedora 31 and OracleLinux 8
## Summary

Summary of the problem.
I have a script that retrieves a passphrase from the gnome-keyring. It works flawless when I use a python script 'gkeyring'. That is a python2 script however. So I wanted to use secret-tool instead (libsecret rpm) as this is (default) part of the OS. But then I use secret-tool to extract a passphrase (which it does fine on its own) and then use autokey the last character is consistently omitted.

## Steps to Reproduce (if applicable)
- secret-tool store --label="somephrase" somephrase somekey
=> whatever
- $secret-tool lookup somephrase somekey
whatever
- create an autokey script
output = system.exec_command("secret-tool lookup somephrase somekey")
keyboard.send_keys(output + "\n")
- execute script

## Expected Results

'whatever' is returned by the autokey script

## Actual Results
```
2020-01-15 09:41:33,853 INFO - service - Matched hotkey phrase/script with prompt=False
2020-01-15 09:41:33,854 DEBUG - service - Script runner executing: Script('somescript')
2020-01-15 09:41:33,854 ERROR - service - Ignored locking error in handle_keypress
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/autokey/service.py", line 207, in __tryReleaseLock
    self.configManager.lock.release()
RuntimeError: release unlocked lock
2020-01-15 09:41:33,884 DEBUG - iomediator - Send via event interface
2020-01-15 09:41:33,885 DEBUG - interface - Sending string: 'whateve'
2020-01-15 09:41:33,900 DEBUG - interface - Send special key: ['<enter>']
whateve
2020-01-15 09:41:33,962 DEBUG - iomediator - Key.SHIFT released
2020-01-15 09:41:33,983 DEBUG - iomediator - Key.CONTROL released
```
## Notes
[...]

PR #486, opened 2020-11-30
fix #483

### LINKED ISSUE #483: #354 fix causes IndexError: string index out of range

fx-kirin:
https://github.com/autokey/autokey/issues/483

Reopening https://github.com/autokey/autokey/pull/484

### DIFF
--- a/lib/autokey/scripting.py
+++ b/lib/autokey/scripting.py
@@ -494,11 +494,12 @@ class System:
                     stdout=subprocess.PIPE,
                     universal_newlines=True) as p:
                 output = p.communicate()[0]
-                if output.endswith("\n"):
-                    # Most shell output has a new line at the end, which we
-                    # don't want. Drop the trailing newline character,
-                    # if the command output something that ends on "\n"
-                    output = output[:-1]
+                if len(output) > 0:
+                    if output.endswith("\n"):
+                        # Most shell output has a new line at the end, which we
+                        # don't want. Drop the trailing newline character,
+                        # if the command output something that ends on "\n"
+                        output = output[:-1]
                 if p.returncode:
                     raise subprocess.CalledProcessError(p.returncode, output)
                 return output
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 .. contents::
 
 
-Version 0.95.10 <2019-02-16>
+Version 0.95.10 <2019-12-16>
 ============================
 
 Bug fixes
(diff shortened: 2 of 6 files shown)


lib/autokey/scripting.py
```diff
@@ -494,11 +494,12 @@ def exec_command(self, command, getOutput=True):
                     stdout=subprocess.PIPE,
                     universal_newlines=True) as p:
                 output = p.communicate()[0]
-                if output.endswith("\n"):
-                    # Most shell output has a new line at the end, which we
-                    # don't want. Drop the trailing newline character,
-                    # if the command output something that ends on "\n"
-                    output = output[:-1]
+                if len(output) > 0:
```
sebastiansam55, line 497: At first glance I had questions about if an error would still occur if the command called had no output but it should be good, it seems that with the addition of `stdout=subprocess.PIPE,` the "output" of a command with no other output (eg a test script like `#!/bin/sh`) will still be a empty string `''`.

Unrelated but is there any reason why we don't swap from
`output[:-1]`
to
`.rstrip()`?

sebastiansam55 (changes requested):
When you open a PR try to include more information in the description than fix x, even a short description is better than nothing, our only active person with git PR approval is not a python expert so do your best to explain what your PR is changing and how, the more information the better. 

This PR is also still on `master` instead of `develop`, when you go to the PR page at the top there are a few dropdowns, make sure you select the right branch for both your fork and for the main autokey fork. I also try to name forks prefixed with what branch it is based on so it is more clear (my fork in the screen shot is `develop_altgr` instead of `altgr`, so it is clear to the approver that the PR is for the develop branch based on the develop branch).

Screenshot from 2020-11-30 09-59-49
Note the 2nd dropdown from the left is on develop, by default github sets it to `master` so you have to change it to `develop` everytime.

Sorry to be so nitpicky! 😬😬😬

fx-kirin:
Thank you for your reviewing. l just wanted to report a bug and show a quick fix of it. I don't have time to go further detail of this change including last one https://github.com/autokey/autokey/issues/354  which causes this bug. You can close this issue and PR if you want and I don't know which implementation is better.

lib/autokey/scripting.py
josephj11: I don't know Python, but rstrip seems more expressive/clear than indexing the last character in the string.

josephj11: Did someone finally copy my Problem Reporting Guide from the autokey-python2 branch to the master? I don't see it. That's why the link points to the old repo.

josephj11: The version should probably be 0.95.11 or 0.96 - depending on whether someone wants to put out a minor release before 0.96. I don't think the date of a released version should be modified.

josephj11 (changes requested):

josephj11:
See also #462.

BlueDrink9:
Closed by #489

PR #487, opened 2020-12-02
[Gtk] Ensure proper python libraries and command line programs AutoKey relies on are installed before launching

sebastiansam55:
Add stuff to check that the appropriate python modules and commands are installed, checks for Xlib, dbus, pynotify and gi, along with other python modules, that in my understanding should be distributed with most versions of python, but better safe than sorry. If any packages are missing the program will quit with `sys.exit`. 

Checks if command line programs are installed as well using `which` from `shutil` module. Only checks for currently required programs to be installed as dictated for `autokey-gtk` in `debian/control`, (`zenity`, `wmctrl`, `ps`).

Error is displayed using the provided `show_error_message` which is a wrapper for `MessageDialog`. Not sure how much can be done about the layout of the information, it is a bit weird the way it is now imo, but also not sure how much effort should be applied to what should be a rare error case/scenario.

Screenshot from 2020-12-02 19-24-26

### DIFF
--- a/lib/autokey/UI_common_functions.py
+++ b/lib/autokey/UI_common_functions.py
@@ -7,13 +7,13 @@ logger = __import__("autokey.logger").logger.get_logger(__name__)
 common_modules = ['argparse', 'collections', 'enum', 'faulthandler', 
             'gettext', 'inspect', 'itertools', 'logging', 'os', 'select', 'shlex',
             'shutil', 'subprocess', 'sys', 'threading', 'time', 'traceback', 'typing',
-            'warnings', 'webbrowser', 'dbus', 'pyinotify', "asddgs"]
+            'warnings', 'webbrowser', 'dbus', 'pyinotify']
 gtk_modules = ['gi', 'gi.repository.Gtk', 'gi.repository.Gdk', 'gi.repository.Pango',
             'gi.repository.Gio', 'gi.repository.GtkSource']
 qt_modules = ['PyQt5', 'PyQt5.QtGui', 'PyQt5.QtWidgets', 'PyQt5.QtCore',
             'PyQt5.Qsci']
 
-common_programs = ['wmctrl', 'ps', "sxvbsd"]
+common_programs = ['wmctrl', 'ps']
 # Checking some of these appears to be redundant as some are provided by the same packages on my system but 
 # better safe than sorry.
 optional_programs = ['visgrep', 'import', 'png2pat', 'xte', 'xmousepos']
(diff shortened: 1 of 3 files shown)


josephj11:
This looks much improved. It looks like it no longer checks for the Xautomation stuff at all. I think we should still check for them, but just inform the user we didn't find them without causing a fatal error. "The following optional components (named) for use with the Xautomation integration are not installed" or something like that.

sebastiansam55:
Picture of how missing "optional" programs will be displayed to the user.

Screenshot from 2020-12-03 16-46-02

josephj11:
I'm good with merging this after it ages a few days and you let me know.

BlueDrink9:
This looks better than nothing. If you can, try not to contribute to code duplication as much as possible. Stick most of it in a library script that you then import, if you can. Possibly create a function `checkModuleImports(moduleList)`?

BlueDrink9:
Also note that the dev branch has a powerful logging system, so for things like Xautomation can just go into the log instead of being GTK messages? And more generally, the dialogue messages should also be logged. I can't remember where to look for examples, but logging should be trivial. Maybe check `engine.py`?

sebastiansam55:
I'm not super versed in python terminology.

By "library script" you mean a new file? So the code can be used in both `qtapp.py` and `gtkapp.py`?

The logging comment makes sense to me, I might just add the logging to the `show_error_dialog` and `show_warning_dialog` methods themself

BlueDrink9:
Yes, a new file. I don't think we have such a shared file for UIs yet; that's something I am planning on fleshing out. For now, maybe create something like `UI_common_functions.py`?

sebastiansam55:
See logging output and what the missing requirement popup look like for Gtk and Qt UI looks like below.

Screenshot from 2020-12-18 13-37-57
Screenshot from 2020-12-18 13-30-43
Screenshot from 2020-12-18 13-30-35

lib/autokey/UI_common_functions.py
```diff
@@ -0,0 +1,64 @@
...
+            missing_modules.append(module)
+
+    return missing_modules
+
+def checkProgramImports(programs):
+    missing_programs = []
+    for program in programs:
+        if which(program) is None:
+            # file not found by shell
+            logger.debug("Commandline Program: \""+program+"\" was not found/able to be used correctly by AutoKey. Check that this program is correctly installed on your system.")
```
BlueDrink9: I would use `logger.info` for optional/non-essential missing imports, and `logger.error` for missing essentials.

lib/autokey/qtapp.py
```diff
@@ -295,6 +303,7 @@ def show_error_dialog(message: str, details: str=None):
         Convenience method for showing an error dialog.
         """
         # TODO: i18n
+        logger.info("Displaying Error Dialog")
```
BlueDrink9: This is where you would use a `logger.debug` statement, I think.

BlueDrink9:
Oops, that review was from two weeks ago. Hopefully it isn't too out of date. If those things are changed, we can probably merge this.

sebastiansam55:
Updated with logging level changes

BlueDrink9 (approved):
@josephj11 I think this is good to merge

PR #455, opened 2020-10-07
Toolbar button for View Script error

am-py:
Hi,

This is to add a toolbar button, as requested in #453 

Also allows script error selection with gtk.

PR #488, opened 2020-12-02
[GTK/Qt] Add error menu item.

### LINKED ISSUE #453: Another small feature request
`## Classification: Enhancement

Could you please add a toolbar button to access "View Script error"? To me, it's the most logical place for it. I am using the gtk version 95.10. Thanks.

sebastiansam55:
A modification of #455 on the develop branch that also includes setting the Gtk Error Icon active/inactive depending on if there are errors left in the error dialog that have not been deleted. 

Fixes #453 

Screenshot from 2020-12-02 15-30-05
Screenshot from 2020-12-02 15-29-58
Screenshot from 2020-12-02 15-11-31
Screenshot from 2020-12-02 15-11-19

### DIFF
--- a/lib/autokey/gtkapp.py
+++ b/lib/autokey/gtkapp.py
@@ -236,6 +236,8 @@ class Application:
         """
         message = "The script '{}' encountered an error".format(error.script_name)
         self.notifier.notify_error(message)
+        if self.configWindow is not None:
+            self.configWindow.set_has_errors(True)
 
     def update_notifier_visibility(self):
         self.notifier.update_visible_status()
--- a/lib/autokey/gtkui/configwindow.py
+++ b/lib/autokey/gtkui/configwindow.py
@@ -889,6 +889,7 @@ close and reopen the AutoKey window.\nThis message is only shown once per sessio
             canMacro = False
             canPlay = False
             enableAny = False
+            hasError = False
         else:
             canCreate = isinstance(items[0], autokey.model.folder.Folder) and len(items) == 1
             canCopy = True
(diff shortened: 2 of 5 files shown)


josephj11:
I don't have that icon set on kubuntu/Qt. I'm a bit confused as to which icon is which. I assume the circle-backslash is the error icon and the circle-dash icon is the delete or do I have it backwards.

I went [here](https://duckduckgo.com/?q=error+icon&t=vivaldi&iax=images&ia=images) and I kind of like the circle-exclamation mark icons for error messages, but that's just a personal preference. I also like circle-x for delete. Circle-dash means "stop" to me - which isn't meaningful in this context. What do you think?

This looks like another one I can merge once it sits a bit.

sebastiansam55:
Most if not all of the icons used by autokey draw from the Gtk/Qt defaults, in this case the error icon for the Gtk literally comes from `Gtk.STOCK_DIALOG_ERROR`. And it looks like the Qt Icons are identical on my system but that might just be because I only have minimal Qt stuff installed.

I tried looking around for where a list of the stock icons could be viewed but didn't see anything obvious. 

Interestingly but [this](https://developer.gnome.org/gtk3/stable/gtk3-Stock-Items.html) page seems to indicate that we should stop using the `STOCK_` references as they are being deprecated in favor of different names.

https://python-gtk-3-tutorial.readthedocs.io/en/latest/iconview.html

Seems to be the correct way of doing icons now? I'm not entirely sure.

A list of the available icons can be viewed with `Gtk.IconTheme().list_icons()` I'd have to read more about it because some of the icon names seem to be related to other programs I have installed.

Just something that might have to be done if `Gtk.STOCK_X` is ever removed from Gtk.

josephj11:
I figured it was something like that. I would defer changes like that until it's clear what to do. I just wasn't sure what they meant. The usual solution for such problems is supplying tool tips on hover. That's nice to have, but not a big priority.

Just for the heck of it, I ran the example code in your link and got a bunch of "depreciated" warning messages, but sometimes those depreciated things hang around for years.

I'm good with merging this after it ages a few days and you let me know.

BlueDrink9:
Yeah, seems fine with me after a quick skim

PR #489, opened 2020-12-02
Fix for issue #483, original by fx-kirin

### LINKED ISSUE #483: #354 fix causes IndexError: string index out of range

sebastiansam55:
Also swaps from striping trailing new line using indexes to `rstrip('\n')`.

See https://github.com/autokey/autokey/pull/486/files#r532656403.

Output of `subprocess` seems to always be at least a empty string: `''`, so checking length shouldn't be a problem, no more than the previous method, `endswith` which also (I think) assumes output to be a string.

Credit goes to https://github.com/fx-kirin for original fix.

### DIFF
--- a/lib/autokey/scripting/system.py
+++ b/lib/autokey/scripting/system.py
@@ -43,10 +43,8 @@ class System:
                     stdout=subprocess.PIPE,
                     universal_newlines=True) as p:
                 output = p.communicate()[0]
-                if len(output) > 0:
-                    # Most shell output has a new line at the end, which we
-                    # don't want. Strip trailing newline.
-                    output = output.rstrip('\n')
+                # Most shell output has a new line at the end, which we don't want.
+                output = output.rstrip("\n")
                 if p.returncode:
                     raise subprocess.CalledProcessError(p.returncode, output)
                 return output
--- a/lib/autokey/scripting/system.py
+++ b/lib/autokey/scripting/system.py
@@ -43,11 +43,10 @@ class System:
                     stdout=subprocess.PIPE,
                     universal_newlines=True) as p:
                 output = p.communicate()[0]
-                if output.endswith("\n"):
+                if len(output) > 0:
                     # Most shell output has a new line at the end, which we
-                    # don't want. Drop the trailing newline character,
-                    # if the command output something that ends on "\n"
-                    output = output[:-1]
+                    # don't want. Strip trailing newline.
+                    output = output.rstrip('\n')
                 if p.returncode:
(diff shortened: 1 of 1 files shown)


mvindiola1:
I ran into this error today. I wanted to add two comments that you might consider.

1. You do not really need the conditional because if you rstrip an empty string you get the empty string back. You could do it cleanly and more expressively in one line like this:
    `output = p.communicate()[0].rstrip('\n')`.

    As an aside, if you want to use python indexing, the correct way to test this would have been:
    `if t[-1:] == "\n":`

2. There are a couple other places where you could do the same thing if you wanted consistency :
https://github.com/autokey/autokey/blob/c2ebcbba75ce40ed080c78216730b93690cb9a21/lib/autokey/scripting/dialog_gtk.py#L42
https://github.com/autokey/autokey/blob/c2ebcbba75ce40ed080c78216730b93690cb9a21/lib/autokey/scripting/dialog_qt.py#L42
https://github.com/autokey/autokey/blob/c2ebcbba75ce40ed080c78216730b93690cb9a21/lib/autokey/scripting/window.py#L244

lib/autokey/scripting/system.py
```diff
@@ -43,11 +43,10 @@ def exec_command(command, getOutput=True):
                     stdout=subprocess.PIPE,
                     universal_newlines=True) as p:
                 output = p.communicate()[0]
-                if output.endswith("\n"):
+                if len(output) > 0:
```
BlueDrink9: Good advice @mvindiola1. @sebastiansam55 if you just remove this conditional, then @josephj11 should merge this change.

As an aside, I couldn't replicate this issue with automated tests, so if you can figure out how to, add the test to `autokey/tests/scripting_api/test_system.py`

BlueDrink9 (changes requested):

lib/autokey/scripting/system.py
BlueDrink9: @sebastiansam55 can you quickly please replace lines 46-49 with

```python
# Most shell output has a new line at the end, which we don't want.
output = output.rstrip("\n")
```

Then repush and we can merge. This has cropped up as a bug elsewhere.

BlueDrink9:
@josephj11 Should be good to merge

PR #494, opened 2021-01-01
Reduce code duplication across UIs

BlueDrink9:
Bunch of refactoring to reduce code duplication across the two UIs


BlueDrink9:
I think this is enough for one PR. I'd like to get people to test the UI and check I haven't missed anything broken.

PR #499, opened 2021-01-12
Separate out code for using Autokey from the UI code

BlueDrink9:
As well as significantly reducing duplication between the initialisation of the two UIs, this should also make it much simpler to create other UIs (including non-graphical UIs).

This PR also adds a new app to run autokey headlessly, without any GUI. This could be turned into a daemon for users that don't want GUI dependencies with autokey, especially when coupled with a more comprehensive scripting system.

Based off #494


BlueDrink9:
This PR is scaring new contributors so I'll close it until I have time to work on it more. I'll sort out any conflicts if changes get merged to develop before this does.

PR #501, opened 2021-01-15
Configure Github Actions as continuous integration for testing and linting

BlueDrink9:
Also fixes a number of linter errors to ensure the first lot of tests pass, and does some more in-depth configuration of pytest


BlueDrink9:
@josephj11 I think this can be merged to develop straight away, since it doesn't actually touch any code paths that users use

PR #520, opened 2021-03-17
Change the default sendmode for phrases

sebastiansam55:
Changes the default send mode for phrases from the keyboard to ctrl-V method. Per Joe's suggestion in #514

PR #521, opened 2021-03-20
Update engine.py

sebastiansam55:
Change default sendmode for engine.create_phrase to CB_CTRL_V

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,8 @@ Changelog
 =========
 .. contents::
 
+- Change the default phrase send mode to `ctrl+v` (paste using clipboard) rather than sending keys one at a time.
+
 
 Version 0.95.10 <2019-02-16>
 ============================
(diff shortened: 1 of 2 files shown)


josephj11:
Note: This is for consistency with #520 which makes the same change in the GUI.

BlueDrink9:
As mentioned in the Gitter, IMO this is not a good idea. Whereas changing the GUI only applies to new phrases, the scripting API is often used to more easily recreate previous scripts; meaning this change may break existing setups.

josephj11:
@BlueDrink9 As usual, I am not merging this now, but I would like to understand your use case better - on Gitter, etc.

I don't like the inconsistency between the GUI and this API, but anyone sophisticated enough to use this API at all should be able to deal with it with no problem.

At a minimum, I would like something added to the documentation pointing out the difference between the two. (And that would probably involve regenerating the API documentation which, last I saw, was something we were still figuring out how to do in the Python 3 branch of AutoKey.)

I definitely want to keep the GUI version change as that will avoid a lot of end user issues.

BlueDrink9:
> At a minimum, I would like something added to the documentation pointing out the difference between the two. (And that would probably involve regenerating the API documentation which, last I saw, was something we were still figuring out how to do in the Python 3 branch of AutoKey.)

TBH considering how many other changes 0.96 will bring, I think this is a reasonable solution. I'll push a note to the changelog

> I definitely want to keep the GUI version change as that will avoid a lot of end user issues.

Oh absolutely, definitely agree the GUI version is good.

PR #524, opened 2021-03-31
Typo - Added missing space

denis-roy:
Commit is self-explanatory :)

And while I have your attention: Keep up the great work! :+1:


josephj11:
Looks OK to me, but it needs to be applied to the develop branch or it may be lost in 0.96. We also need to check if it needs fixing in the Qt front end.

BlueDrink9:
What Joe said about the `develop` branch. But don't worry about it, I'll sort it when I get a chance. Same re: the QT issue. I'll leave this PR open until as a reminder, then I'll accept it.

denis-roy:
Here is the PR for `develop` -> #525

josephj11:
Replaced by #525

Issue #526, opened 2021-04-02
pasting images with autokey


roof98:
Is there a way to get autokey to paste an image from local storage? (Think pasting an email footer)

sebastiansam55:
https://askubuntu.com/questions/759651/how-to-copy-an-image-to-the-clipboard-from-a-file-using-command-line

`xclip -selection clipboard -t image/png -i ~/pic.png`

In autokey just make a script bound to whatever hotkey you desire and then;
```
system.exec_command("xclip -selection clipboard -t image/png -i ~/pic.png")
```

sebastiansam55:
We should add this to the TODO list as it would be neat for this to be more integrated into autokey itself.

roof98:
not bad. It would be nice to be able to read from something like rtf to handle images

sebastiansam55:
Can you elaborate on what you mean by "reading from rtf"? There is probably some python library out there that would allow you to work within that type of format

roof98:
you can add images to a .rtf (I believe). Instead of having phrases read from .txt, if they could read from .rtf and appriately handle images, that would be nice.

Side question: the command you listed seems to successfully copy an image. Can I use xclip with -o to also output the image? 

Thanks

sebastiansam55:
As long as you are using autokey you can do something like;
```
keyboard.send_keys("<ctrl>+v")
```

If you have `xdotool` installed; `xdotool key ctrl+v`

`xclip -o` outputs to stdout so it probably would not be terribly helpful for pasting images, you probably figure some way to make it work but either of the above should do the trick

josephj11:
Can this be expanded to also work with images from webpages?

PR #527, opened 2021-04-02
Add set_clipboard_image method to Gtk and Qt clipboard implementations.

sebastiansam55:
Fix for #526, adds `set_clipboard_image` methods to both the Qt and Gtk interfaces for the clipboard. Uses native library as opposed to `xclip` as was originally suggested.

`set_clipboard_image` takes one argument, a path to a image file on the local filesystem that will be loaded into the clipboard for pasting.


sebastiansam55:
Open to any suggestions on implementation! Or function name or anything else.

josephj11:
What about putting an image from a webpage into the clipboard? Would it first have to be downloaded by a script? Wouldn't this be a frequent use case? I didn't look at the code.

sebastiansam55:
I'm not sure what the most common use case for something like this is, just throwing it out there as it was requested and implementation seemed pretty trivial. 

Idk I feel like most people would be using something like this with local files? Like what is the scenario where you know the url of the file, have gone to the trouble of using autokey to automate the pasting of it, but don't use it often enough to have it saved locally?

Implementation of URL based clipboard function in a autokey script would look something like this;

```python
import requests
import tempfile
url="https://avatars.githubusercontent.com/u/1707320?s=80&v=4"
r = requests.get(url)

with tempfile.NamedTemporaryFile() as f:
    f.write(r.content)
    clipboard.set_clipboard_image(f.name)
```

sebastiansam55:
Re: implementation

What do you think about adding it into `set_clipboard` main and adding an additional optional argument like `content_type` which could be set to text by default, images and I believe that there is a way to "copy" files in no the clipboard?

Like when you copy paste in a gui file browser I believe that is done via the same sort of method. I'll look into it more Monday, just a thought I had

sebastiansam55:
I should also wrap the gtk version in the gdk enter threads like the other ones, I had no issues in briefly testing it but there is probably a reason the other methods wrap like that

josephj11:
I have to think about it more. My first impression was that whenever I am doing anything with images, they're usually on a someone else's webpage. I would like to write a script to save a Dilbert or other cartoon to a local file with a descriptive file name, probably from an input dialog box. I'd have to try to actually build it to see what is required to get it to work. I'm sure I could do this given some time to work on it.

BlueDrink9 (approved):
I'm happy with it as-is. I'd prefer changes like this to changes that alter existing interfaces.

BlueDrink9:
Actually, since I mentioned it in the Gitter, could you please add this change to the CHANGELOG as part of this PR

BlueDrink9:
> I should also wrap the gtk version in the gdk enter threads like the other ones, I had no issues in briefly testing it but there is probably a reason the other methods wrap like that

I too would like to know what they do. I found some inconsistencies that seemed not to have mattered while doing refactors, so I suspect it may be a case of monkey see = do in some places. Still, best be safe for now.

sebastiansam55:
Only tested the async/threads enter/leave on gtk as you can probably guess from the commits lol. 

I tested the original changes on both, just don't have qt setup currently.

BlueDrink9:
Do you know how to fixup commits? `git rebase -i 5e5fe88c26943a2fb5fe26b7a0c926c8d0683c9c`, then change that last commit to a `fixup` instead of `pick`. Then force push back to the branch. Just keeps the commit history tidier.

Creating the changelog made me realise just how messy the history is atm.

BlueDrink9:
> > I should also wrap the gtk version in the gdk enter threads like the other ones, I had no issues in briefly testing it but there is probably a reason the other methods wrap like that
> 
> I too would like to know what they do. I found some inconsistencies that seemed not to have mattered while doing refactors, so I suspect it may be a case of monkey see = do in some places. Still, best be safe for now.

Also at least one instance generated a deprecation warning, so it may be that it used to be an issue but is not anymore.

sebastiansam55:
> Do you know how to fixup commits? `git rebase -i 5e5fe88c26943a2fb5fe26b7a0c926c8d0683c9c`, then change that last commit to a `fixup` instead of `pick`. Then force push back to the branch. Just keeps the commit history tidier.
> 
> Creating the changelog made me realise just how messy the history is atm.

Not at all, I am vaguely aware that it is possible to delete commits if you accidentally commit a password or something but it's a bit over my head. I'll work on it!

Is there anything else this needs before merge?

BlueDrink9:
Nope, otherwise fine. If you want I can sqaush those two commits myself, but I thought I'd give you the learning experience. Do you use a git gui?

Basically I don't want commits that fix previous commits in the same PR. They should be squashed as a single commit

BlueDrink9 (approved):

PR #595, opened 2021-08-22
Add `send_notification` to GTK and Qt `dialog` modules, uses `zenity`…

sebastiansam55:
… and `kdialog` respectively.

`kdialog` has the option to add a timeout explicitly to the notification (referred to as passive popups in the documentation), but on my Gtk based machine it sends the notification the same way as `zenity`, I assume that if you are using a Qt desktop manager the timeout will function as expected.

For the Gtk based notification the easiest way to make it have both a "title" and a "message" was to separate them with a newline, I assume that this is what is happening at some level on the other ways of sending notification (looks the same as ones sent via `notify-send` or `kdialog`)

Other than that I think it is pretty straight forward what's happening in this commit.

References can be found below;
https://unix.stackexchange.com/questions/516618/what-are-the-differences-disadvantages-of-zenity-vs-notify-send
https://develop.kde.org/deploy/kdialog/#non-interrupting-notifications
https://help.gnome.org/users/zenity/stable/notification.html.en

### DIFF
--- a/lib/autokey/scripting/dialog_gtk.py
+++ b/lib/autokey/scripting/dialog_gtk.py
@@ -46,6 +46,19 @@ class GtkDialog:
 
         return DialogData(return_code, output)
 
+    def send_notification(self, title, message, icon="autokey", **kwargs):
+        """
+        Sends a notification using C{zenity}
+
+        @param title: Title to be used in the notification
+        @param message: Message displayed in the notification
+        @param icon: Icon for zenity to use. Defaults to Autokey icon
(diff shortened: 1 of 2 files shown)


sebastiansam55:
Seems like some of those checks failed on installing dependencies, no idea if that is something to do with my commit or if the servers are down right now or something.

BlueDrink9:
Failing tests looks like a GH actions issue, but actually was just a lack of `apt update`. Should be fixed now.

BlueDrink9 (approved):

Issue #566, opened 2021-06-26
Add new Paste using option for slow applications


josephj11:
## Classification:

Enhancement

## Reproducibility:

N/A

## Version

AutoKey version: N/A

Used GUI (Gtk, Qt, or both): 

Both

## Summary

Some applications respond to user input very slowly. Usually, pasting from the clipboard fixes this, but not always.

We need a way to type things, especially phrases, very slowly, but it can't be implemented globally without making AutoKey almost unusable. It needs to be an exception, available upon demand.

I am proposing adding a new option to the Paste using drop-down box:

Paste using throttled Keyboard

This should work similarly to my [type_slow()](https://github.com/autokey/autokey/wiki/Advanced-Scripts) function, but with the full functionality of the current Paste using keyboard (i.e. full macro support).

## Notes

This has already been researched to some extent. See #151. I don't know if Thomas' code is still accessible.

The option name is a first approximation. A better name would be welcomed.

Since different applications respond at different speeds, it would be helpful to be able to specify character emission speed with an optional parameter for each phrase with a reasonable default value - maybe something like 0.1 seconds/character. The user should be able to set things to "just slow enough" on a case by case basis.

Some applications may be affected by user system load, but there's probably nothing we can do about that.

This might require adding one or more new fields to the phrase metadata sidecar files. This would hopefully be done in a backwards compatible fashion.

bjohas:
Hi @josephj11 - thanks for comment on the enhancement.

I've been experimenting with this (slightly modified):
```
def type_slow(string1, string2, startdelay=0.0, backdelay=0.0, delay=0.3, typedelay=0.0):

  time.sleep(startdelay)

  for c in string1:
    keyboard.send_keys("<backspace>")
    time.sleep(backdelay)
 
  time.sleep(delay)
  
  for c in string2:
    keyboard.send_keys(c)
    time.sleep(typedelay)

  return

type_slow("abc","alphabet")
```
The main thing seems to be the `startdelay` and the `delay` parameters, possibly `backdelay` as well.

The delay parameters would depend on the system and responsiveness of the application. I wonder whether the 'paste slow' option would need to be configurable in that way?

herrdeh:
Thank you very much. I never tried scripts, so it took me a while to understand that I need to replace my phrase by a script. Got it now...

[...]

PR #596, opened 2021-08-24
Develop slowtypemethod

sebastiansam55:
Adds the option to have a delay between sending keystrokes, as has been suggested as a solution to issues with trigger phrases not being completely removed. If we could get someone who has this problem to confirm that this would be a fix I'll work on adding the option to the Gtk UI.

Right now I have it set up to use the same delay for both the backspaces to remove a trigger word and the phrase itself.

If you want to test this you need to clone this branch `git clone https://github.com/sebastiansam55/autokey --branch develop-slowtypemethod`, go to the `lib` directory in the terminal and launch via `python3 -m autokey.gtkui -lc` (Qt should reflect these changes as well). Then open the json file of the phrase that you want to type with a delay and add `"type_delay": 0.1` to the end, making sure that the json is properly formatted.

My config file is pictured below
Screenshot from 2021-08-24 07-28-04

The change uses python's `time.sleep()` method and is only activated when `type_delay` is greater than or equal to 0.

If someone can confirm this fix I can;
* add this to the UI
* add it for abbreviation triggers for scripts
* Add a `keyboard` module method

### DIFF
--- a/lib/autokey/interface.py
+++ b/lib/autokey/interface.py
@@ -789,7 +789,7 @@ class XInterfaceBase(threading.Thread):
                     logger.warning("Unable to send character %r", char)
                 logger.info("Type delay of "+str(type_delay))
                 if type_delay >= 0:
-                    self.localDisplay.flush()
+                    self.flush()
                     time.sleep(type_delay)
             except Exception as e:
                 logger.exception("Error sending char %r: %s", char, str(e))
--- a/lib/autokey/interface.py
+++ b/lib/autokey/interface.py
@@ -692,10 +692,10 @@ class XInterfaceBase(threading.Thread):
 
         return None, None
 
-    def send_string(self, string):
-        self.__enqueue(self.__sendString, string)
+    def send_string(self, string, type_delay=0):
+        self.__enqueue(self.__sendString, string, type_delay)
         
-    def __sendString(self, string):
+    def __sendString(self, string, type_delay=0):
         """
         Send a string of printable characters.
         """
@@ -787,6 +787,10 @@ class XInterfaceBase(threading.Thread):
                         self.__releaseKey(Key.SHIFT)
                 else:
                     logger.warning("Unable to send character %r", char)
+                logger.info("Type delay of "+str(type_delay))
+                if type_delay >= 0:
+                    self.localDisplay.flush()
+                    time.sleep(type_delay)
             except Exception as e:
                 logger.exception("Error sending char %r: %s", char, str(e))
 
--- a/lib/autokey/iomediator/iomediator.py
+++ b/lib/autokey/iomediator/iomediator.py
@@ -142,7 +142,7 @@ class IoMediator(threading.Thread):
         
     # Methods for expansion service ----
 
-    def send_string(self, string: str):
+    def send_string(self, string: str, type_delay=0):
         """
         Sends the given string for output.
         """
@@ -170,14 +170,14 @@ class IoMediator(threading.Thread):
                         else:
                             self.interface.send_modified_key(section[0], modifiers)
                             if len(section) > 1:
-                                self.interface.send_string(section[1:])
+                                self.interface.send_string(section[1:], type_delay)
                             modifiers = []
                     else:
                         # Normal string/key operation
                         if Key.is_key(section):
                             self.interface.send_key(section)
                         else:
-                            self.interface.send_string(section)
+                            self.interface.send_string(section, type_delay)
                             
         self._reapply_modifiers()
         
@@ -242,6 +242,7 @@ class IoMediator(threading.Thread):
         """
         for i in range(count):
             self.interface.send_key(Key.BACKSPACE)
+            self.flush()
 
     def flush(self):
         self.interface.flush()
--- a/lib/autokey/iomediator/iomediator.py
+++ b/lib/autokey/iomediator/iomediator.py
@@ -15,6 +15,7 @@
 
 import threading
 import queue
+import time
 
 from autokey.configmanager.configmanager import ConfigManager
 from autokey.configmanager.configmanager_constants import INTERFACE_TYPE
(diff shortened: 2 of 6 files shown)


josephj11:
Nice. Looking forward to seeing if this addresses some of the other issues we are seeing.

BlueDrink9:
Very good. ~~Do you think it would be too difficult to make this a per-phrase option rather than global?~~ EDIT: Nvm, my apologies it is per-phrase.

sebastiansam55:
Well it is per phrase, it pulls the type delay from the json file for each phrase, and phrase.py sets the default value to 0, so if one is not present it's 0. And in the `set_string` arguments it defaults to 0 in the arguments for the function so when something calls it that doesn't include the variable it skips the `time.sleep`.

lib/autokey/interface.py
```diff
@@ -787,6 +787,10 @@ def __sendString(self, string):
                         self.__releaseKey(Key.SHIFT)
                 else:
                     logger.warning("Unable to send character %r", char)
+                logger.info("Type delay of "+str(type_delay))
+                if type_delay >= 0:
+                    self.localDisplay.flush()
```
BlueDrink9: I don't like that this is duplicated at [iomediator.py:247](https://github.com/autokey/autokey/pull/596/files#diff-3a6cf8fe36360a25afed01c46f7ce6976402ffee4ebb8b683991559c90d3793fR247).

I think it would be better to remove any mention of `type_delay` from interface.py, put this 3/4 lines in a function in iomediator.py, and replace your calls to `send_key` with that new function (if that makes sense)

lib/autokey/iomediator/iomediator.py
```diff
@@ -236,12 +237,15 @@ def send_up(self, count):
...
 
-    def send_backspace(self, count):
+    def send_backspace(self, count, type_delay=0):
         """
         Sends the given number of backspace key presses.
         """
         for i in range(count):
             self.interface.send_key(Key.BACKSPACE)
+            if type_delay >= 0:
+                time.sleep(type_delay)
```
BlueDrink9, line 247: See previous note about duplicating these 3 lines

lib/autokey/model/phrase.py
```diff
@@ -84,7 +85,8 @@ def get_serializable(self):
             "abbreviation": AbstractAbbreviation.get_serializable(self),
             "hotkey": AbstractHotkey.get_serializable(self),
             "filter": AbstractWindowFilter.get_serializable(self),
-            "sendMode": self.sendMode.value
+            "sendMode": self.sendMode.value,
+            "type_delay": self.type_delay
```
BlueDrink9: If syntax permits, may as well add a comma at the end of the line to make tidier diffs for next time. (See [Black](https://pypi.org/project/black/) for reasoning)

lib/autokey/interface.py
BlueDrink9: Oh actually I forgot how bloated sendString is on this fork; my refactor branch tidied it up a huge amount. Leave this as-is for now, I think.

BlueDrink9 (commented):
The backend stuff looks great, considering the mess we are working with. I can't think of any better way to implement this

BlueDrink9:
As a side note, if you rebase on develop and push again the tests should be fixed.

josephj11:
This looks like a step in the right direction. The function that @bjohas enhanced in #566 has three separate optional delay parameters, so I'm not sure if reducing it to just one will cover all cases.

BlueDrink9:
Also, if you add this to a gui don't forget to also add it to `engine.create_phrase` in the scripting api if appropriate.

(Also update changelog)

bjohas:
> This looks like a step in the right direction. The function that @bjohas enhanced in #566 has three separate optional delay parameters, so I'm not sure if reducing it to just one will cover all cases.

I haven't experimented extensively with it, but it seems to me that the typing/deleting speed is one thing but that introducing a delay when switching from deleting to typing seems to be important. So you could be losing a lot of time introducing delays between letters, when really what you want is a delay when switching from deleting to typing. 

Of course - once all the delays are large enough, it doesn't matter, but I had to introduce quite significant delays between letters. Introducing a delay when switching from deleting to typing meant faster execution overall.

The model I ended up with is in #566, where I actually didn't have a delay between letters at all, but just a delay when switching over. I don't know why that is, but somehow deleting and typing seem to be handled differently. Hope that helps.

lib/autokey/scripting/keyboard.py
```diff
@@ -45,7 +60,7 @@ def send_keys(self, key_string, send_mode: typing.Union[
 
         Usage: C{keyboard.send_keys(keyString)}
 
-        @param key_string: string of keys to send. Special keys are only possible in keyboard mode.
+        @param key_string:
```
BlueDrink9, line 63: What reasoning for removing this docstring?

BlueDrink9:
I have just rebased this against the newly-refactored `develop`. I suggest you reset --hard to https://github.com/BlueDrink9/autokey/tree/sebastiansam55-develop-slowtypemethod

allanlaal:
@sebastiansam55 just tried branch `develop-slowtypemethod` tip. seems to work ok with medium load and loooong urls  :)

altough Quit will remove `"type_delay` from autokey.json

allanlaal:
I set `"type_delay": 0.001`

I can't remember if this has already been discovered, but it seems with superfast keyboard emulation typing there is not enough time for entering keycombo characters like `: / ? = &  ` (shift + numrow keys on my estonian layout):
`httpsxgis.maaamet.eemapsisappid82useridat11620958zlevel10,549052.10876467,6546996.6812135setlegend1821`

Expected:
`https://xgis.maaamet.ee/maps/XGis?app_id=UU82A&user_id=at&LANG=1&WIDTH=1620&HEIGHT=958&zlevel=10,549052.10876467,6546996.6812135&setlegend=UUKAT1_82=1`

the tool I use to diff the results: https://www.convertonline.io/url/diff

"type_delay": 0.3 is the quickest that works for me (2.50GHz × 8 i7 at 80..150% (load: 7..14) + 32GB RAM, SSD) 
this workaround is by definition so slow that with longer autocompletes I might aswell go make coffee while I wait, which kinda negates the point of using text files to copypaste stuff from vs Autokey. has there been any other alternatives to slow typing? I tend to *use* my computer at its limit and other apps need 10x more load to start acting weird :|

PR #602, opened 2021-08-31
Feature: Create headless (non-GUI) entrypoint app

BlueDrink9:
Refactor: Strips out a great deal of the application logic that was duplicated between GTK and QT UIs.
Feature: Create a GUI-free headless entrypoint to autokey.

Combined, these also help a lot with test 'coverage', since the headless app can be automatically run and shutdown. While it doesn't help test behaviour per se, it does help catch a lot of semantic errors because it runs a lot of code pathways (all the initialisation code).


sebastiansam55:
Great stuff!

josephj11:
Glad you got this ready to go. It's a big piece of work and hopefully eliminates some technical debt,

PR #603, opened 2021-08-31
Add distinction between left and right modifier keys

sebastiansam55:
WIP, like others have noted there are some major backend issues with making the distinction between the left and right shift.

### DIFF
--- a/lib/autokey/gtkui/data/hotkeysettings.xml
+++ b/lib/autokey/gtkui/data/hotkeysettings.xml
@@ -76,8 +76,8 @@
                 <property name="spacing">5</property>
                 <property name="homogeneous">True</property>
                 <child>
-                  <object class="GtkToggleButton" id="controlButton">
-                    <property name="label" translatable="yes">Control</property>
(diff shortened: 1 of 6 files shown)


josephj11:
Could `<shift>` and the others be made to default to the left or right variant so that (most) old scripts would still work?
If someone has remapped one of the shift keys on their keyboard, this might cause them problems, but that should just be documented.

BlueDrink9:
I looked into rebasing this onto develop, but the single commit is too huge. I'll ask you to do it yourself I'm afraid.

rbreaves:
This is a very worthy inclusion, I have a large set of complicated remaps that only work with this sort of distinction setup.. I believe I can now make use of this project!

sebastiansam55:
big changes related to this coming very soon, will be on a different commit!

josephj11:
Cool!!

How will this be exposed in the Record Hotkey dialog? I assume that it should somehow default to the current behavior of left and right ones being identical. Maybe a checkbox could be added to disinguish between left and right variants when the keys are subsequently pressed.

sebastiansam55:
see #844

PR #606, opened 2021-09-03
Refactor: Sweeping tidyup of backend codebase

BlueDrink9:


BlueDrink9:
@sebastiansam55 @josephj11 
This is indeed massive, so I don't expect much of a code review per se. What I would appreciate is a look around to check that functionality is retained. I expect there to still be a few bugs I haven't caught myself.

josephj11:
I still haven't had the time/gumption to configure/install a test version, so I can't test this for you.

BlueDrink9:
@josephj11 would this help?
`git clone https://github.com/bluedrink9/autokey`  
Then you can just switch the branch and run autokey.
```
cd autokey
git checkout refactor
python3 -m autokey.qtui
python3 -m autokey.gtkui
```

lib/autokey/interface.py
```diff
@@ -671,116 +678,121 @@ def __sendString(self, string):
...
+        if keyCode is not None:
+            if offset == 0:
+                if self.localDisplay.lookup_string(ord(char)) is None:
+                    # No reasonable translation of key to string found
+                    self.__sendByTypingAsUnicodePoint(char, focus)
+                else:
+                    self.__sendKeyCode(keyCode, theWindow=focus)
+            if offset == 1:
+                self.__send_keycode_with_modifiers_pressed(
+                    keyCode, Key.SHIFT, focus)
```
sebastiansam55, line 723: looks like this introduced the issue to be addressed in #824  the __send_keycode_with_modifiers_press function is expecting a list of modifiers that it can iterate through, passing a single `Key.SHIFT` value will have the for loop treat it as a string

PR #626, opened 2021-11-11
Add more keywords to .desktop files

mikkovedru:


BlueDrink9:
Thank you for this contribution! Please rebase this against the `develop` branch, because that branch also has some changes to these keywords. Let me know if you can't figure out how to do that on your own

mikkovedru:
@BlueDrink9 
I am a bit confused and need clarification.

The 2 files in question, before my changes, are identical in both `master` and `develop` branches. So no rebasing is needed.

Should I remove this pull request and create a new one targeting the `develop` branch? Or is this target branch change possible within the same pull request?

BlueDrink9:
It is possible, but I see you are right about the .desktop files. One of the other metadata files (I think setup.py) has keywords changed. I wonder if the .desktop files are meant to be regenerated from the setup.py…

I will investigate and get back to you

mikkovedru:
@BlueDrink9 
I checked it and they are not. The keywords mentioned in the `setup.py` file: https://github.com/autokey/autokey/blob/5156578ee9dc0a0fa421bce956d4bb06f077484e/setup.py#L214 
are simply keywords for the installation package, not the installed software itself. 

But yeah, it makes sense to update this list as well. I will do it now.

mikkovedru:
@BlueDrink9 
I created a new updated PR #628 .

I think this PR can be closed now.

PR #628, opened 2021-11-12
Add more keywords to setup.py and .desktop files

mikkovedru:
Fixed and updated version of the Pull Request #626.


josephj11:
What are these keywords used for?

mikkovedru:
> What are these keywords used for?

For suggesting an appropriate program when searching in the Start menu or similar. That's when talking about the `.desktop` files.

The `setup.py` one is for the software/package center.

PR #636, opened 2021-12-03
Update wait_for_click to return boolean if click was received before …

### LINKED ISSUE #615: mouse.wait_for_click should have a return code like the keyboard counterpart
## Classification:

Feature (New)

## Reproducibility:

Always

## Version

AutoKey version: 0.95.10

Used GUI (Gtk, Qt, or both): Gtk (Gnome) 

Installed via: (PPA, pip3, …). Honestly can't remember.

Linux Distribution: Ubuntu 20.04

## Summary

`mouse.wait_for_click` does not have a return code so there is no way to know if it was clicked or if it timed out.

- grab return value from `mouse.wait_for_click()`
- log the return value, it is always `None`

## Expected Results

- similar to `keyboard.wait_...` it should have a return code.

## Actual Results

- always return `None`

sebastiansam55:
…timeout.

Fix #615

### DIFF
--- a/lib/autokey/scripting/mouse.py
+++ b/lib/autokey/scripting/mouse.py
@@ -81,10 +81,14 @@ class Mouse:
 
         @param button: they mouse button click to wait for as a button number, 1-9
         @param timeOut: maximum time, in seconds, to wait for the keypress to occur
+        @return: Returns true if mouse is clicked, false if timeout is reached.
         """
         button = int(button)
         w = autokey.iomediator.waiter.Waiter(None, None, button, None, None, timeOut)
-        w.wait()
+        self.mediator.listeners.append(w)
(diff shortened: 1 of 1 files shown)


josephj11:
Thanks for this. I assume the return value will be ignored by existing code that doesn't know about it.

While you're there, there's a typo in line 62. `they` should be `the`.

josephj11:
Nit picking: PR title would be better if "if" was replaced by "indicating whether" because it always returns the boolean, not conditionally. This has no impact on the code itself, just the title.

Added click received boolean return code to Wait_for_click

is another way to say it.

flow-c:
I've just tried the `develop` branch (commit e0a2753b) and the problem is fixed there! 

Mouse clicks are recognised as expected.

Issue #601, opened 2021-08-29
Changing an autokey file while autokey gui is open on another x desktop causes crash


BlueDrink9:
## Classification:

Crash

## Reproducibility:

Always

## Version

AutoKey version: develop 9dec16f7b86ad012bd95dca517e71eab20daa561

Used GUI GTK (QT verified not to be affected)

Installed via: git

Linux Distribution: Arch (bspwm)

## Summary

Summary of the problem.

## Steps to Reproduce (if applicable)

1. Open autokey-gtk gui
2. Switch to another workspace/virtual desktop
3. Edit an autokey .json file in an external editor, while gui isn't visible but is open on the other window

## Expected Results

Autokey gui reloads data in the background

## Actual Results

Autokey crashes

```
2021-08-29 12:40:54,152 WARNING - autokey.configmanager.configmanager - No action taken for create/update event at /home/user/.config/autokey/data/abc/4913
2021-08-29 12:40:54,153 DEBUG - autokey.monitor - Reporting IN_DELETE event at /home/user/.config/autokey/data/abc/4913
2021-08-29 12:40:54,654 WARNING - autokey.configmanager.configmanager - No action taken for delete event at /home/user/.config/autokey/data/abc/4913
2021-08-29 12:40:54,655 DEBUG - autokey.monitor - Reporting IN_MOVED_FROM event at /home/user/.config/autokey/data/abc/folder.json
2021-08-29 12:40:54,724 DEBUG - autokey.iomediator.iomediator - Key.SUPER pressed
2021-08-29 12:40:55,156 WARNING - autokey.configmanager.configmanager - No action taken for delete event at /home/user/.config/autokey/data/abc/folder.json
2021-08-29 12:40:55,156 DEBUG - autokey.monitor - Reporting IN_CREATE event at /home/user/.config/autokey/data/abc/folder.json
2021-08-29 12:40:55,264 DEBUG - autokey.iomediator.iomediator - Key.SUPER released
2021-08-29 12:40:55,658 ERROR - autokey.model.folder - Error while loading json data for abcd
Traceback (most recent call last):
  File "/home/user/src/autokey/lib/autokey/model/folder.py", line 124, in load_from_serialized
    self.inject_json_data(data)
  File "/home/user/src/autokey/lib/autokey/model/folder.py", line 137, in inject_json_data
    AbstractHotkey.load_from_serialized(self, data["hotkey"])
[...]

PR #638, opened 2021-12-06
Potential fix for Segfault from changing files when open in AK window

sebastiansam55:
Fix or at least mitigate #601. It stopped happening after I made these changes, but not convinced it's completely gone, if anyone can replicate please let me know.

Also fixes an issue with the modifiers being set to None (sets to empty list instead) which can throw a NoneType error in some scenarios.


BlueDrink9:
Hmm, I agree that it doesn't seem like the ideal fix (which would check the window is visible/active), but it seems to fix it for me, so it's better than the current code. Thanks Sam!

PR #639, opened 2021-12-06
[Gtk] Autocomplete in text editor

sebastiansam55:
Adds basic autocompletions to the text editor for AutoKey. Will suggest all the methods of each module ie typing `store.` will then popup a Gtk Autocompletion window offering `get_value(`, `set_value(` as completions.

It's manually added, but better than nothing?

### DIFF
--- a/lib/autokey/gtkui/configwindow.py
+++ b/lib/autokey/gtkui/configwindow.py
@@ -42,6 +42,7 @@ locale.setlocale(locale.LC_ALL, '')
 
 
 from . import dialogs
+from . import gtkcompletion
 from .settingsdialog import SettingsDialog
 
 import autokey.configmanager.configmanager as cm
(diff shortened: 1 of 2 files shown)


BlueDrink9:
I'm pretty wary of this to be honest. Is it better than nothing? Currently, yes. However, if any of the modules change, then we have a whole bunch of bug reports about missing modules and things that seem like they should work, not working. Manual entry is also a big maintenance burden. For that reason, I don't think this would be good to merge. Maintenance burden should be a priority for a project in the health that Autokey is in, so in this case I think it outweighs the new feature :/

If I'd known you were working on this, I might have dissuaded you earlier, sorry. I hope it hasn't taken too many hours.

If you can find a way to do it automatically, then this is absolutely good, and we can merge.

sebastiansam55:
No worries, those reasons all make sense. 

I was just curious how autocomplete worked in editors, so it was mostly a learning exercise for me.

PR #660, opened 2022-02-10
Sync localDisplay after emitting a fake_input key event

gharris1727:
Problem:

I was trying to script the game Elite Dangerous, and found that none of the input methods described in the documentation were suitable for writing deterministic scripts. I found that the send_key and send_keys functions were not registered in the application at all, and the press_key/release_key functions were intermittent, and behaved differently almost every time I invoked the script.

Solution:

I discovered that by inserting a sync call after the fake_input calls, that the key codes were able to be registered by the application, and the scripts became very reliable with some dead-reckoned sleeps. The scripts that I have written and are now usable thanks to this patch are here: https://github.com/gharris1727/elite-scripts/tree/main/autokey-scripts .

I do not have insight into the deeper implications of this change (performance or otherwise), and have not manually tested it's behavior with any other applications at this time.

Thanks for maintaining this software, i'm glad to be able to make use of it!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ Create a GUI-free headless entrypoint to autokey, which can be run without GUI l
 .. ---------
 
 - Fix crash in qt macro recording window.
+- Fix fake keyboard events not being emitted in a timely manner in some cases
 
 Version 0.96.0-beta.9
 ============================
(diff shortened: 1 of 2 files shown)


josephj11:
Welcome to the AutoKey community! One of our devs will take a look at this when they get a chance.

Just in case you haven't noticed them, we have two support forums you can participate in if you want to. I pasted in a boiler plate about them below, but it is *not* intended as a comment on this pull request.

----

BTW: Whenever you have a question (as opposed to an issue) it's better to post it on our [list](https://groups.google.com/forum/#!forum/autokey-users) or on [Gitter](https://gitter.im/autokey/autokey). The list is more user oriented whereas more technical questions will get a better answer on Gitter.

Both of these venues have a number of active users, so you are likely to get more, better, and faster answers on them than here on GitHub where only our very busy developers and a few users see the posts. We also prefer to have the developers developing rather than taking time to answer questions and manage issues that the user community could handle for them.

BlueDrink9:
This is interesting! I'll have a closer look at this when I get a chance, to see if this is the best place for those lines and investigate what effects the sync actually has. I might ask you to test them, too.

Issue #663, opened 2022-02-11
Remove import time from example scripts


josephj11:
### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

Documentation

### Which AutoKey GUI did you use?

Both

### Which AutoKey version did you use?

0.96.0 Beta 10

### How did you install AutoKey?

GitHub debs

### Can the issue be reproduced?

N/A

### What are the steps to reproduce the issue?

Just install AutoKey and view the sample scripts.

### What should have happened?

`Display Window Info` and `Phrase from selection` example scripts include `import time` statements that are not needed.

These statements should be removed.

PR #672, opened 2022-02-16
remove unused import

0xwho7:


0xwho7:
#663

josephj11:
Thanks!

We try (for the most part) to leave the master branch unchanged between releases. PRs should be made relative to the development repository. In this case, (if this is correct) it's probably safe enough to go directly into the beta repository. (I'm not one of the devs.)

I'm not sure where this is in the code, but it appears to be within AutoKey itself. I haven't looked at this part of the code. What I was addressing in the issue were two of the example script files that come with AutoKey that include import time statements. I don't know if AutoKey itself needs to import time.

If you're interested in becoming more active in the project, consider joining us over on [Gitter](https://gitter.im/autokey/autokey) where the devs and advanced users hang out and discuss things.

0xwho7:
Thanks for clarifying, i wasn't exactly sure which files you were referring to, so i just went through the imports and removed the one which wasn't being used. I'll definitely join the Gitter!

BlueDrink9:
Joseph is right about changes to master. Especially in the tests, because they have been totally re-written so this PR would just cause merge conflicts. Thank you for contributing!

PR #765, opened 2023-01-31
Update bug.yaml

Elliria:
Add a section at the top to find out if the user is using a Xorg or Wayland session.


Elliria:
Add Xorg/Wayland section to the issue report form.

Issue #773, opened 2023-02-18
Update Python versions in the develop branch


Elliria:
### AutoKey is a Xorg application and will not function in a Wayland session. Do you use Xorg (X11) or Wayland?

Xorg

### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

Enhancement

### Which AutoKey GUI did you use?

None

### Can you briefly describe the issue?

All references to Python versions in the **develop** branch of the source code need to be updated to currently-supported versions.

### Can the issue be reproduced?

Always

### What are the steps to reproduce the issue?

Submit a pull request to the project.

### What should have happened?

It should pass all tests.

### What actually happened?

It fails on Python versions.

josephj11:
Research:
    (All line numbers taken from 0.96.0 Beta 10)

See https://github.com/autokey/autokey/issues/58 IDK if it will help.

setup.py no longer refers to python.dbus, so that proposed change was made.

See setup.py lines 36 - 38, 191. It's testing for Python 3.5 or greater and probably needs to be bumped, but it won't help this issue. Looks like you got these.

autokey/debian/py3dist-overrides contains:

dbus-python python3-dbus
python-xlib python3-xlib

/autokey/.github/workflows/python-test.yml 
158        pip install dbus-python gobject pygobject PyQt5 qscintilla
Looks like you got this.

which contains both offenders, but IDK what that file is used for. Looks important.

*** Did you catch this one? ***

autokey/pip-requirements.txt contains:

dbus-python
pyinotify
python-xlib
PyGObject
PyQt5
QScintilla
pyhamcrest

again, both offenders. Looks important.

Looks like gobject and PyGObject don't appear anywhere else.

Elliria:
Thank you for all the research. I didn't catch that one and I'll see what I can figure out from all of the things you found.

By the way, you'll want to make sure you're looking at [AutoKey's develop branch](https://github.com/autokey/autokey/tree/develop) so you can examine the most current files and the ones I'm trying to change.

Elliria:
The #58 issue looks interesting. Two of the packages most often recommended there (`libdbus-1-dev` and `libdbus-glib-1-dev`) are already AutoKey dependencies, but a comment by [ding3my](https://github.com/autokey/autokey/issues/58#issuecomment-318249538) recommended installing `dbus-python-devel` and that isn't in the list of AutoKey's dependencies.

[...]

PR #774, opened 2023-02-18
Develop

Elliria:
Update all Python versions to satisfy #773.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: [3.11.1]
     steps:
     - uses: actions/checkout@v2
       with:
@@ -64,7 +64,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: [3.11.1]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
@@ -110,7 +110,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: [3.11.1]
     needs: [build_pypi_wheel, build_deb]
     steps:
     - uses: actions/checkout@v2
@@ -125,7 +125,7 @@ jobs:
         ref: ${{ github.ref }}
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         wait-interval: 10
-        check-name: 'pytest (3.9)'
+        check-name: 'pytest (3.11.1)'
 
     - name: Download built dist as artifact
       uses: actions/download-artifact@master
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.9]
+        python-version: [3.7, 3.11.1]
 
     steps:
     - uses: actions/checkout@v2
@@ -81,7 +81,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.9]
+        python-version: [3.7, 3.11.1]
 
     steps:
     - uses: actions/checkout@v2
@@ -141,7 +141,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.9]
+        python-version: [3.7, 3.11.1]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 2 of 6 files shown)


Elliria:
All I did was update the Python versions. All tests passed locally with **tox**.

Elliria:
This is about those dependencies. We're going to need help from someone who understands what needs to be done before we can do any more commits to this project.

Elliria:
I thought I'd add that all tests do not pass locally. I had been cloning the repository incorrectly and was cloning the master branch instead of the development branch. That has now been corrected and the tests fail in my local copy as well.

Elliria:
Also:

Lint(3.10.1) fails at "Set up Python 3.6" with this error message: "Error: Version 3.6 with arch x64 not found"

pytest(3.10.1) fails at "Set up Python 3.10.1" with this error message: Error: Version 3.10.1 with arch x64 not found"

test-install (3.6) fails at "Set up Python 3.6" with this error message: "Error: Version 3.6 with arch x64 not found"

Each of those errors points to [this GitHub page](https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json) that shows all available versions. From what I can see, the versions mentioned do offer **arch x64**, so I don't know what to make of it.

josephj11:
The fact that it's still trying to do anything with Python 3.6 is a problem. AFAIK, your changes should have made that go away.

The entry for 3.10.1 says "20.04" and older, so that's probably why it's not found. It looks like you have to get to 3.11.2 to get "22.04", but I really don't know what I'm looking at.

Elliria:
Good catch. Very few versions on that page will work for Ubuntu 22.04. Python **3.12** is still alpha, so I don't know if we should support it just yet. What do you think?

Also, as for it testing for Python **3.6**, I'm completely confused. I've either missed at least one reference to an unsupported version or, as a really wild guess, is it possible that GitHub is testing the **master** branch, too, as part of the pull request and is failing it for that reason?

Meanwhile, I'll comb through the files and see if I can spot something I've overlooked.

Thanks for all of your patience and your amazing research skills.

Elliria:
These are the currently-supported versions:
```
3.7.13
3.7.15
3.8.12
3.8.13
3.8.14
3.8.15
3.8.16
3.9.12
3.9.13
3.9.14
3.9.15
3.9.16
3.10.4
3.10.5
3.10.6
3.10.7
3.10.8
3.10.9
3.10.10
3.11.0-beta.1
3.11.0-beta.2
3.11.0-beta.3
3.11.0-beta.4
3.11.0-beta.5
3.11.0-rc.1
3.11.0-rc.2
3.11.0
3.11.1
3.11.2
3.12.0-alpha.2
3.12.0-alpha.3
3.12.0-alpha.4
3.12.0-alpha.5
```
I had chosen **3.7.0** and **3.11.1** for my changes and **3.7.0** isn't supported. Thanks to you, I now know how to work that file, so the next attempt will be **3.7.13** and **3.11.2**, but I've got a bit of real-life to do and will probably do that tomorrow. Either way, I'll be closing this pull request as well. One of these days, I hope to get this right.

josephj11:
3.12 - we don't have any need to do bleeding edge. If they change some syntax, it could cause many changes. If they change semantics, it's much worse. If we come later, most things we run into will either already be fixed or lots of people will have encountered them and be available for help.

Just because I'm not good at Python doesn't mean I'm not a good programmer. (I'm not saying you implied that.) I once knew or used around 20 different languages including three assembly languages. I needed to acquire a lot of debugging skills to survive that experience. Most early programming (of my generation) involved compiled languages. Compiled languages are a little easier because they catch more syntax errors earlier, but far harder because you don't have the source at run time and have to look at things like memory dumps when things go wrong. Assembly language is the hardest because its semantic content is very low (most of it relies on comments) and you have to understand the physical hardware instruction set and what's physically connected to what and how. However, you can make assembly language more tractable by writing macros that expand into larger pieces of code and using those was what made me go looking for something like AutoKey in the first place.

A wise man, Edgar Cayce, once said that this universe was built upon time, space, and patience. ;) There tends to be a lot more of the first two available most of the time.

If you were to insert some pure garbage into the master branch (temporarily!) and then ran your tests, it would be clear whether the master was considered because the garbage would blow up the tests if it was encountered.

Elliria:
I agree on not needing bleeding edge. Besides, once we get these failed checks sorted out and make sure we know all the correct steps to take to update the Python versions, it will be a quick and simple matter to update them later when the new versions are out of alpha and beta or old versions are dropped.

Elliria:
I think you're great at programming and I'm heavily admiring your research skills when it comes to figuring out these errors. You've got an eagle-eye that I wasn't blessed with.

josephj11:
My kindergarten teacher gave me that knickname, LOL  but it didn't go anywhere because I didn't have much interaction with the other kids back then.

Having had to stare for long periods of time at many masses of initially incomprehensible information helps develop the skill. :)

Elliria:
Ah, then you've been doing this your whole life. No wonder you're so good at it.

PR #777, opened 2023-02-20
Update python-test.yml - revert to focal runner

userdocs:
short term immediate fix

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   Lint:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     strategy:
       matrix:
@@ -77,7 +77,7 @@ jobs:
 
   pytest:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     strategy:
       matrix:
@@ -137,7 +137,7 @@ jobs:
   test-install:
     # Just runs basic app options to ensure pip installation has included relevant imports and put app in path.
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     strategy:
       matrix:


Elliria:
@josephj11, this is a pull request from a nice person on IRC in the GitHub channel who has made a fix that should make it so we can commit to the project from now on. It reverts **master** to Ubuntu Focal so it will pass all tests and we will need to fix **develop** to do so as well, but that can come later.

josephj11:
OK. That makes sense. Doesn't sound like it will have any adverse effects.

Should I merge it?

Do you know what to do for develop? And how do we then go forward for jammy, etc.?

userdocs:
because you had `ubuntu-latest` it that was linked to 20.04 at the time this action was created, things broke when [ Github changed latest to 22.04](https://github.com/actions/runner-images#available-images)

So reverting this in master is the fastest way to make it work without overhauling the whole thing, as you see here

https://github.com/userdocs/autokey/actions/runs/4227238818

Lots of warnings and actions requiring updates to fix first as well.

Then start the process of migrating to latest.

Elliria:
This pull request is a temporary change. If I understand what needs to happen, it's safest to keep the **master** branch reverted to Focal for now simply so that it will work. Meanwhile, we can start the upgrade process in the **develop** branch, updating its workflows to work with Jammy.

To do that, we will need to replace **foo** in this `https://packages.ubuntu.com/search?suite=jammy&arch=any&searchon=names&keywords=foo` link with each of the package names in the [apt-requirements.txt](https://github.com/autokey/autokey/blob/develop/apt-requirements.txt) file to do a search for them and remove any packages that have no candidate for Jammy. Otherwise, we will get an error.

Note that this clean-up process is likely to introduce errors. We may have to go into other project files to make similar changes if the packages are referenced elsewhere. We may also have to make additional changes as a result of the impact of the removals, and possibly add new packages in their place.

We will also need to use [https://pypi.org/](https://pypi.org/) to search for each of the libraries/modules listed in the [pip-requirements.txt](https://github.com/autokey/autokey/blob/develop/pip-requirements.txt) file and check their versions and requirements, updating those if necessary, and possibly also updating the [apt-requirements.txt](https://github.com/autokey/autokey/blob/develop/apt-requirements.txt) file with any resulting package changes.

There will be a lot more to do after that. Apparently, the path from Focal to Jammy is going to be one bug at a time.

If we can get this sorted out, we'll no longer need to revert to Focal in the **master** branch. When that time comes, we can undo the changes from this pull request to put the **master** branch back the way it was by changing the three instances of **20.04** in the [.github/workflows/python-test.yml](https://github.com/autokey/autokey/blob/master/.github/workflows/python-test.yml) file back to **latest**. Meanwhile, we can create an issue that reminds us to do that before the next release or when the **develop** branch is stable (whichever comes first).

All of this information is thanks to [https://github.com/userdocs](https://github.com/userdocs), who was able to see what was wrong, provide a temporary fix, and do a great job of sharing the details above. I plan on doing my best to dive into it all and help to make it happen.

Elliria:
It also looks like [the available GitHub Actions Runner Images](https://github.com/actions/runner-images#available-images) are going to be an invaluable resource for a lot of what we need to do and would also have been useful in determining the Python versions [kicking self in head]. That page is now permanently bookmarked over here.

josephj11:
All of that sounds lovely (and a lot of work!). What I'm still not clear on is: should I merge this PR or wait for you to do other things first?

userdocs:
merge this now.

Check it works. 

If not, revert pr.

josephj11:
@Elliria OK. As soon as you're ready to work on it, I'll merge this. I just don't want to merge it and have it sit idle/untested (on our end) for a while when it might need to be reverted.

Elliria:
If we merge this pull request now (which I recommend, too), we'll be able to successfully merge pull requests without the errors that we've been getting. 

If we don't merge it in, all pull requests will continue to fail the tests, because our code base isn't configured for Jammy yet and that's what GitHub is currently using.

We won't need to revert this merge, at least not in the classic sense. What will happen is that I'll do my best to figure out how to configure our project for Jammy, testing it all out in the **develop** branch. 

Once that's done and working, I can do a new pull request on this same file and just change the three instances of **20.04** back to **latest** and you can merge it in. It will then be as though this had never happened.

Also, if you'd like me to give pull requests a test-drive after this merge, I can make a small needed change to the **CHANGELOG.rst** file and do a quick pull request. If that goes through without errors, we're in business.

josephj11:
Looks good. Go ahead with your work. I don't see a need to test this separately from what you are already doing.

josephj11:
> It also looks like [the available GitHub Actions Runner Images](https://github.com/actions/runner-images#available-images) are going to be an invaluable resource for a lot of what we need to do and would also have been useful in determining the Python versions [kicking self in head]. That page is now permanently bookmarked over here.

Could you elaborate a bit? I don't undestand what this is about.

Elliria:
The tests that we've been failing are GitHub actions. GitHub has a collection of images. I assume an image is like an ISO of an operating system. Those are the images that GitHub actions can use.

When we specify **latest** in the file from that pull request you just merged, we're referring to whatever GitHub currently considers to be **ubuntu-latest** in [this Available Images chart](https://github.com/actions/runner-images#available-images). Up until a few weeks ago, **ubuntu-20.04** was considered to be **ubuntu-latest** in that chart. GitHub recently changed it to **ubuntu-22.04**, at which point our pull requests started to fail the tests. This is because our code is configured for Ubuntu 20.04 and not Ubuntu 22.04.

Not only will that chart be a great resource as far as making it very easy for us to quickly check which image is currently the latest (especially if anything goes wrong again and we're wondering why), but the **Included Software** column is glorious. If you click on **ubuntu-22.04** in that column, it takes you to the [this GitHub page](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md), which contains a veritable feast of information about GitHub's expectations of the packages and version numbers associated with that release of Ubuntu (and that image, specifically). We'll be able to repeat that to get fresh information any time new releases roll out.

Now it's a matter of going back and forth between that page, the [Ubuntu Packages Search](https://packages.ubuntu.com/search?suite=jammy&arch=any&searchon=names&keywords=dbus) page, the [Python Package Index](https://pypi.org), and possibly also the [Ubuntu manifest](https://releases.ubuntu.com/jammy/ubuntu-22.04.1-desktop-amd64.manifest) (which lists all the packages included in the specified Ubuntu release) and comparing package and library/module names and versions against what's in our code to see what needs to be changed. I've already begun that process in my local code.

PR #778, opened 2023-02-22
workflow python-test.yml updated to work with ubuntu-latest

userdocs:
- All actions updated to current version major
- commented - trigger workflow manually
- commented - Only run on py extension push
- yaml and workflow formatting
- Python version 3.6 change to 3.7 this is the oldest supported version for 22.04, see - https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
- not squashed so you can see the changes better in the stages they were applied.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -4,31 +4,34 @@
 name: Python test
 
 on:
+  # workflow_dispatch:
   push:
+    # paths:
+    #  - "**.py"
     # Test on tags to ensure that the version metadata has been updated in tag as well.
     tags:
       - "*"
-    branches: [CI, master, develop, beta]
+    branches: [master, develop, beta]
   pull_request:
     branches: [master, develop, beta]
 
 jobs:
   Lint:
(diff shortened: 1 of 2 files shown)


Elliria:
Oh, my goodness. You've done it! So, we're done upgrading?

Elliria:
By the way, why did you get rid of CI?

userdocs:
There is no CI branch in this repo - https://github.com/autokey/autokey/branches

I've done the bare minimum to make it work with 22.04, to eliminate warnings and run without error. 

This should be all that is needed to modernise the workflow and not try to clean up or interfere with anything else.

Elliria:
This is **wonderful**! [virtual hug] Thank you **so** much for your help! You're an absolute angel!

BlueDrink9:
Outstanding! Thank you for sorting that out!

The `CI` was there intentionally, the idea being you could check out a CI branch off master for testing changes to the CI. Probably fine to remove it, easy to put back I guess

Issue #772, opened 2023-02-16
Change sudo apt to sudo apt-get in the source code


Elliria:
### AutoKey is a Xorg application and will not function in a Wayland session. Do you use Xorg (X11) or Wayland?

Xorg

### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

Enhancement

### Which AutoKey GUI did you use?

Both

### Can you briefly describe the issue?

The source code uses `apt` in a few places where it should use `apt-get` to ensure the use of stable code and to prevent the "apt does not have a stable cli interface" warning.

### Can the issue be reproduced?

Always

### What should have happened?

All references to **sudo apt** should be changed to **sudo apt-get** in these files:

* [/debian/build.sh](https://github.com/autokey/autokey/blob/develop/debian/build.sh)
* [github/workflows/python-test.yml](https://github.com/autokey/autokey/blob/develop/.github/workflows/python-test.yml)
* [.github/workflows/build.yml](https://github.com/autokey/autokey/blob/develop/.github/workflows/build.yml)
* [INSTALL](https://github.com/autokey/autokey/blob/develop/INSTALL)

Elliria:
This should happen in **master** and **develop** and **beta** since it affects each of them.

Elliria:
The **develop** branch is done.

Elliria:
The **beta** and **master** branches are now done, so this issue is complete and can be closed.

PR #781, opened 2023-02-23
Change apt to apt-get to satisfy issue #772.

Elliria:


Elliria:
I'm still getting errors [sigh]. Is this because **develop** hasn't yet been upgraded?

josephj11:
Something is clearly wrong because, AFAIK, we ditched Python 3.6 all together as being too old, so why is it even being tested?

I'm no git expert, but I woud assume that since main and develop are separate branches, what you fixed on main wouldn't affect develop unless you do some sort of merge or the PR cherry picking @BlueDrink9 mentioned elsewhere.

I don't think develop is like an OOP class that inherits things from its parent -  main (except when it was initially forked and got everything). If that was the case, you'd probably never see messages about bits of main being ahead of develop.

Elliria:
This was a test to see if I can do PRs without errors now that **master** is sorted out. In this PR, no fixes were put into place for the Python versions being out-of-date. All I changed were the **apt** references.

I'm trying repeatedly to pass the tox tests in a clone of my fork of the AutoKey project and it's failing each time. I've copied [the changes](https://github.com/autokey/autokey/pull/778) that @userdocs did to **master** over to my fork's **develop** branch and done the steps in issue #773, but the tox tests are still failing.

If I, instead, clone AutoKey's **develop** branch with `git clone --branch develop --single-branch https://github.com/autokey/autokey.git` and just change the three instances of **ubuntu-latest** in the `.github/workflows/python-test.yml` file to **ubuntu-20.04** in an effort to freeze **develop** at that Ubuntu version for the purposes of the tests (instead of trying to upgrade it) and then run `tox`, the test fails with:
```
ERROR: could not install deps [pytest, pytest-cov, pytest-xvfb, pyhamcrest>=1.8.1, PyQt5, dbus-python, PyGObject]; v = InvocationError("/home/elliria/Desktop/autokey/.tox/test/bin/python -m pip install pytest pytest-cov pytest-xvfb 'pyhamcrest>=1.8.1' PyQt5 dbus-python PyGObject", 1)
```

I'm in over my head and when I do searches on that error, I get advice that's suited for full-blown motorcycles while what I need is advice for training wheels or even a tricycle. I suspect that it might actually pass if it were done in a virtual machine that has Ubuntu 20.04 installed in it, but that's my best guess and I haven't tested that yet.

All of this is making me strongly suspect that I'm not what you need at all. I'm more of the little errand-girl who can do things like making the changes in this PR from **apt** to **apt-get**, fixing formatting errors that cause preview issues, fixing lint errors, etc. There's plenty of that sort of work that needs to be done on this project, so I would still consider those sorts of contributions worth doing.

At the moment, however, we're in need of someone who can understand what they're looking at and realize what's needed to sort out these PR issues. Otherwise, we won't be able to move forward. I'm pretty sure that the only way that anyone can submit a PR that doesn't fail right now is by making it against **master**, and that's something we don't want to do.

userdocs:
You would need to bring the changes I made over to develop.

This one to revert from latest (which is the same issue in develop that master had to begin with) 

https://github.com/autokey/autokey/pull/777

Or this one to upgrade it to latest, making 777 obsolete.

https://github.com/autokey/autokey/pull/778

Pick one, apply the full changes from the chosen PR to the develop branch and see what happens.

Elliria:
> You would need to bring the changes I made over to develop.
> 
> This one to revert from latest (which is the same issue in develop that master had to begin with)
> 
> #777

I currently have this in the local clone of my fork, but it also contains the **apt** and **apt-get** changes I made in this pull request. That's the code that's giving me the errors in my previous comment.

I'll start over with a fresh fork and only do the #778 change to it, since that's what we really want.

Should I not bother with the local **tox** tests? I've done the #777 and the #778 changes in separate attempts already yesterday and each of them failed the tests, so I tried other stuff. All attempts so far don't get me past those tests.

Also, which order do I do the commits in? From top to bottom or bottom to top? I'm guessing that they should be done from bottom to top:

778commits

userdocs:
The order make no difference here as you would squash them and consider it a single set of changes, as you see in the PR, the overall outcome of these commits.

https://github.com/autokey/autokey/pull/778/files

You can just copy and replace the files from master to dev, the outcome will be the same

Commit the changes and see if the test in the PR pass.

Elliria:
Okay, that sounds good. I'll do that now. I'll destroy the current fork first, so this PR will be closed. Thank you so much for the help!

PR #785, opened 2023-02-24
Upgrade develop branch to work with ubuntu-latest.

Elliria:

PR #791, opened 2023-02-27
Update development status and Python version.

Elliria:


Elliria:
I'm very curious whether this fixes the commit issue.

josephj11:
I don't do merges - especially to master - without an explicit request to do so.

Elliria:
I'm waiting to hear back from @BlueDrink9 on [this](https://github.com/autokey/autokey/pull/785#issuecomment-1448731853) before deciding whether to make the explicit request. I got the impression that he doesn't want [line 185 in setup.py](https://github.com/autokey/autokey/blob/master/setup.py#L185) to be changed. If he says it shouldn't change, I'll have to redo this PR.

BlueDrink9:
You know you can modify PRs, right? Just push the change to the same branch?

Elliria:
I didn't know that, but will try it.

Elliria:
Like this?

Elliria:
Or I could just scrap the fork and do a nice clean PR.

josephj11:
> Or I could just scrap the fork and do a nice clean PR.

I think that's what squash is for. It will probably be useful other times, so figuring it out on a tiny change like this where it's trivial to figure out if it worked as expected might not be a bad idea.

Elliria:
Okay, I learned squash this morning, but don't think it would be the right thing to use for what I did here since it seems to be for combining commits you do what to keep. In this case, I'm trying to undo commits I don't want to keep. I'll close this and do it over.

PR #794, opened 2023-02-27
Update metadata in setup.py.

Elliria:
* Update Development Status in setup.py to 0.96.0.
* Minor formatting clean-up of CHANGELOG.rst.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Important misc changes
 ----------------------
 
 - Update Development Status in setup.py to 0.96.0.
+- Fix visibility of code block in CONTRIBUTORS.rst.
 
 Features
 ---------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,17 +2,14 @@
 Changelog
 =========
 
-..
-   Version Develop
-   =======
-   Bug fixes
-   ---------
-
 Version Develop
 ============================
 
-.. Important misc changes
-.. ----------------------
+Important misc changes
+----------------------
+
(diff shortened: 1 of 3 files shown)


Elliria:
These are all for the **develop** branch, but [this one's](https://github.com/autokey/autokey/pull/794/commits/fa61f01968301513ff2dd1d54616256e6b857c9d) change to [line 206 in **develop**](https://github.com/autokey/autokey/blob/develop/setup.py#L206) is the same as the change to [line 185 in **master**](https://github.com/autokey/autokey/blob/master/setup.py#L185) that I'm waiting to hear back on from @BlueDrink9, so this one should wait, too.

BlueDrink9:
Both should keep version as 4 for beta

Elliria:
> Both should keep version as 4 for beta

Okay, I'll undo those. But...

[cue "Wheel of Fortune" music]
I'd like to buy a "Why?"

We're not currently in beta, are we? Or are we technically in beta the moment we make any change after a tagged release?

josephj11:
This makes sense to me.

As soon as you make changes to a released branch in develop, it's no longer that release, but it's not the next release either. It's in limbo, so labelling it as Beta makes sense. That has nothing to do with actually creating a Beta branch ATM, but when develop is ready to go (for anything beyond trivial changes), the next step would be to create a Beta branch for QA, so that label will be appropriate.

Elliria:
> As soon as you make changes to a released branch in develop, it's no longer that release, but it's not the next release either. It's in limbo, so labelling it as Beta makes sense.

Yeah, it does seem appropriate.

Elliria:
Doing this one over.

PR #796, opened 2023-03-02
Fix visibility of code block.

Elliria:
Adjust white-space on code-block to make it visible.


Elliria:
Please merge this one, @josephj11. It's a syntax change that makes a code block that was hidden visible, which will make that part of the file less confusing to the reader.

PR #797, opened 2023-03-03
Add labels and other-labels sections to issue template.

Elliria:


Elliria:
Please merge this one, @josephj11. It's a copy of the issue template I had done in my dummy repository and contains only the changes pertaining to the two new sections we want in the form.

PR #802, opened 2023-03-05
Update apt to apt get.

Elliria:
This satisfies changes in **master**, which is one of the three branches described as needing it in issue #772.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,8 +38,8 @@ jobs:
           ${{ runner.os }}-
     - name: Install dependencies
       run: |
-        sudo apt update
-        sudo apt install $(cat apt-requirements.txt)
+        sudo apt-get update
+        sudo apt-get install $(cat apt-requirements.txt)
         python -m pip install --upgrade pip
         pip install flake8 pytest pytest-cov wheel
         pip install -r pip-requirements.txt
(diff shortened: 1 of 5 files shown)


Elliria:
Please merge this one in, @josephj11. It satisfies changes in **master**, which is one of the three branches described as needing it in issue #772.

josephj11:
Noticed this yesterday: On the commits list above, they all say verified, but only the last one has a green check mark next to it. Any idea why it's like that?

I would have merged this yesterday because it looked obvious, but I waited for your request. I want to avoid problems, so I'm not guessing about merges or any other git/GitHub actions.

Elliria:
No idea. I did it the same way I've been doing the others (aside from that one that gave us the **patch_1** branch). Also, it had all green check-marks for me. I wonder why it's showing you something different. It didn't cause any errors during the merge, did it?

I'm planning on doing the **beta** in the same way this evening or tomorrow morning. It will be interesting to see if the green check-marks show up for you on that one or not.

Elliria:
By the way, I have a feeling that the green check-marks are never going to get old for me since we had so much trouble in initially getting them. I wait and watch for them when I do these PRs and still have an absolute "squee" moment every time I see them. If they didn't all show up, I'd comment as part of the PR. I know we're both still finding our way in all of this and are on high-alert as a result, and that's probably a good thing.

PR #804, opened 2023-03-07
Change apt to apt-get.

Elliria:
This satisfies changes in **beta**, which is one of the three branches described as needing it in issue #772.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,8 +38,8 @@ jobs:
           ${{ runner.os }}-
     - name: Install dependencies
       run: |
-        sudo apt update
-        sudo apt install $(cat apt-requirements.txt)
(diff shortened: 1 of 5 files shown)


Elliria:
Please merge this one in, @josephj11. It satisfies changes in **beta**, which is one of the three branches described as needing it in issue #772. Also, I got all green check-marks. What do you see at your end?

josephj11:
I thought we already did this, but I pushed it anyway. (Maybe the last one was to develop.)

I suspect that what I am seeing (all commits verified, but a check mark only on the last one) is some sort of static display because you did the commits and the PR, so no update/retest was triggered by me doing the merge.

Elliria:
We've been doing it on three different branches. That was the last of it and I closed the issue for it as completed as well.

PR #808, opened 2023-03-09
Rephrase Wayland warning and formatting.

Elliria:
- Rephrase the Wayland warning for readability.
- Reformat the warning in the "Installation" section to match the Wayland warning.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,7 +22,7 @@ About
 =====
 `AutoKey`_, a desktop automation utility for Linux and X11, formerly hosted at `OldAutoKey`_. Updated to run on Python 3. 
 
-**Important**: This is an X11 application, and as such, will not function 100% on distributions that default to using Wayland instead of Xorg.
+**Important**: This is an X11 application and, as such, will not function correctly when Wayland is in use instead of Xorg.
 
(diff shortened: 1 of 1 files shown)


Elliria:
Please merge this in, @josephj11. I did it from the AutoKey repository instead of my fork, so it will create the **patch-1** branch, but that's fine, and will be deleted once the PR is merged. This contains the rewording you suggested and the comma-fix I missed.

I also decided to reformat the warning in the **Installation** section just beneath that Wayland warning to match the formatting of the Wayland warning and make that part of the document less "busy" or distracting.

## Before:
We had something like this, where the two warnings were kind of competing with one another and Wayland was losing:

**Important** This is the Wayland warning.

**This is the Installation warning.**

## After:
We now have something like this, which I find more aesthetically pleasing:

**Important**: This is the Wayland warning.

**Important**: This is the Installation warning.

## But wait... To complicate matters:

We could have something like this if you like:

**Important: This is the Wayland warning.**

**Important: This is the Installation warning.**

Or we could have something like this if you like:

**This is the Wayland warning.**

**This is the Installation warning.**

Issue #227, opened 2018-12-06
Update or remove outdated files in the repository root directory


luziferius:
## Classification:

Bug, Enhancement

## Summary

Update the supplementary files in the autokey repository root to the current state.
Currently, several files are out-of-date and unusable.
Update: 
- [ ] - `autokey.spec` The openSUSE (?) package build specification is still on version 0.94.1. Update it to be able to build RPMs of the newest version
- [ ] - `extractDoc.py` This was seemingly used to extract the Qt autocompletion specification. This is still old Python 2 code and contains the old source path. Update this to be able to update the autocompletion automatically.
- [ ] - `PKG-INFO` This is still on version 0.90.4 and should either be updated or removed.
- [x] - `TODO` This TODO file isn’t really helpful and can probably be removed. https://github.com/autokey/autokey/commit/3f4bf23f8ad140def989ede70db6b29f74e9499d
- [x] - `INSTALL` is slightly outdated, mentions `python` instead of `python3` https://github.com/autokey/autokey/commit/bfa96d1a339e26bd3ae67787920cbc1a0b9a0601

## Version 

AutoKey version: all up to the current git master and develop branch

sebastiansam55:
RE: `extractDoc.py` I tried a few different ways of updating it but found the way that worked best (avoiding the pains of circular imports) was to create a script that basically did the same thing. 
```
import inspect
import re


module_names = [keyboard, clipboard,engine,dialog,system,window,mouse, highlevel]
module_names_str = ["keyboard", "clipboard","engine","dialog","system","window","mouse", "highlevel"]
method_regex = re.compile('Usage: C{(.*?\()')

def get_completions(module, outFile):
    for name, attrib in inspect.getmembers(module):
        if inspect.ismethod(attrib) and not name.startswith("_"):
            doc = attrib.__doc__
            for line in doc.split("\n"):
                match = re.search(method_regex,line)
                try:
                    print(match.group(1))
                    outFile.write(match.group(1)+"\n")
                except:
                    #no match was found in the line
                    pass
        elif inspect.isfunction(attrib) and not name.startswith("_"):
            doc = attrib.__doc__
            for line in doc.split("\n"):
                match = re.search(method_regex,line)
                try:
                    print(match.group(1))
                    outFile.write(match.group(1)+"\n")
                except:
                    #no match was found in the line
                    pass

with open("/home/sam/Desktop/api.txt", "w+") as outFile:
    for module in module_names_str:
        outFile.write(module+"\n")
[...]

PR #809, opened 2023-03-12
Bump AutoKey version in two files.

Elliria:
- Bump Python version in **autokey.spec** AND **PKG-INFO**.


Elliria:
Please merge this in, @josephj11. It will satisfy part of issue #227.

PR #811, opened 2023-03-12
Replace contents of beta README.rst with development notice.

Elliria:
Replace the contents with a notice letting visitors know this is not an active branch and linking them to a resource that will let them know about where and how development is done.


Elliria:
Please merge this one if you're happy with it, @josephj11. Otherwise, feel free to let me know if there's something I should change.

josephj11:
I'm merging this because it won't hurt anything, but why am I not deleting the entire branch?

Elliria:
That's up to you. I knew you were thinking of doing it, but since it's still there and you had mentioned doing this to the [README.rst](https://github.com/autokey/autokey/blob/beta/README.rst) file, I figured I'd do it so that we at least make it known that we don't want anything happening there.

Would you like me to try deleting the **beta** branch? This pull request is offering me the **Delete branch** button that will remove it from my fork. I have no idea whether that will cause it to be deleted from the AutoKey repository, but we could try if you like.

Elliria:
Here's what I see on this page now that the PR has been merged:

DeleteBranchNotice

Elliria:
Just curious, but if we deleted the **beta** branch, would we note that in the **master** branch's [CHANGELOG.rst](https://github.com/autokey/autokey/blob/master/CHANGELOG.rst) file or would it just go quietly into the night?

josephj11:
That's weird. You shouldn't have permission to do anything with the main or develop branches. It will probably delete it from your private GitHub repository.

If you want to try it, I'm sure we'll learn something in the attempt. And, if something unexpected happens it won't be on the public branches and we can avoid it when I do the actual beta.

Before you try anything questionable, make backups of anything you don't want to lose.

The beta is not quite a public branch. it sort of is when it's active. I don't see any reason to annouce it or document it in any public facing manner.

Elliria:
I should have the power to do anything I like inside of my fork, because I own it. I'm not sure what will happen when I sync with AutoKey after deleting it, though. What we're talking about is deleting the **beta** branch, which we're pretty certain (totally certain?) we no longer need. If you're certain that we don't need it and that we can make a copy of **develop** and make that the new **beta** branch when it comes beta-time, then there's no risk in my trying to delete it from my fork. If I do and nothing happens at your end, you'll have to be the one to delete it. If it works, we'll know that it can be done that way. Let me know if you're sure you want me to do it.

Also, just out of curiosity, can you go to [https://github.com/autokey/autokey/branches](https://github.com/autokey/autokey/branches) and take a screenshot of what you see for me? This is what I see when I do that on my fork:

branches

josephj11:
[This](https://www.dropbox.com/s/gy8lbbb4rf26kjk/GitHub_Branches.png?dl=0) is what I see.

josephj11:
Go ahead and try to delete the beta branch. We'll see what happens.

Elliria:
Thanks. You can drag those into here if you like. I'll drag yours in so we can see them on the same page:

WhatJoeSees

Elliria:
> Go ahead and try to delete the beta branch. We'll see what happens.

Okay. Catch you on the other side.

Elliria:
It's gone from my fork, but still exists in the AutoKey repository. That might change when I do my next pull request. If not, you'll need to delete it at your end.

josephj11:
That's what I suspected would happen.

josephj11:
If I haven't heard any objection in a while, I will delete the beta branch and the one that didn't look useful (I'll go back and remind myself which one that was.)

josephj11:
What differences of interest are you seeing between my list of branches and yours?

Elliria:
I was most interested in the **Your branches** section. Apparently, it says that to each of us for the same two branches, so since you're the owner of the AutoKey repository and I'm the owner of the fork, each of us owns those branches in our respective copies of the repository. You also get to see the deployments (the green check-mark and red x symbols), but I can't see those in my fork.

Elliria:
> If I haven't heard any objection in a while, I will delete the beta branch and the one that didn't look useful (I'll go back and remind myself which one that was.)

I would wait to delete the branch that didn't look useful until after @sebastiansam55 and/or @BlueDrink9 and/or @luziferius has/have a chance to chime in on those, because neither of us can be certain about what should be gotten rid of vs what should still get worked on.

BlueDrink9:
Don't delete beta. CI workflows make use of it; tags on the beta branch automatically release beta versions. We want it there so devs know about it. Mentioning the above in contributors.rst might be a good idea if I forgot to put it in there already.

This PR wasn't the best idea (but well-intentioned, which is welcome!) Now, if we merge beta without remembering to revert this change, it'll say in master that "this branch isn't in use". It's a mistake waiting to happen. I'm unclear on the reasoning behind it: beta is a non-default branch, so I would hope anyone that lands on it would understand enough about git to compare it to master and make their own decisions about what they should use. I don't see the reason for putting a specific note in the readme. Most people will assume the readme is the same as master (for the reason I mentioned at the start of this paragraph).

This isn't a criticism of you at all, just an fyi and a question with my thoughts

Elliria:
I'm glad you jumped in, @BlueDrink9. We'll leave **beta** in place.

Maybe we should replace the **README.rst** in [beta](https://github.com/autokey/autokey/blob/beta/README.rst) with the **README.rst** we're working on in [develop](https://github.com/autokey/autokey/blob/develop/README.rst) since the **beta** one was also outdated. Once @josephj11 merges the PR with the final changes to that one, I can do the same to **beta**.

We're not fully sure of the states of each of the branches and how they get into those states:
* **beta** is 37 commits behind and 14 commits ahead of **master**.
* **develop** is 48 commits behind and 307 commits ahead of **master**.

That seems like quite a mess to untangle. Shouldn't it be impossible for **develop** to be further behind **master** than **beta**?

Also, for safety's sake, should we just ignore **beta** and do all our updates to **develop** or should we take anything we're confident in in **develop** and do it to **beta**, too?

josephj11:
I think doing everything twice (in develop and beta) just opens up all sorts of possiblities for erorrs. Develop might also include things that are "good", but not intended for the next release. I'm not sure how we handle anything like that, but if there are any, they shouldn't be in beta.

When we are ready to make a beta "release", we should make it in an empty branch or a complete copy of develop minus anything that need to be held back. No second guessing if everything is updated, etc.

josephj11:
If there's any reasonable way to resolve  these "problems" with the various branches being out of sync/sequence on commits, we should definitely do that. Something's bound to get lost or added by accident if we don't.

josephj11:
@BlueDrink9 This brings up the main question: what is the procedure for creating a new beta and then a new release?
Is this docmented anywhere?  I was assuming that we'd just copy develop to beta and be done unless there are things to be held back, but your comment implies doing some sort of merge.

BlueDrink9:
Don't know if it is documented, probably not if not in CONTRIBUTERS. Process would probably be reset --hard to master to ensure they are in sync, then cherry-pick or merge any commits from develop we want to include. Alternatively, hard reset beta to develop then treat it as a frozen develop

Elliria:
Thanks for the information. I've noted it in my personal wiki and will be adding it to the [CONTRIBUTORS.rst](https://github.com/autokey/autokey/blob/develop/CONTRIBUTORS.rst) file once I understand it. How does **reset --hard to master** look, exactly, in action, and where are you when you do it and what will it do?

[This YouTube video](https://www.youtube.com/watch?v=PRDUK2WRzjo) shows it being used to throw out certain activities. The host used `git reset --hard 06df67a` to leapfrog over the two most-recent commits in his reflog and then did a push to the remote repository, which then only received everything except for those two commits that he had leapfrogged over. I'm assuming it can also be used in different ways.

BlueDrink9:
Hard reset basically forces a branch to be the same as another one at that point in time. `git reset --hard master` would set the current branch to a snapshot of master - as if you had checked out master then branched off it

Elliria:
Ah, thanks. That makes sense. I'll mess around with this in a dummy repository to see how it behaves.

PR #824, opened 2023-03-28
Fix for capitalized characters not sending in phrases

sebastiansam55:
Need to send a list to __send_keycode_with_modifiers_pressed

### DIFF
--- a/lib/autokey/interface.py
+++ b/lib/autokey/interface.py
@@ -720,10 +720,10 @@ class XInterfaceBase(threading.Thread, AbstractMouseInterface, AbstractWindowInt
                     self.__sendKeyCode(keyCode, theWindow=focus)
             if offset == 1:
                 self.__send_keycode_with_modifiers_pressed(
-                    keyCode, Key.SHIFT, focus)
+                    keyCode, [Key.SHIFT], focus)
             if offset == 4:
                 self.__send_keycode_with_modifiers_pressed(
-                    keyCode, Key.ALT_GR, focus)
+                    keyCode, [Key.ALT_GR], focus)
             if offset == 5:
                 self.__send_keycode_with_modifiers_pressed(
                     keyCode, [Key.SHIFT, Key.ALT_GR], focus)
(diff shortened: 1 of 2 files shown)


sebastiansam55:
@josephj11 idk how you want to handle review for PRs for me since I have merge access

for this one it's pretty straight forward, it doesn't work before the change but does after. 

looks like the issue was introduced in a refactor
https://github.com/autokey/autokey/pull/606/files#r1150960214

sebastiansam55:
It also looks like there is some issue with phrases and SendModes other than Keyboard. 

Using "Clipboard Ctrl+Shift+V" will put `<ctrl>+<shift>+v` instead of the correct value, other sendmodes seem to do the same

josephj11:
I don't know enough to decide what to merge, etc. I do it when someone tells me to - because there was no one else to do it. If @BlueDrink9 has stepped down as lead developer, then you're next in line of succession - not because you agreed to to that or want that responsibility, but because there's no one else who knows as much as you do.

Youl've been with the project for a long time and don't break things, etc., so I trust you.

What you want to do is up to you as it is in any FOSS project run exclusively by unpaid volunteers.

josephj11:
I assume this is an issue in the develop branch only. It does not affect 0.96.0.

This is exactly the sort of thing we have to be able to deal with when making any changes to our development/release strategy.

This large piece of work by @BlueDrink9 had to be added to a common code base somewhere so others could test it and find problems like this that were likely to occur in such a large set of changes. But, because we have develop, it didn't break production. Without develop, big changes like this would be ill advised. Things would have to be done in smaller chunks to manage risk, etc.

josephj11 (approved):
This change is simple enough for me to understand. If it works as you say it does, I see no reason not to merge it into develop. Either one of us can do that, but if someone runs a git blame later and it points back to me, that isn't particularly helpful.

It's still a good idea to ask for input before doing merges, etc. because maybe we'll see something or have questions, etc. The problem is that neither @Elliria nor I have much knowledge of the AutoKey codebase to be able to really review anything technically.

PR #825, opened 2023-03-28
Add documentation and <clipboard> macro.

sebastiansam55:


josephj11:
What does this do if the clipboard is empty?

Do you have an opinion about a variant of this that works with the PRIMARY selection?

I'm not sure how to use this new feature yet, but it looks promising.

I recently posted a script on Gitter that creates a markdown link using the clipboard for the URL and the PRIMARY selection for the label. It might be cool to do that in a phrase.

While I'm thinking about it, can the `<clipboard>` macro be used more than once in the same phrase? Seems like that ought to work. Not sure how I would use that though.

sebastiansam55:
Great suggestion, added in https://github.com/autokey/autokey/commit/05743b37440d4aa95a656c4b58cd28b477b932f5

`<clipboard clipboard="selection">` to use it

PR #829, opened 2023-04-01
Abstract mouse/window/keyboard interfaces 

sebastiansam55:
# TLDR
With my recent discovery that I can poll the mouse location and interact with windows easily (on gnome shell at least) it opens the path toward (hopefully) easily port AutoKey over to full support on Wayland with minimal to no loss of major features

# STATUS
Abstracted Window interface, working with AutoKey gnome shell extension to get autokey working on Wayland/gnome distros (Ubuntu 22.04 for ex). 

The longer term plan is to have things setup internally so that we can easily add additional interfaces for different display servers and different desktop environments. 

https://github.com/sebastiansam55/autokey-gnome-extension

This GNOME shell extension is a fork of one that provides a number of dbus methods that allow us to interact with windows under wayland. We can get the active window information and move/resize/close/maximize/minimize. 

It will also allow us to poll the location of the mouse pointer, with this we can use uinput to move the mouse in x y until it is approximately in the correct location. TBD How hard that will actually be to make look smooth/fast. 

Unfortunately Wayland does not allow the pointer to be warped.

# TODO

- [ ] Add checking that Autokey Gnome Extension is properly installed
- [ ] Add config file/settings option to set window interface
- [x] ~~Add additional dbus functions to extension for mouse interactions~~ Wayland does not allow anything other than polling location
- [ ] Abstract mouse interface
- [ ] implement mouse movement via uinput events
- [ ] implement mouse clicks
- [ ] finalize and publish gnome shell extension
- [ ] update tests for window and other files
- [ ] kde has plasma extensions but I need to do more research on it.
 
Something needs to be done about the Button and Key classes in model. They need to be abstracted in someway that will allow `Key.LEFT` to correspond to the correct scan code/key code per the active keyboard system. 

# New

## uinput events
For hotkeys, under uinput we should be able to capture **any event**, so everything from foot peddles to mouse buttons, gamepads and keyboards we should be able to record and listen for as an event for a hotkey. Very cool.

## left/right modifiers
I am also going to strive to bring left and right differention to both the hotkey modifiers and the keyboard send system, have to either bring support for that to x11 keyboard or make it backwards compatible. Setting the generic version to left makes the most sense.

IE: For x11 there is currently only `Key.SHIFT` with no diffentiotion between left/right. For uinput I'll be trying to set it to respect all `Key.LSHIFT` `Key.RSHIFT` and `Key.SHIFT`. the last would be sent as left shift

## internal changes
Mouse, Keyboard, window interfaces are all going to be separated out to be independent of each other, with this done it will be easier to implement support for Wayland.

# Scenarios:

## GNOME/Wayland
- [x] windows via GNOME extension
- [ ] keyboard
  - [ ] read
  - [ ] write
- [ ] mouse
  - [x] read via GNOME extensions
  - [ ] write

## KDE/wayland
- [ ] windows via plasma/KDE extension?
- [ ] keyboard
  - [ ] read
  - [ ] write
- [ ] mouse
  - [ ] read
  - [ ] write

## [GNOME / KDE]/x11
- [x] windows
- [x] keyboard
  - [x] read
  - [x] write
- [x] mouse
  - [x] read
  - [x] write

# CHANGES 
`window` is going to change scripting API wise, will be breaking change, ultimately it's my fault for not being forward looking when adding those methods, they were tightly specific to x11

`mouse` will likely evolve, right now there's really only move and click but you don't send a click or don't send a move if you don't want to do both, would like to split those out

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ __pycache__
 *.egg-info
 *.pyc
 .tox/
+.eggs/
 
 # IDE's
 .idea
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ Important misc changes
 - Various updates to the **README.rst** file to satisfy issue #681.
 - Various updates to the **README.rst** file to satisfy #pullrequestreview-1336342159.
 - Update action versions in build.yaml to latest.
+- Breaking changes to Window scripting interface to be more cross compatible.
+- Adds partially implemented 'tk' interface.
 - Update Qt/GTK "Run" button in interface to run on F5
 
 Features
@@ -25,6 +27,8 @@ Bug fixes
 - Fix crash in qt macro recording window.
 - Fix fake keyboard events not being emitted in a timely manner in some cases
 - Upgrade the **develop** branch to satisfy issue #773.
+- Splits missing program checks to display server (x11, wayland) specific
+- Update scripting/window to use iomediator instead of interface directly
 
 Version 0.96.0-beta.9
 ============================
--- a/lib/autokey/UI_common_functions.py
+++ b/lib/autokey/UI_common_functions.py
@@ -24,10 +24,13 @@ gtk_modules = ['gi', 'gi.repository.Gtk', 'gi.repository.Gdk', 'gi.repository.Pa
 qt_modules = ['PyQt5', 'PyQt5.QtGui', 'PyQt5.QtWidgets', 'PyQt5.QtCore',
             'PyQt5.Qsci']
 
-common_programs = ['wmctrl', 'ps', 'xrandr']
+# wmctrl, xrandr are x11 specific programs.
+x11_programs = ['wmctrl', 'xrandr']
+common_programs = ['ps']
 # Checking some of these appears to be redundant as some are provided by the same packages on my system but 
 # better safe than sorry.
-optional_programs = ['visgrep', 'import', 'png2pat', 'xte', 'xmousepos']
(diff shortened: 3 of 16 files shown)


josephj11:
This is quite encouraging although I don't pretend to understand the specifics. Let us know (on Gitter, etc.) if there's anything we can do to help.

chrislee35:
Thanks for kicking this off, @sebastiansam55 — this draft basically set the direction the project ended up going for Wayland/GNOME support, even though it didn't land as this specific PR.

Checking in on it today, everything this PR set out to do has since been achieved independently, some of it by you:

- `lib/autokey/scripting/abstract_window.py` and `window_gnome.py` exist on `develop` today, but via commit `99c870e` ("Windows Interface"), pushed directly to `develop` in September 2024 — not through this PR.
- The window/mouse/keyboard abstraction layer this PR wanted (`sys_interface/abstract_interface.py`) actually predates this PR — BlueDrink9 added it back in 2021.
- Mouse movement under Wayland via uinput — the biggest open TODO here — is fully implemented in `lib/autokey/uinput_interface.py`'s `UInputInterface.move_cursor()`.
- The GNOME shell extension approach lives on as `autokey-gnome-extension/` (dlk3, December 2024).

Given all of that, plus this being a 3.5-year-old draft with no reviews that GitHub now reports as unmergeable (conflicting against current `develop`), I don't think there's anything left to salvage here versus what's already on `develop`. I'll close this out as superseded — thanks again for pointing the project in this direction.

josephj11:
Closed as superseded

Issue #779, opened 2023-02-22
Add a CodeTriage badge to AutoKey


Elliria:
### AutoKey is a Xorg application and will not function in a Wayland session. Do you use Xorg (X11) or Wayland?

Xorg

### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

Enhancement

### Which AutoKey GUI did you use?

None

### Can you briefly describe the issue?

We need to add a CodeTriage badge, like this [![Open Source Helpers](https://www.codetriage.com/autokey/autokey/badges/users.svg)](https://www.codetriage.com/autokey/autokey) one, to the AutoKey project now that [AutoKey is listed there](https://www.codetriage.com/autokey/autokey).

### Can the issue be reproduced?

N/A

### Anything else?

Here's the Markdown:
```
[![Open Source Helpers](https://www.codetriage.com/autokey/autokey/badges/users.svg)](https://www.codetriage.com/autokey/autokey)
```
Here's the image URL in case we need to put it inside of an HTML file:
[https://www.codetriage.com/autokey/autokey/badges/users.svg](https://www.codetriage.com/autokey/autokey/badges/users.svg)

josephj11:
Where would we put this? We don't have a project website other than our GitHub page.

Elliria:
Perhaps in the [CONTRIBUTORS.rst](https://github.com/autokey/autokey/blob/develop/CONTRIBUTORS.rst) file and/or perhaps in the [Bug reports and Pull Requests](https://github.com/autokey/autokey/blob/develop/README.rst#bug-reports-and-pull-requests) section of the [README.rst](https://github.com/autokey/autokey/blob/develop/README.rst) file.

Elliria:
This is done in the [README.rst](https://github.com/autokey/autokey/blob/beta/README.rst) file in the **beta** branch.

Elliria:
This is done in the [README.rst](https://github.com/autokey/autokey/blob/develop/README.rst) file in the **develop** branch.

Elliria:
This is done in the [README.rst](https://github.com/autokey/autokey/blob/master/README.rst) file in the **master** branch.

Issue #681, opened 2022-03-22
Update README.rst


josephj11:
### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

Documentation

### Which Linux distribution did you use?

N/A

### Which AutoKey version did you use?

0.96.0 Beta 10

### How did you install AutoKey?

N/A

### Can you briefly describe the issue?

Needs updating and to be made more user friendly

### Anything else?

Documentation section should first link to our wiki. New in this release/new in this major version should be added with more descriptive text. This should be a bit more promotional ("look at these great new features" ... )

API documentation link should point to the new improved format document.

Elliria:
If you're talking about the first link in the document, it currently points to the AutoKey main project page at [https://github.com/autokey/autokey](https://github.com/autokey/autokey) and I can see how that could be disconcerting to visitors who aren't used to seeing repository file listings.

Would pointing it lower down on the same AutoKey main project page, like [https://github.com/autokey/autokey#readme](https://github.com/autokey/autokey#readme) work? That way, the newcomers would have a more user-friendly greeting while those of us who are comfortable with the repository file listings or who would like to get at them would know to scroll up.

Elliria:
I missed updating the API link when I did the recent update and will do that after the current PR is merged.

josephj11:
The links I'm talking about are in the document under the Documentation heading. It looks like they have mostly been fixed already. One that I missed is on the next line about Examples. The wiki link there should point to https://github.com/autokey/autokey/wiki/Example-Scripts instead of to the head of the wiki.

The Running unit tests section should be deleted. It needs to be added to CONTRIBUTORS.rst (maybe under the Testing section). The section here should be replaced by a contributors section that links to CONTRIBUTORS.rst after a few lines of welcoming/encouragement.

new_features.rst needs a complete rewrite. It is extremely old. We might want to just delete it and replace it with a wiki page for ease of editing. The example code it contains needs to be saved unless there is already better code in our wiki.

CONTRIBUTORS.rst is also very old and needs a rewrite. It says "I" when it should say "we", etc.

```
Just open a PR for discussion and I’ll try to answer any questions that arise. I suggest writing new features on top of develop.
```

should be 


```
Just open a PR against the develop branch for discussion and we’ll try to answer any questions that arise.

We prefer to leave our main branch relatively untouched between releases as a reference for current issues and user support.
```

[...]

PR #830, opened 2023-04-02
Update README.rst file to satisfy part of issue #681 and issue #779.

Elliria:
Update the README.rst file in beta to match the one in develop, and then note the change in the CHANGELOG.rst file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,16 +18,16 @@ jobs:
       matrix:
         python-version: [3.10.1]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
 
     - name: Cache pip
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
         # This path is specific to Ubuntu
         path: ~/.cache/pip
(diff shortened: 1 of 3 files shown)


Elliria:
This is the update to the [README.rst](https://github.com/autokey/autokey/blob/beta/README.rst) file in the **beta** branch to make it match the [README.rst](https://github.com/autokey/autokey/blob/develop/README.rst) file in the **develop** branch.

This is also the update to the actions in the **beta** branch's [build.yml](https://github.com/autokey/autokey/blob/beta/.github/workflows/build.yml) file to match the updated versions in the **develop** branch's [build.yml](https://github.com/autokey/autokey/blob/develop/.github/workflows/build.yml) file.

This is also updates to the [CHANGELOG.rst](https://github.com/autokey/autokey/blob/beta/CHANGELOG.rst) file to note those changes.

josephj11:
I will merge this whenever you request it.
I assume that it will continue our commits ahead/behind problems that still need to be resolved systemically.

Elliria:
> I will merge this whenever you request it.

Let's not do it that way any more. A pull request is already a request, so it's redundant to do an additional request on top of that. I don't mind writing a description for each pull request s I've been doing, so I plan on continuing to do that.

> assume that it will continue our commits ahead/behind problems that still need to be resolved systemically.

Yes, and although it will add to the versioning mess, the diffs will either be identical or more similar this way (depending on whether there were differences other than that, which I didn't check). It will also make one less thing to have to do once the **beta** branch does get used.

Heads up that there are more changes coming for the **beta** and **master** branches as part of these updates, but I'll note them as I do them.

josephj11:
OK. Feom now on, I will merge your PRs when they come in. I will review them as much as I can and let you know if I have any concerns before merging them.

PR #831, opened 2023-04-03
Update two links in the README.rst file in the develop branch.

Elliria:
* Update link in the **Contributing or modifying the source** section to point to the **master** branch instead of the **develop** branch.
* Update the link in the **Changelog** section to point to the **master** branch instead of the **develop** branch.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Important misc changes
 - Various updates to the **README.rst** file to satisfy #pullrequestreview-1336342159.
 - Update action versions in build.yaml to latest.
 - Update Qt/GTK "Run" button in interface to run on F5
+- Update two links in the **README.rst** file.
 
 Features
 ---------
--- a/README.rst
+++ b/README.rst
@@ -85,13 +85,13 @@ Contributing or modifying the source
 Pull requests are welcome from anyone who would like to modify or contribute to the source code. Useful tips for working with and testing the code can be found in the `CONTRIBUTORS.rst`_ file. AutoKey also participates in `CodeTriage`_, where members can sign up to receive a periodic email with a link to an open AutoKey issue that needs help.
 
 .. _CodeTriage: https://www.codetriage.com/autokey/autokey
-.. _CONTRIBUTORS.rst: https://github.com/autokey/autokey/blob/develop/CONTRIBUTORS.rst
+.. _CONTRIBUTORS.rst: https://github.com/autokey/autokey/blob/master/CONTRIBUTORS.rst
 
 Changelog
 =========
 Our `changelog`_ is the best source of information for what's new and fixed in each release.
 
-.. _changelog: https://github.com/autokey/autokey/blob/develop/CHANGELOG.rst
+.. _changelog: https://github.com/autokey/autokey/blob/master/CHANGELOG.rst
 
 License
 =======


Elliria:
I realize we thought we were done with this file, @josephj11, but when I did the same change to the [README.rst](https://github.com/autokey/autokey/blob/beta/README.rst) file in the **beta** branch, I noticed the three links at the bottom and realized they should all point to the **master** branch no matter which branch the **README.rst** file is in, so this file in the **develop** branch needed to have two of its links fixed. Once this one is merged, those two will be identical and the one in **master** branch will need to be done.

josephj11:
If either of these two linked files (and any other similar files) have been updated since the master release, why wouldn't the links point to the updated files in the branch they're in?

Is there a way to start a link with "this branch" (a relative link like in file paths) so it doesn't break when the file is moved to the next more permanent branch?

Elliria:
> If either of these two linked files (and any other similar files) have been updated since the master release, why wouldn't the links point to the updated files in the branch they're in?

For several reasons. If you read the three links at the bottom of the **README.RST** file in the **beta** and **develop** branches, they point to:
* the **CONTRIBUTORS.rst** file as a reference for folks who would like to contribute to the project
* the **CHANGELOG.rst** file as the best source of information for what's new in and fixed **in each release**
* the **LICENSE** file 

Taking those on one at a time:

Most visitors to the site will encounter and deal with the files in the **master** branch first, since those are the defaults. Anyone who does development will know to go poking around in the various branches to see if there are differences in the files they're interested in.

The **CONTRIBUTORS.rst** file should be identical in all three branches since it's not branch-specific, but is intended for use as a general reference.

The **LICENSE** file should always be identical in all three branches for obvious reasons, and when that changes (which it will from time to time), it will need to be updated in all three branches at the same time, also for obvious reasons.

The only file that should vary is the **CHANGELOG.rst** file, since those are branch-specific. That's irrelevant for the purposes of the **README.rst** file, though, for three reasons:
* First, and foremost, is that the **README.rst** file links to the **CHANGELOG.rst** file as the best source of information for what's new and fixed **in each release** and since the **master** branch is intended to have the least amount of change in it at any given moment, its changes will remain the closest to what they were when the current release was tagged, whereas the others may vary wildly. 
* Second, regardless of how odd our branch structure is, it's my guess that, at some point, **beta** and **develop** will somehow be merged into **master** ...or... **beta** will replace **master** (we have yet to find out how that's done in this project). In either case, when those get merged in, the individual branch-specific links in the **README.rst** file that would then be on **master** would point to either the **beta** or **develop** (as the case may be) branch and would need to be immediately changed. This would be extra work, an extra step to be aware of, an extra point of failure if it was overlooked or neglected, and a potential source of confusion for users or developers. For thoroughness' sake, I should add that the differences could be handled as conflicts during the merges, but these would be yet more (avoidable) conflicts that would each have to be dealt with on an individual basis as part of creating a release.
* Third, if we decided to go ahead and use individual branch-specific links that required us to have to do clean-up after each merge that results in a new release, the moment we made the necessary immediate update to the **README.rst** file on the **master** branch, it would then be 1 commit ahead of the other branches, starting the cycle of messy commits all over again.

For all of those reasons, I felt it was a good idea to just choose **master** for those links and be done with it. We're trying for the most stable possible environment under our given circumstances, and this is one small step in that direction.
 
> Is there a way to start a link with "this branch" (a relative link like in file paths) so it doesn't break when the file is moved to the next more permanent branch?

Last, but not least, there's no way to create a relative link since the **README.rst** file has no awareness of its own existence or its place in the world, so those links must be specific.

josephj11:
TL;DR: OK. I'm good with this.

Thanks for the very detailed response.

I have some minor quibbles with your position, but none  significant enough to even discuss.

IANAL, but I believe that since our license is stated as "GPL 3.0", we really can't ever change it. The license itself precludes making it more restrictive and since we didn't specify it as "GPL" or "GPL 3.0 or later", i think it's completely static. This supports your ideas here.

(AFAIK, the only time the version really made a difference was in the transition from GPL 2 to GPL 3 because GPL 2 disallowed the use of licensed material with or for anything that supported DRM (digital rights management) and GPL 3 allowed it. Since AutoKey doesn't address DRM at all, even this probably woudn't matter for us anyway.)

Elliria:
I'm not a lawyer, either, and I don't know if something that gets released under the GPL stays at that version of the GPL forever regardless of updates to the GPL or if it uses the updated versions. Since you're the keeper of the keys, it seems like it might be important for you to know, so we might want to look into it to find out what the "best practices" for such things are.

PR #833, opened 2023-04-06
Update the actions in the build.yml file.

Elliria:
This will make all the actions in the [build.yml](https://github.com/autokey/autokey/blob/master/.github/workflows/build.yml) file in the **master** branch current with those in the [GitHub Marketplace](https://github.com/marketplace).


Elliria:
All three branches will have had this done once this is merged in. I haven't found an issue mentioning this file, specifically, but will look again and update it if there is one.

Issue #682, opened 2022-03-22
Update man pages


josephj11:
### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

UI/Usability

### Which Linux distribution did you use?

N/A

### Which AutoKey version did you use?

0.96.0 Beta 10

### Can you briefly describe the issue?

Man file does not list `--version` option.

Not sure if anything else is needed.

### Anything else?

Add --version option.

Check if anything else has changed.

Elliria:
We have three man pages in the [/autokey/doc/man](https://github.com/autokey/autokey/tree/develop/doc/man) directory. From what I can see, this is what's needed, but I'd love some feedback:

1. Correct the punctuation.
2. Correct the spelling.
3. Make the author reference identical in all three files.
4. Change **configuration window** to **main window** to make it obvious what will happen and since that's what that window is called in all other documentation.
5. Add missing single-character command-line switch to the **--help** switch.
6. Add these new command-line switches with these definitions to the **OPTIONS** section:
   * [--cutelog-integration](https://github.com/autokey/autokey/blob/master/lib/autokey/argument_parser.py#L57-L62)
     * Connect to a locally-running **cutelog** instance with default settings to display the full program log. See the **cutelog** GitHub page at [https://github.com/busimus/cutelog](https://github.com/busimus/cutelog) for more information.
   * [-m, --mouse](https://github.com/autokey/autokey/blob/develop/lib/autokey/argument_parser.py#L63-L68)
     * Enable verbose (debug) logging and include mouse-button events.
   * [-V, --version](https://github.com/autokey/autokey/blob/develop/lib/autokey/argument_parser.py#L52-L56)
     * Print the current AutoKey version to standard output and then exit.
7. Alphabetize the **OPTIONS** section.
8. Update the date.

*Note that I don't see anything else in [the list of what's new in AutoKey 0.96.0](https://github.com/autokey/autokey/releases/tag/v0.96.0) that belongs in a man page. The rest will go nicely into the [new_features.rst](https://github.com/autokey/autokey/blob/develop/new_features.rst) file, which will also be getting an update. I did, however, find some more goodies not mentioned there when I looked through the [argument_parser.py](https://github.com/autokey/autokey/blob/develop/lib/autokey/argument_parser.py) file, so those have been added above.*

Elliria:
[...]

PR #839, opened 2023-04-16
Update the man pages.

Elliria:
Fix, update, and add content to the [autokey-gtk.1](https://github.com/autokey/autokey/blob/develop/doc/man/autokey-gtk.1), [autokey-qt.1](https://github.com/autokey/autokey/blob/develop/doc/man/autokey-qt.1), and [autokey-run.1](https://github.com/autokey/autokey/blob/develop/doc/man/autokey-run.1) files to make all three man pages current.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,8 @@ Important misc changes
 - Update two links in the **README.rst** file.
 - Fix, update, and add content to the man pages.
 - Fix Qt reference and update the wording in reference to KDE and Qt in the **autokey-qt.1** man page.
+- Update the date and remove excess wording in the **autokey-gtk.1** and **autokey-qt.1** man pages.
+- Update the date in the **autokey-run.1** man page.
 
 Features
 ---------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Important misc changes
 - Update action versions in build.yaml to latest.
 - Update Qt/GTK "Run" button in interface to run on F5
 - Update two links in the **README.rst** file.
+- Fix, update, and add content to the man pages.
 
 Features
 ---------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Important misc changes
 - Update Qt/GTK "Run" button in interface to run on F5
 - Update two links in the **README.rst** file.
 - Fix, update, and add content to the man pages.
+- Fix Qt reference and update the wording in reference to KDE and Qt in the **autokey-qt.1** man page.
 
 Features
 ---------
--- a/doc/man/autokey-gtk.1
+++ b/doc/man/autokey-gtk.1
@@ -2,7 +2,7 @@
 .\" First parameter, NAME, should be all caps.
 .\" Second parameter, SECTION, should be 1-8, maybe w/ subsection.
 .\" Other parameters are allowed: see man(7), man(1).
-.TH AUTOKEY-GTK "1" "April 18, 2023"
+.TH AUTOKEY-GTK "1" "April 19, 2023"
 .\" Please adjust this date whenever revising the man page.
 .\"
 .\" Some roff macros for reference:
@@ -41,7 +41,7 @@ options start with two dashes (`--').
 A summary of options is included below.
 .TP
 .B \-c, \-\-configure
-Show the AutoKey main window on start-up.
+Open the AutoKey main window.
 .TP
 .B \-\-cutelog-integration
 Connect to a locally-running \fBcutelog\fP instance with default
--- a/doc/man/autokey-gtk.1
+++ b/doc/man/autokey-gtk.1
@@ -2,7 +2,7 @@
 .\" First parameter, NAME, should be all caps.
 .\" Second parameter, SECTION, should be 1-8, maybe w/ subsection.
 .\" Other parameters are allowed: see man(7), man(1).
-.TH AUTOKEY-GTK "1" "April 16, 2023"
+.TH AUTOKEY-GTK "1" "April 18, 2023"
 .\" Please adjust this date whenever revising the man page.
 .\"
 .\" Some roff macros for reference:
@@ -41,8 +41,7 @@ options start with two dashes (`--').
 A summary of options is included below.
 .TP
 .B \-c, \-\-configure
-Show the main window on start-up or at any time while the program is
-running.
+Show the AutoKey main window on start-up.
 .TP
 .B \-\-cutelog-integration
 Connect to a locally-running \fBcutelog\fP instance with default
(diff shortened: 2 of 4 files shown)


Elliria:
Hey there. All three man pages have been updated. You can preview the changes in the source code with the GitHub interface or, if you'd like to view the man pages as you would view any man page on your system, you can go to the [doc/man](https://github.com/Elliria/fork/tree/develop/doc/man) directory on my fork (where the changed files currently are), and copy the contents of the files into files on your computer and then open the directory you put the files into and run the `man ./foo` command for each file, replacing **foo** with the name of the file.

I've also zipped up the three original files, the three changed files, and a **README.txt** file listing the changes and providing viewing instructions and am attaching that zip file here in case you'd like to see them that way:

[man-page-changes.zip](https://github.com/autokey/autokey/files/11242482/man-page-changes.zip)

Elliria:
Also, once this has been merged in, it will satisfy issue #682.

josephj11:
doc/man/autokey-gtk.1
39 why the dash in command-line?

doc/man/autokey-gtk.1
30 Qt - not QT

30 Qt is independent of KDE. I don't think we do anything specific to KDE other than use Kdialog, so I'm not sure KDE should be explicitly mentioned here.
44 If autokey-qt is invoked while it is already running, AFAIK, it will display the main menu even if -c is not specified as long as --version is not specified.

I have no idea what it will do if --verbose is specified or if it is invoked while autokey-gtk is running. These are so far  off the edge that I have never thought of them before.

60 I have never seen the --mouse switch before. When did that get added? How do you use it without making the log huge?

Elliria:
> doc/man/autokey-gtk.1 39 why the dash in command-line?

This is in all three man pages and is because **command-line** is a compound adjective that was used before the noun it describes, so it requires a dash.

> doc/man/autokey-gtk.1 30 Qt - not QT

Good catch. I'll fix that momentarily and update this PR. I got the upper-case variant from the original files, but should have caught it, especially since I use **Qt**.
 
> 30 Qt is independent of KDE. I don't think we do anything specific to KDE other than use Kdialog, so I'm not sure KDE should be explicitly mentioned here.

Since we use **KDialog**, it's fitting to mention **KDE**. I'll change the wording a bit momentarily to make it a little gentler.

> 44 If autokey-qt is invoked while it is already running, AFAIK, it will display the main menu even if -c is not specified as long as --version is not specified.

Interesting. I hadn't realized that. I just tried it with **autokey-gtk** and **autokey-qt** and they both do that. I had no idea. It doesn't belong in the man page, though, but should definitely be added to the wiki.
 
> I have no idea what it will do if --verbose is specified or if it is invoked while autokey-gtk is running. These are so far off the edge that I have never thought of them before.

Me, neither, but that isn't suggested in the man page, so it would only occur if someone happened to come up with the idea to try it on their own.
 
> 60 I have never seen the --mouse switch before. When did that get added? How do you use it without making the log huge?

I'm not sure when it was added, but it's in the AutoKey 0.96.0 source code, so it's included here, because the purpose of this PR is to make the man pages current. I assume you would use it for debugging when you want to verify whether your mouse-clicks are registering and/or if they're occurring at the correct times and you probably wouldn't be running it that way for long.

Elliria:
Should I remove "or at any time while the program is running" from the **-c** command-line switch description since it's not necessary to use that switch when AutoKey is already running?

josephj11:
I supposed it wouldn't hurt to remove that, but since it works eitiher way, it's no big deal.

Elliria:
I've removed it from the **autokey-gtk.1** and **autokey-qt.1** man pages and updated the dates on all three man pages to today.

josephj11 (commented):
Maybe this should be "Show the AutoKey main window" because it does it any time, not just on startup - from the point of view of AutoKey - not from the point of view of the current instance of AutoKey.

josephj11 (commented):
Maybe the original wording is better than this or my suggestion in the other PR.

I missed it before, but I see one test failed for some reason.

Elliria:
Good catch. Also, I think it should be "Open" instead of "Show" since "Show" would be used for displaying data in the terminal window in response to a command.

The original wording is out, because it was: "Show the configuration window on startup, even if this is not the first run." That used "Show" and contradicted itself with the "on startup" part. I like your reduced version better as having the power of the "less is more" philosophy.

I hadn't noticed [the failed test](https://github.com/autokey/autokey/actions/runs/4732008871/jobs/8397628895) either. Interesting. It says the operation was canceled. I have no idea what would cause it to be canceled, but can see why that would be marked as a failure. Let's see if the next test (which will happen shortly after I make the above change and update all the dates again) goes through fine.

Elliria:
Okay, I've changed that line in the **autokey-gtk.1** and **autokey-qt.1** man pages to "Open the AutoKey main window." I've also updated the date in all three man pages.

Last, but not least, I decided not to update the **CHANGELOG.rst** file because it already contains the entries from yesterday's changes, which are identical to the entries I would put in for today's changes and would just add duplicate entries.

josephj11 (commented):
Just curious: It says you don't need verbose if you specify mouse. I assume that works, but I'vje never tried the mouse stuff at all.

44 is the text you wanted to improve.

Elliria:
I got the information about the mouse command-line switches from the [this section in the argument_parser.py](https://github.com/autokey/autokey/blob/master/lib/autokey/argument_parser.py#L63-L68) file. I don't run AutoKey 0.96.0, so I haven't tested it. From its description, I take it that it will give you the usual verbose logging and then also note any detected mouse events as part of the output.

I did improve line 44. Is that not showing for you? Here are the diffs for the most recent commits, with the text highlighted in green being my most recent edits:

* [The GTK man-page diff](https://github.com/autokey/autokey/pull/839/commits/651415620c84c41d35ebea7e844e7b914d67253b)

* [The Qt man-page diff](https://github.com/autokey/autokey/pull/839/commits/2f19456d64c4debbeb633db52564ea04d7ca6910)

I got to those by clicking the commit comments in the most recent list of commits just above these three messages.

You can also see their current state [in the **develop** branch on my fork](https://github.com/Elliria/fork/tree/develop/doc/man) since that's where I made the changes and where this pull request is coming from.

I also highly recommend copying the contents of each one into a text file (**foo.txt** in this example) and then viewing it as an actual man page with `man ./foo.txt` as a much gentler way of looking them over. All that markup makes the source files nasty to read.

josephj11 (commented):
I'm looking at this one [7897289](https://github.com/autokey/autokey/pull/839/commits/78972897b8f4d512c54936aee0459a055f304e74) which shows a partial correction. No idea why we're looking at different things. Maybe my browser is caching pages? The commit hash doesn't match the one in your link.

Oops! there was another set of commits  lower in this thread that I didn't see.
I'm looking again.

Elliria:
Yep, you got it. The originals never change and every new commit gets added inline below. They probably should have had it all congregating at the top of the page, instead, with perhaps some labels to identify what came first and what's new. That might be a good suggestion to make to GitHub.

PR #842, opened 2023-04-21
Change clipboard wording in the documentation for clarity.

Elliria:
* Update the wording for [the clipboard entry](https://github.com/autokey/autokey/blob/master/lib/autokey/scripting/clipboard_gtk.py#L101) in the [/lib/autokey/scripting/clipboard_gtk.py](https://github.com/autokey/autokey/blob/master/lib/autokey/scripting/clipboard_gtk.py) file.

* Update the wording for [the clipboard entry](https://github.com/autokey/autokey/blob/master/lib/autokey/scripting/clipboard_qt.py#L89) in the [/lib/autokey/scripting/clipboard_qt.py](https://github.com/autokey/autokey/blob/master/lib/autokey/scripting/clipboard_qt.py) file.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -104,6 +104,7 @@ Other changes
 - Update pip installation requirements
 - Add CONTRIBUTERS.rst
 - Internal Code cleanup. The configuration handling module is split into multiple modules inside a dedicated package.
+- Change clipboard wording in the AutoKey documentation.
 - AutoKey now has a working test environment again. `pytest` based unit-tests can be launched from the source checkout using `python3 setup.py test`
 
 **New Dependencies (test-time only)**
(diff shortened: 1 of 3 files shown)


Elliria:
Hey there, @josephj11, this is the documentation change we talked about at the bottom of [this discussion](https://github.com/autokey/autokey/discussions/822) to distinguish between clipboard and selection.

Note that I inserted the **CHANGELOG.rst** comment instead of putting it at the end of the section, as we usually do, because someone did something a bit unusual in that file at the end of that section and I didn't want to disturb it, so I just put my info directly above it.

josephj11:
lib/autokey/scripting/clipboard_qt.py
"onto" should be "into" - looks like a typo.

Elliria:
They're both that way and it was deliberate, because clipboards don't have insides, so you clip things onto them. Many's the time I've been made horribly uncomfortable by the man with the clipboard next to me with his pen hovering above the check-boxes and just itching to fill them in.

josephj11:
OK.  I'm used to clipboard managers which seem like they have history "in" them, but clipboards physically don't have insides as you pointed out.

Elliria:
I missed a couple of references to **selection** in those files, so I'm about to do another pull request to take care of those for consistency's sake.

PR #855, opened 2023-05-08
Add the Environment section to the .gitignore file.

Elliria:
Add the Environment section to the .gitignore file so that virtual environments (which we will soon be required to use and may already use) will be ignored by GitHub.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,15 @@ __pycache__
 *.pyc
 .tox/
 
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
 # IDE's
 .idea
 .vscode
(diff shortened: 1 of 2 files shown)


Elliria:
The section I added came from the `.gitignore` file that GitHub creates for you if you specify a **Python** `.gitignore` file when creating a repository. These are the contents of the new section:
```
# Environments
.env
.venv
env/
venv/
ENV/
env.bak/
venv.bak/
```
The environment names on that list will be ignored. If custom names that aren't on the list are used when creating virtual environments, they won't be ignored.

josephj11:
That was fast!

This will (eventually) have to propagate  to the master branch for the next release.

Elliria:
I think they always do. I'm pretty sure the **develop** branch becomes the **master** branch when a new release is created, but I'll leave that to Those Who Know to determine.

josephj11:
I recently got a response from @BlueDrink9 which essentially said that the release methodology isn't documented anywhere within AutoKey, but that it could be done by reinitializing a new branch to a prior state (such as the current master state) and then cherry picking things from develop to implement the desired changes.

In any case, there is no automagic process to do so. It would be great if we could (gradually)  document and improve this situation.

Elliria:
I saw that and agree that it should be documented. First, we'll need to properly and fully understand all of the mechanics of it. That will most likely involve the use of a dummy repository that won't mind potentially suffering horrendous damage in the initial attempts until an air-tight process can be captured for posterity. I also agree that this could help us to dig ourselves out from under our complex branching scheme with understanding, which would be wonderful.

Issue #845, opened 2023-04-23
Update AutoKey command-line switches


Elliria:
### AutoKey is a Xorg application and will not function in a Wayland session. Do you use Xorg (X11) or Wayland?

Xorg

### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

UI/Usability

### Choose one or more terms that describe this issue:

- [ ] autokey triggers
- [X] autokey-gtk
- [X] autokey-qt
- [ ] beta
- [ ] bug
- [ ] critical
- [X] development
- [ ] documentation
- [X] enhancement
- [ ] installation/configuration
- [ ] phrase expansion
- [ ] scripting
- [ ] technical debt
- [X] user interface

### Which AutoKey GUI did you use?

None

### Can you briefly describe the issue?

The [**verbose** command-line switch](https://github.com/autokey/autokey/blob/develop/lib/autokey/argument_parser.py#L41) needs to be updated to add the **-v** switch and to deprecate the **-l** switch for the next AutoKey release, with the plan being to remove the **-l** switch in the next AutoKey release after that one.

### Can the issue be reproduced?

Always

### What are the steps to reproduce the issue?

View the man page by typing `man autokey` into a terminal window and look at the **-l, --verbose** entry in the **OPTIONS** section.

### Anything else?

As was mentioned in [issue #682](https://github.com/autokey/autokey/issues/682) and [discussion #840](https://github.com/autokey/autokey/discussions/840), the single-character command-line switch that's used to display [the AutoKey version](https://github.com/autokey/autokey/blob/develop/lib/autokey/argument_parser.py#L52-L56) requires an **upper-case** **V** instead of a lower-case **v**, which is different from the other command-line switches. In looking all the command-line switches over, we came to the consensus that we should leave that one alone and update the **verbose** command-line switch, deprecating the **-l** switch that it currently uses and adding the **-v** switch to it as a temporary addition and future replacement.

This would give us these command-line switches for those two options in the next release of AutoKey:
* -v, --verbose, -l
* -V, --version = version

We would then make it known that **-l** is being deprecated, but has been left in the next version for backwards-compatibility, but will be removed in the future release after that one.

I did a bit of research and although it would be unusual to have two short command-line switches and one long one for an option, it's not forbidden.

To make this happen, these tasks will need to be done:

[...]

PR #864, opened 2023-05-16
Update and add to the GTK and Qt man pages.

Elliria:
This is a major overhaul of the GTK and Qt man pages with these changes:
* Move the date to the top of the document.
* Update the date.
* Move TeX instructions up with the rest of the instructions.
* Add text to the **NAME** section.
* Reformat the **SYNOPSIS** section.
* Reformat the **DESCRIPTION** section.
* Remove unnecessary information from the **DESCRIPTION** section.
* Update the text in the **DESCRIPTION** section.
* Move the wiki link from the **DESCRIPTION** section to the **SEE ALSO** section.
* Update the introductory text in the **OPTIONS** section.
* Add short command-line switch to the **cutelog** option.
* Rename long command-line switch in the **cutelog** option
* Reformat the **cutelog** option.
* Update the text in the **cutelog** option's description.
* Move the **cutelog** link from the **OPTIONS** section to the **SEE ALSO** section.
* Update the text in the **mouse** option's description.
* Add short command-line switch to the **verbose** option.
* Deprecate the -l command-line switch for the **verbose** option.
* Update the text in the **verbose** option's description.
* Update the text in the **version** option's description.
* Sort the **OPTIONS** section alphabetically.
* Remove man-page-authorship attribution from the **AUTHOR** section.
* Update the text in the **AUTHOR** section.
* Add the **SEE ALSO** section.

Also, only the **Qt** man page contains a reference to **KDialog** and only the **GTK** man page contains a reference to **Zenity**.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ Important misc changes
 - Add the "Environment" section to the .gitignore file.
 - Update the help menu, deprecating one entry, adding several entries, updating existing wording, and sorting the entries.
 - Update the logger by removing an unneeded space and making the **cutelog** reference match the new command-line switch for it in the help menu.
+- Update and add to the GTK and Qt man pages (major overhaul with current command-line switches and lots more information).
 
 Features
 ---------
--- a/doc/man/autokey-gtk.1
+++ b/doc/man/autokey-gtk.1
@@ -1,9 +1,14 @@
 .\"                                      Hey, EMACS: -*- nroff -*-
+.\" Please adjust this date whenever revising the man page:
+.TH AUTOKEY-GTK "1" "May 15, 2023"
+.\"
 .\" First parameter, NAME, should be all caps.
 .\" Second parameter, SECTION, should be 1-8, maybe w/ subsection.
 .\" Other parameters are allowed: see man(7), man(1).
-.TH AUTOKEY-GTK "1" "April 19, 2023"
-.\" Please adjust this date whenever revising the man page.
+.\"
+.\" TeX users may be more comfortable with the \fB<whatever>\fR and
+.\" \fI<whatever>\fR escape sequences to invoke bold-face or italics,
+.\" respectively.
 .\"
 .\" Some roff macros for reference:
 .\" .nh        disable hyphenation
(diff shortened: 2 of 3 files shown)


Elliria:
I started writing these changes to satisfy issue #845 and ended up going hog-wild and doing a major overhaul. I hope you all like it and I welcome feedback and suggestions for improving them.

The **SEE ALSO** section was added at the bottom and the links that were up in the other sections got moved down there to join the new ones that had been added. I then decided to add even more to make the man pages sort of a "one-stop shop" for some of the more significant locations associated with AutoKey. There are more changes, but you can read all about them up above.

To check out these changes to the man pages in a gentler way than by viewing the raw source code (which is really rough):
1. Visit the [GTK man page](https://github.com/Elliria/autokey-fork/blob/develop/doc/man/autokey-gtk.1) or the [Qt man page](https://github.com/Elliria/autokey-fork/blob/develop/doc/man/autokey-qt.1) in my AutoKey fork.
2. Click the **Download raw file** icon on the right.
3. Open a terminal window in the same directory as the file(s) you downloaded.
4. Type `man ./autokey-gtk.1` or `man ./autokey-qt.1` to see them as actual man pages.

Then, for comparison, you can type `man autokey-gtk` or `man-qt` to see your currently-installed versions of those files.

josephj11 (commented):
24 "and mouse" doesn't seem quite right because our API does other things as well - like the window API module.

29 "that's customized..." "with a front end that's customized" - or something else because autokey-gtk is just the front end and both use autokey-common for everything else.

91 0.96.0 and newer

josephj11 (commented):
Similar changes as suggested for the gtk front end.

56 Why remove and then exit? That's what actually happens when called standalone and it's what a second instance of AutoKey does when another instance is already running.
(missed marking this in the gtk version)

Both versions: man x regex describes BRE and ERE regex dialect syntax. There are some differences in the dialect used by Python, so this is somewhat misleading.

See [this](https://stackoverflow.com/questions/33907156/does-the-python-regular-expression-module-use-bre-or-ere) and [this](https://gist.github.com/CMCDragonkai/6c933f4a7d713ef712145c5eb94a1816).

Elliria:
> "and mouse" doesn't seem quite right because our API does other things as well - like the window API module.

You're right. How about removing **keyboard** from the description so that the kind of automation isn't specified at all?
```
autokey-qt - automation utility for GNOME and GTK
```
...and...
```
autokey-qt - automation utility for KDE and QT
``

**Comment by Elliria** [comment]
> "that's customized..." "with a front end that's customized" - or something else because autokey-gtk is just the front end and both use autokey-common for everything else.

How about:
```
autokey-gtk is a desktop automation utility for Linux and X11 with a GTK front end for use with GNOME.
```
and
```
autokey-qt is a desktop automation utility for Linux and X11 with a Qt front end for use with KDE.
```

Elliria:
> 91 0.96.0 and newer

Looks good. I'll replace it.

Elliria:
> 56 Why remove and then exit? That's what actually happens when called standalone and it's what a second instance of AutoKey does when another instance is already running.
(missed marking this in the gtk version)

I'm confused. It's just a command that prints the AutoKey version to standard output in a terminal window and does nothing else. If there was any explanation as to what it was entering and then exiting from, I'd be all for it, but unless I'm missing something obvious, there's no exiting, because there's no entering.

I suspect that the only reason the **exit** part of the sentence was ever in there at all was because some of the information in the man page seems to have been gotten from the default first entry in the **options** section of the help text, which uses it. You can see it by typing `autokey -h` or `autokey --help` in a terminal window.

That entry is the bane of my existence when it comes to rewriting the help text and I fervently wish I could edit it (for several reasons), but they've made that incredibly difficult and/or impossible to do. I've tried lots of different ways to do it to no avail, so we're stuck with it in the help text for now. We're not stuck with it in the man page, however, since we've got full control in there, so I took the axe to it with relish.

Elliria:
> Both versions: man x regex describes BRE and ERE regex dialect syntax. There are some differences in the dialect used by Python, so this is somewhat misleading.
> 
> See [this](https://stackoverflow.com/questions/33907156/does-the-python-regular-expression-module-use-bre-or-ere) and [this](https://gist.github.com/CMCDragonkai/6c933f4a7d713ef712145c5eb94a1816).

My goodness, but what a mess. Why on earth isn't this standardized? It's hard enough to learn as it is, but giving us 15 different engines to learn is downright mean.

Anyway, how about we replace the man-page references from the **SEE ALSO** section with links to the [re -- Regular expression operations](https://docs.python.org/3/library/re.html) and [Regular Expression HOWTO](https://docs.python.org/3/howto/regex.html) pages on the official Python website?

josephj11:
And then exits - you're right. I was thinking of how -c works by mistake.

Desktop automation - you're rewrite is correct, but people might not really know what that is. IDK how to improve it without an explanation that's too long.

New regex links - good.

Regex  dialects - This is FOSS - the wild west to some extent - nothing we can do about it. From what I understand, PCRE is the most powerful one, but that's about forward and backward references that hurt my head. When I have to deal with anything context dependent, I almost immediately switch to awk where I can do things one step at a time and use real variables and degbugging print statements.

What I'm really hoping for is the enhancement someone made (@sebastiansam55 ??) that replaced the filters with user Python code - which is the same idea as using awk, but more appropriate for a Python based project. It never made it into a PR, so I don't know its current status.

sebastiansam55:
Not something that I remember working on but sounds cool. Not sure about the value proposition there, I don't use the window filters in the first place but what would be the benefit of allowing python code instead of regex?

josephj11:
@sebastiansam55 Here are a few things:
Negative filters (all windows except those that match) are a PITA to code now, especially before window title and class tests are separated (I think I saw some work on that)
Complex regexes are essentially write only code - if they don't work, it's usually easier to start over than to debug them - especially with forward and back references...

My print to file script (in the wiki) fails spectacularly when I launch it in the wrong window, but the browser window and the print dialog have the same window title and class, so a normal filter won't work - I'd have to call visgrep if I wanted to avoid that problem and that would already have committed the hotkey as triggered.

I use window filters a lot. E.g. I have two print to file scripts - for two different print dialogs and they both use Ctrl+p as their hotkey.  They were a lot easier to code separately than as one huge case statement. (And there are several other print dialogs I need scripts for that would also use Ctrl+p.)

Elliria:
I'm going to close this pull request unmerged because I made the mistake of updating my fork to incorporate @sebastiansam55's recent PR. I shouldn't have done that, because it adds an extra commit to my fork that will just create a vicious cycle of commits back and forth between my fork and the AutoKey repository without end (I've experimented on this like crazy with dummy repositories and haven't figured out how to get around it). I'll create a new PR with the latest changes in it.

PR #874, opened 2023-07-05
Fix all occurrences of "they key" typo

Elliria:
Replace all occurrences of "they key" with "the key" in the AutoKey documentation.

Five occurrences each in:
* `doc/scripting/lib.scripting-pysrc.html`
* `doc/scripting/lib.scripting.Keyboard-class.html`
* `lib/autokey/scripting/keyboard.py`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,7 @@ Important misc changes
 - Remove special handling of ignoreCase and matchCase options in abbreviation settings dialogs, allowing phrases to trigger on any input case while matching input case in the output (see #588).
 - Update the GTK and Qt man pages.
 - Update date, formatting, and NAME section in the GTK and Qt man pages.
+- Fix typo: Replace all occurrences of "they key" with "the key" in the AutoKey documentation.
 
 Features
 ---------
--- a/doc/scripting/lib.scripting-pysrc.html
+++ b/doc/scripting/lib.scripting-pysrc.html
@@ -106,7 +106,7 @@
 <a name="L48"></a><tt class="py-lineno">  48</tt>  <tt class="py-line"><tt class="py-docstring">        </tt> </tt>
 <a name="L49"></a><tt class="py-lineno">  49</tt>  <tt class="py-line"><tt class="py-docstring">        Usage: C{keyboard.send_key(key, repeat=1)}</tt> </tt>
 <a name="L50"></a><tt class="py-lineno">  50</tt>  <tt class="py-line"><tt class="py-docstring">        </tt> </tt>
-<a name="L51"></a><tt class="py-lineno">  51</tt>  <tt class="py-line"><tt class="py-docstring">        @param key: they key to be sent (e.g. "s" or "&lt;enter&gt;")</tt> </tt>
+<a name="L51"></a><tt class="py-lineno">  51</tt>  <tt class="py-line"><tt class="py-docstring">        @param key: the key to be sent (e.g. "s" or "&lt;enter&gt;")</tt> </tt>
 <a name="L52"></a><tt class="py-lineno">  52</tt>  <tt class="py-line"><tt class="py-docstring">        @param repeat: number of times to repeat the key event</tt> </tt>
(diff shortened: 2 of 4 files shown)


Elliria:
This is the first time we've updated the documentation using this new method, @josephj11. I'm curious whether the changes will be instant when you merge it.

A quick way to check would be to do a search for "**the key**" and/or "**they key**" in the [Keyboard API](https://autokey.github.io/api/keyboard.html) page immediately afterwards.

josephj11:
Ah, the wonders of copy and paste. The time proven way to propagate errors!

Elliria:
Blast. I forgot. We have to make changes in the **master** branch in order to update the documentation. I'll do these again on **master**.

josephj11:
OTJT (on the job training) LOL

Elliria:
It gets worse. Since I made the changes directly on the **master** branch, it created separate pull requests for each change, so you're about to have your hands very full with multiple pull requests.

josephj11:
So I see. No problem. We're all learning as we go. But next time we want to go from develop to master, are these going to show up as out of sync?

Elliria:
Probably. I'll get it right one of these days, though.

Elliria:
Hey there. I'm still able to find **they key** on the [Keyboard API](https://autokey.github.io/api/keyboard.html) page, which means the changes to the documentation that were made in merges #875, #876, and #877 haven't been applied to the documentation yet. I looked in the notes in my personal wiki, but don't find anything about running the action manually, but I seem to remember that that's what has to happen. It looks like the [pages-build-deployment](https://github.com/autokey/autokey/actions/workflows/pages/pages-build-deployment) action is the one that will make it happen. Is that right, @sebastiansam55? If so, that will be for you to do, @josephj11. I can see the actions, but not run them.

josephj11:
When I follow your link, I can see past actions, but not how to trigger it.
The Epydoc one has a run button in it, but not this one.

Elliria:
Whoops. I forgot. The new documentation is in its own repository instead of in the AutoKey repository. The [Build Autokey pages](https://github.com/autokey/autokey.github.io/actions/workflows/pages.yml) action should do it.

josephj11:
That worked after I hit it twice.

Got two warnings about save state being depreciated or something like that.

Elliria:
Okay, good. Now we know what we'll need to do in the future when we make changes to the documentation. It's always nice to know for sure where the walls of your box are. Meanwhile, the warnings are definitely a concern that needs to be dealt with.

Apparently, as of October of last year, GitHub has [started putting up a warning](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) for anyone who uses the `save-state` or `set-output` commands via `stdout`. The fact that you got the warning means that that action uses either or both of them.

[This YouTube video](https://www.youtube.com/watch?v=HyktX7eYqrU) explains both and what to do about them, but his example is different enough from what I see in our [pages.yml](https://github.com/autokey/autokey.github.io/blob/master/.github/workflows/pages.yml) file that I can't figure out what needs to change or what to replace it with. Hopefully, @BlueDrink9 or @sebastiansam55 will know.

josephj11:
Did a little searching:
[This](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) covers the basics.

It says to upgrade some packages and how to fix the individual commands. It looks like those commands may be in the "libraries" of actions we use, not directly in our code. If so, upstream fixes will be needed and will hopefully be available by upgrading versions.

I looked up all of the actions invoked by pages.yml. None of these include the offending statments - at least in their current versions. The last one had a commit about removing depreciated stuff, but it looked like something different.

I don't understand these actions at all, so I may have missed something. Maybe they call other actions that I didn't see.

Elliria:
I'm not sure how to upgrade action versions other than specifying them in the **build.yml** file, which this repository doesn't have.

It looks like [line 42](https://github.com/autokey/autokey.github.io/blob/master/.github/workflows/pages.yml#L42) is triggering the warning.

I don't know enough to be useful in this. Hopefully someone more knowledgeable will step in here.

Issue #448, opened 2020-10-02
Sharing variables across scripts does not work


ldorigo:
## Classification:

Bug (I think?)

## Reproducibility:

Always

## Version

AutoKey version: 0.95.10

Used GUI (Gtk, Qt, or both): QT

Installed via: pacman (or was it AUR?) on archlinux.

Linux Distribution: Arch

## Summary

In this script: https://github.com/autokey/autokey/wiki/Scripting#key-wrapper it looks as if when you use engine.run_script("foo"), the script foo has access to variables defined in the script from which it's being called. When I try to do the same it doesn't work. If it's supposed to work that way, then it's a bug (it doesn't work and gives me a non defined error).

If it's not supposed to work that way, then this is 1. a documentation error (remove that script from the wiki) and 2. a feature request to allow sharing data/variables among scripts (store.set_global could be used, but changing run_script to accept an optional parameter "data" that gets passed to a variable in the target script would be nicer).

## Steps to Reproduce (if applicable)

Create two scripts, "source" and "target".

`source`:

```python
#Enter script code
test = " hello"
engine.run_script("target")
```

`target`:
```python
print(test)
```

## Expected Results

"test" gets printed to the console

## Actual Results

I get a "NameError: name 'test' is not defined".

ldorigo:
I asked around in the gitter chat and it seems like using store.get/set_global_value is the way to go. So I think the solution here would simply be to remove that script from the wiki (or change it to reflect the proper way to do it)

josephj11:
@ldorigo The wiki got reorganized, so your link no longer works. It's [here](https://github.com/autokey/autokey/wiki/Advanced-Scripts#key-wrapper) now.

@Elliria the above example script looks like it probably never worked. It needs to be fixed or removed (if you want to look at it).

As soon as the example is fixed or removed, this issue can be closed as completed.

In AutoKey, all normal variables are in the scope of the current script.

The engine API starts a new script with its own scope, so the variables from the calling script are not available in the called script.

As you discovered, using AutoKey stored local and global variables gets around this limitation. So does importing a module/functions from the Modules directory (but those can't contain any API calls without some extra work). 
So, bottom line is that this feature works as intended, but the example code appears to be wrong.

Adding data parameters to the engine API call is an interesting idea. That should be requested in a separate enhancement request issue. I don't know the code, so I'm not sure whether that would be easy or hard to implement.

[...]

PR #885, opened 2023-07-17
Update engine.py

Elliria:
Remove extra unneeded curly bracket from [line 350 of engine.py](https://github.com/autokey/autokey/blob/master/lib/autokey/scripting/engine.py#L350) for the [API documentation](https://autokey.github.io/api/engine.html#autokey.scripting.Engine.run_script).

### DIFF
--- a/lib/autokey/scripting/engine.py
+++ b/lib/autokey/scripting/engine.py
@@ -347,7 +347,7 @@ Folders created within temporary folders must themselves be set temporary")
         """
         Run an existing script using its description or path to look it up
 
-        Usage: C{engine.run_script(description, 'foo', 'bar', foobar='foobar'})}
+        Usage: C{engine.run_script(description, 'foo', 'bar', foobar='foobar')}
 
         @param description: description of the script to run. If parsable as
         an absolute path to an existing file, that will be run instead.


Elliria:
After this is merged, @josephj11, the documentation will need to be updated by [running the action manually](https://github.com/autokey/autokey.github.io/actions/workflows/pages.yml) like you did last week.

josephj11 (commented):
How did you even find this error? Do the new docs have an error in them caused by this? It ought to break something.

It's in a blockquote comment, so Python itself won't care.

Elliria:
The OP in issue #448 mentioned a possible enhancement for the the `engine.run_script()` API call, so I went poking around in the documentation to see if that already exists. I saw this line in [the new documentation](https://autokey.github.io/api/engine.html#autokey.scripting.Engine.run_script):

```
Usage: engine.run_script(description, 'foo', 'bar', foobar='foobar')}
```
I then copied the code from it into AutoKey without noticing that errant curly-bracket:
```python
engine.run_script(description, 'foo', 'bar', foobar='foobar')}
```
I then tried to access **foobar** and got an error about an unclosed bracket. That's when I noticed that curly-bracket at the end. I deleted it from my script, leaving me with this working code:
```python
engine.run_script(description, 'foo', 'bar', foobar='foobar')
```
The script then ran without an error despite the fact that my lack of knowledge made it so that I still wasn't able to figure out how to access **foobar**. Anyway, since doing away with the curly-bracket in my script got rid of the error, I decided to do away with it in the documentation, too.

It's just documentation, so it won't break anything inside of AutoKey, but it caused me to copy something that was incorrect into my script, which could happen to anyone else as well, so it's good to clean it up.

josephj11:
I  just looked at the actions and saw past page builds, but I didn't see how to actually run it so these changes would be in the generated docs.

I'd like to see your test code for the keyword argument. Seems like it should work.

Elliria:
Whatever it is that you need to do, you did it a week ago when we fixed a typo. My notes say that there should be something here that you can click to build the AutoKey pages: [Build Autokey pages](https://github.com/autokey/autokey.github.io/actions/workflows/pages.yml)

If it's not there, maybe you have to go to the main [Actions](https://github.com/autokey/autokey.github.io/actions) from that link or by clicking **Actions** in the GitHub toolbar. You might also need to click **All workflows** in the sidebar on the left.

Whatever you need to do, please let me know the steps, so I can update my notes. This will be added to the development notes that I'm putting together so that others beyond us will also know how it's done if and/or when we need to be replaced.

Elliria:
> I'd like to see your test code for the keyword argument. Seems like it should work.

Sure. Hopefully you'll see something obvious that I overlooked:

## Test 1
Test1 code:
```python
# This is the code from the documentation (with the errant curly-bracket removed from the end):
engine.run_script(description, 'foo', 'bar', foobar='foobar')
```
Test1 result:



## Test 2

Test2 code:
```python
# This is my modification of it because I have no idea how to use the description and am getting an error if I leave it in:
engine.run_script('foo', 'bar', foobar='foobar')

# This is an attempt to use the specified variable in a dialog:
dialog.info_dialog(foobar)
```
Test2 result:



## Test 3

Test3 code from Testing1 script:
```python
# This is an attempt to run the Testing2 script, which will attempt to use the foobar variable:
engine.run_script('Testing2', foobar='foobar')
```
Test3 code from Testing2 script:
```python
# This is an attempt to use the variable that was specified in the Testing1 script in a dialog:
dialog.info_dialog(foobar)
```
Test3 result:

Elliria:
This was tested in AutoKey 0.95.10 and 0.96.0 with the same results in each.

josephj11:
Your link worked and I ran the workflow.

Wherever I was before looked the same but tlere was no run button.

Running the documentation workflow manually

Go to

https://github.com/autokey/autokey.github.io/actions/workflows/pages.yml

Then press the Run workflow button and wait

Elliria:
Thanks. I noted it in my wiki.

PR #886, opened 2023-07-17
Update CHANGELOG.rst

Elliria:
Note the change to the engine.py file for the API documentation.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Important misc changes
 - Bump Python version in **autokey.spec** AND **PKG-INFO**.
 - Update the **README.rst** file to satisfy part of issue #681 and issue #779.
 - Update the actions in the **build.yml** file to the latest versions in the GitHub Marketplace.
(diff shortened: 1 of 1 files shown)


Elliria:
GitHub just refuses to put the two changes into one PR, so you get two for this change.

josephj11 (commented):
It's not so much that it's unneeded (which it appears to be), but that it's unbalanced. Something's not right in the state of Denmark if nothing noticed a fundmental syntax error like this.

Elliria:
I'm not sure I understand what you mean by it being unneeded and unbalanced. These two PRs are just about a typo in the documentation. Somebody accidentally typed a **}** at the end of a piece of example code and nobody noticed it until today. Once you merge PR #885, which has the fix, and this PR, which has the changelog entry noting that the fix was made, it will be history.

There's no automated code that would have been capable of warning us of it because it's in the form of a comment. Also, we rely on our own eyes to catch errors and we humans are capable of not seeing things. Even though lots of us have looked at that line of code for however long it's been that way, we just never saw that little curly guy.

PR #892, opened 2023-08-02
Overhaul the CHANGELOG.rst file.

Elliria:
* Use imperative mood for all statements.
* Change formatting of some statements.
* Fix syntax for unclosed code insertion.
* Update heading levels to fix GitHub rendering issue (file is now "pretty" on GitHub, links work, and file can now be previewed when editing on GitHub).
* Fix typos.
* Fix grammar.
* Fix punctuation.
* Add missing "New window API method" section.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,144 +5,160 @@ Changelog
 Version 0.96.0
 ============================
 
+Bug fixes
+---------
 
-Important misc changes
-----------------------
+- Both Qt and GTK versions will reload hotkeys after a keymap change event.
+- Fix locking issue.
+- Expose Alt_GR as a hotkey modifier on GTK.
+- (GTK) Fix GUI lock-up, if multiple script error notifications are posted in quick succession. The notifications
+  are now rate-limited and won’t post more than one notification per second. Fixes issue #383.
+- Fix issue with pip installation reporting a missing module.
+- **Scripting API**
+   - Fix API call `system.exec_command()` crashing if output capturing is active, but the executed command has
+     empty output. Fixes issue #379.
+- **Packaging**
+   - Fix AutoKey PNG icon size. Now, the icon size is 96x96 pixels, fixing Lintian warnings on Debian. Fixes issue
+     #369.
 
-- Script and phrase metadata are no longer stored as hidden dotfiles. Existing scripts should be automatically converted, but if switch back to versions prior to this one, you will need to copy or symlink them back to dotfile form.
-- Scripting API files are now in Python packages, which may require adjusting imports if you have scripts that import them directly.
-- Change the default phrase send mode to `ctrl+v` (paste using clipboard) rather  than sending keys one at a time.
-- This version represents some significant refactoring since the previous update, so bug reports will be highly appreciated.
-- Add two sections to the issue template.
-- Change all **sudo apt** references in **master** to **sudo apt-get** to satisfy part of issue #772.
-- Bump Python version in **autokey.spec** AND **PKG-INFO**.
-- Update the **README.rst** file to satisfy part of issue #681 and issue #779.
-- Update the actions in the **build.yml** file to the latest versions in the GitHub Marketplace.
+Documentation
+-------------
+
+- Change clipboard wording in the AutoKey documentation.
+- Update formatting and wording in the AutoKey GTK and Qt clipboard API documentation.
+- Remove one reference to **X** in the AutoKey GTK clipboard API documentation.
+- Remove one reference to **mouse** from the AutoKey Qt clipboard API documentation.
+- Fix typo: Replace all occurrences of "they key" with "the key" in the AutoKey documentation.
 - Remove extra unneeded curly bracket from line 350 of engine.py for the API documentation.
-- Update the actions in the **pages.yml** file to the latest versions in the GitHub Marketplace.
+- Update the **README.rst** file to satisfy part of issue #681 and issue #779.
 
 Features
 ---------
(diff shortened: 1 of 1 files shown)


Elliria:
This is a major overhaul of the [CHANGELOG.rst](https://github.com/autokey/autokey/blob/master/CHANGELOG.rst) file on the **master** branch. The diff will look a bit frightening, but data wasn't lost. A lot of data was moved around and several changes were made (see pull-request notes above).

See how the file looks after these changes [on my AutoKey fork](https://github.com/Elliria/autokey/blob/master/CHANGELOG.rst).

**Note:** The `CHANGELOG.rst` files on each of the branches are different from one another, so I'm doing each of them separately.

josephj11:
This is really difficult to review. I assume that nothing was lost, but I can't reasonably verify that. I tried printing out both versions, but gave up comparing them almost immediately.

In the future, maybe something like this could be done on the side first in smaller steps so the changes would be easier to verify (even if the final actual PR is all at once like this one is).

A few specific comments:

The first subheading is bug fixes, but an enhancement - exposing Alt-GR is listed.
This should be moved under enhancements.

Under Documentation, the line about README.rst refers to two issues without any summary information. It shouldn't be necessary to look up those two issues to get a general sense of what was done. This is especially true because one of them is qualified with "part of".

General questions (not specific to this PR):

GUI section notes that the application now watches some files and updates itself when it changes. I would like this explored to see exactly what it does and what it doesn't do and what it might take to add this into the Qt front end as well. We should at least open an issue for that even if we can't do it now.

A crash in the Qt macro recorder was fixed. Last I heard, macro recording (which should be renamed to script recording) hasn't worked for years in either front end. I would like a status update on this feature. Even if it works fine, it needs to be checked that it still works fine relative to any new/modified features that are added. it seems like a number of our new features can't be accessed while recording a script. I'm unsure how the whole feature works (if it does work at all).

Elliria:
> This is really difficult to review. I assume that nothing was lost, but I can't reasonably verify that. I tried printing out both versions, but gave up comparing them almost immediately.
> 
> In the future, maybe something like this could be done on the side first in smaller steps so the changes would be easier to verify (even if the final actual PR is all at once like this one is).

Then I'll start over and do it that way so you can get it one small step at a time. It will be many steps.

Elliria:
> The first subheading is bug fixes, but an enhancement - exposing Alt-GR is listed.
This should be moved under enhancements.

It's in the **Bug fix** section in the current [CHANGELOG.rst](https://github.com/autokey/autokey/blob/master/CHANGELOG.rst) file. Deciding whether to put it somewhere else is an issue independent of this pull request that will require investigation. Since I didn't write that entry, I can only recommend that you look through the git history and find the commit that accompanies that CHANGELOG entry and investigate what was done on your own or use git blame to find out who made the change and contact that person to ask about it. You could also do a search through the issue to see if you can find one that mentions something like this.

If you find the issue it references, please let me know so that I can add it to this entry.

Elliria:
> Under Documentation, the line about README.rst refers to two issues without any summary information. It shouldn't be necessary to look up those two issues to get a general sense of what was done. This is especially true because one of them is qualified with "part of".

That is an issue independent of this pull request. Since I wrote that entry, though, I can tell you that it is necessary to look them up, which is exactly why I mentioned them both. CHANGELOG entries are not at all about a general sense, but about specifics.

If you had looked the two issues up, you'd have seen that the changes needed to be done to each branch, so the changes done to **this** branch would only satisfy one third of the changes required to complete the issue, which wanted all three branches to get the change.

Last, but not least, the commit that accompanied that CHANGELOG entry gave the details of exactly what was done. All of that can be found in the git history.

Elliria:
> GUI section notes that the application now watches some files and updates itself when it changes. I would like this explored to see exactly what it does and what it doesn't do and what it might take to add this into the Qt front end as well. We should at least open an issue for that even if we can't do it now.

That is an issue independent of this pull request. Since I didn't write that entry, I can only recommend that you look through the git history and find the commit that accompanies that CHANGELOG entry and investigate what was done on your own or use git blame to find out who made the change and contact that person to ask about it. Also, as you pointed out, if you have a Qt wish, you might want to open a separate issue for it.

Elliria:
> A crash in the Qt macro recorder was fixed. Last I heard, macro recording (which should be renamed to script recording) hasn't worked for years in either front end. I would like a status update on this feature. Even if it works fine, it needs to be checked that it still works fine relative to any new/modified features that are added. it seems like a number of our new features can't be accessed while recording a script. I'm unsure how the whole feature works (if it does work at all).

That is an issue independent of this pull request. Since I didn't write that entry, I can only recommend that you look through the git history and find the commit that accompanies that CHANGELOG entry and investigate what was done on your own or use git blame to find out who made the change and contact that person to ask about it.

Elliria:
Your interest in deep-dives into some of the entries made by the various developers made it clear that investigating those entries could and might involve examination of the git history. That made me realize that the entries will work as search phrases for git investigations if they remain unchanged, but would make such investigations difficult otherwise.

As a result, my new goal for overhauling the CHANGELOG will be done like this:
1. Fix the syntax of the heading levels in the document to fix the GitHub rendering issue, making the document display and behave properly
2. Organize the existing information by alphabetically-sorted categories as needed so that each version's information is in the same place.
3. Convert multi-line entries into one-liners (this will probably involve changing some of the text).
4. Fix the syntax of the one broken code-insertion (this will definitely be a change, but is necessary).

I'm going to close this issue and start over with small changes each time.

Issue #893, opened 2023-08-02
AutoKey 0.96.1 goals, ideas, and known issues...


elmarr:
Good day, I am new at this so I might be in the wrong place.
I love autokey and have been using it for years, in fact I wish it were possible to replace the autocorrect in llibreoffce and all other places, since it operates across all apps.
However, I find that it has an odd behaviour:
I use the - or = to be a trigger in a phrase, so for example, 
sa with a dash (-) will expand to South Africa (as it has just done now). But I find it sometimes expands to 
here are 6  events from LO Writer:
SaSouth Afric sa-South Afri sa-South Afri sa-South Afri sa-South Afri sa-South Afri
strangely, it seems to always work fine in Calc 
What am I doing wrong?
I started with Ubuntu in 2009 and then switched to lmLinux Mint Cinnamon: (again note the lm lefover from the lmc with dash) in 2012 and have been using that since

Elliria:
Welcome, @elmarr. You're in the right place. Your issue seems to be missing the answers to the default questions, though. I'm not sure what went wrong.

It sounds to me like you're having a timing issue. This is such a common issue that I've decided to borrow some of @josephj11's helpful text and add it to the **FAQ** page in our wiki. The [Why is the text messed up when I use a phrase with LibreOffice Writer?](https://github.com/autokey/autokey/wiki/FAQ#why-does-nothing-happen-when-i-start-autokey) question should solve the issue, but if it doesn't or if you need help in writing a script, feel free to come back in here and we'll help to figure it out.

josephj11:
When you have a question like this, it's better to start out posting on one of our [forums](https://github.com/autokey/autokey/wiki/Community). See boiler plate below.

Then, if it appears to be a valid issue, it can be filed here on GitHub.

Joe

----

BTW: Whenever you have a question (as opposed to an issue) it's better to post it on our [list](https://groups.google.com/forum/#!forum/autokey-users) or on [Gitter](https://gitter.im/autokey/autokey). The list is more user oriented whereas more technical questions will get a better answer on Gitter.

Both of these venues have a number of active users, so you are likely to get more, better, and faster answers on them than here on GitHub where only our very busy developers and a few users see the posts. We also prefer to have the developers developing rather than taking time to answer questions and manage issues that the user community could handle for them.

New users are encouraged to spend some time browsing through our [wiki](https://github.com/autokey/autokey/wiki). There are a number of helpful articles there as well as many example AutoKey scripts.

[...]

PR #896, opened 2023-08-06
Rename the bug.yaml file to bug.yml.

Elliria:
Change the extension in the [bug.yaml](https://github.com/autokey/autokey/blob/master/.github/ISSUE_TEMPLATE/bug.yaml) file's name from `.yaml` to `.yml`.


Elliria:
Since one of the things that was unusual about issue #893 was that a blank issue template was used, I realized that that can be prevented by the addition of a [config.yml](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser) file, so I was about to add one when I realized that our issue template had the wrong extension on its name. Apparently, that extension is accepted by GitHub, because it's been working, but it's better to give it the "politically-correct" extension, so it's now renamed.

PR #900, opened 2023-08-07
Remove unnecessary white-space from the config.yml file.

Elliria:
Remove the unnecessary white-space from the [config.yml](https://github.com/autokey/autokey/blob/master/.github/ISSUE_TEMPLATE/config.yml) file that had gotten there from copying and pasting.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,2 +1,1 @@
-	blank_issues_enabled: false 
- 
+blank_issues_enabled: false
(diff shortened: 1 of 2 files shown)


Elliria:
After this is merged, the [bug.yaml](https://github.com/autokey/autokey/blob/develop/.github/ISSUE_TEMPLATE/bug.yaml) file on the **develop** branch will also need to be renamed to `bug.yml`, its contents will then need to be updated to match those in the same file on the **master** branch (three new sections will be added - the Xorg question and two label sections), and the `config.yml` file will need to be copied over to the **develop** branch. Then, unless the files should also be put onto the **beta** branch, this will be done.

PR #899, opened 2023-08-06
Add the config.yml file to prevent blank issues.

Elliria:
Create the `config.yml` file in the [/.github/ISSUE_TEMPLATE](https://github.com/autokey/autokey/tree/master/.github/ISSUE_TEMPLATE) directory with a configuration statement that prevents blank issues from being offered and created.

PR #901, opened 2023-08-08
Rename the bug.yaml file to bug.yml on the develop branch.

Elliria:
* Rename the bug.yaml file to bug.yml.
* Update the [CHANGELOG.rst](https://github.com/autokey/autokey/blob/develop/CHANGELOG.rst) file:
  * Add the **Other changes** heading to the **Version Develop** section.
  * Note the renaming of the `bug.yaml` file to `bug.yml`.


josephj11:
Merged and saw that it worked as intended.

Elliria:
This PR didn't actually change anything, because the action uses the files on the **master** branch to do its thing. It just brought the name of this file (the one on the **develop** branch) into compliance. The fix to the issue template was done two days ago on the **master** branch in PR #899 and PR #900 that BlueDrink merged in.

This file (the one on **develop**) is about to get another edit to make its contents identical to those in its counterpart on the **master** branch. This branch will also get a copy of the new `config.yml` file that the **master** branch just got.

My reason for doing these changes is that the **develop** branch is considered to be more modern than the **master** branch, so if a file is missing or different or only present on the **develop** branch, it's likely to (and probably should) be assumed that the state of the file on the **develop** branch is the desired and new-and-improved state for that file and its state on the **master** branch can be discarded when it comes time for the next release. To that end, anything that we've added to or changed or taken away from **master** that's desired should also be done to **develop**.

Anyway, those two PRs will be next.

josephj11:
I thought it was something like that. It did prove that the renamed file still works.

Elliria:
Yeah. I actually tried it out in my dummy repository first and it worked perfectly there. Then, when I did it here on the **master** branch and it didn't work, I was worried for a moment that there must have been something I overlooked. Then I realized, to my chagrin, that I had somehow added white-space on both sides of the single command in the file I had created, breaking the file. Luckily, that was a quick fix to remove and it solved the problem. Live and learn.

josephj11:
No white-space privilege allowed here!

PR #940, opened 2024-04-11
Change erroneous close to resize_move

K0stov:
Erroneous description in documentation.

### DIFF
--- a/doc/scripting/lib.scripting.Window-class.html
+++ b/doc/scripting/lib.scripting.Window-class.html
@@ -489,7 +489,7 @@
   </tr></table>
   
   <p>Resize and/or move the specified window</p>
-  <p>Usage: <code>window.close(title, xOrigin=-1, yOrigin=-1, width=-1, 
+  <p>Usage: <code>window.resize_move(title, xOrigin=-1, yOrigin=-1, width=-1, 
   height=-1, matchClass=False)</code></p>
   <p>Leaving and of the position/dimension values as the default (-1) will 
   cause that value to be left unmodified.</p>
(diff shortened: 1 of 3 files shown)


josephj11:
Thanks so much for your contribution, @K0stov . Welcome to the AutoKey community!

I'm the de facto project manager (like Dilbert's Pointy Haired Boss :)  ). We have no currently active developers, so you're stuck with me. I'm a programmer, but I am not skilled in Python and am not an AutoKey developer. I do know some things about the source code, but not in depth. I know a very little bit about Git and GitHub. I can merge changes, but only simple ones I understand.

I am reviewing this with the help of a former developer. Here are some initial responses.

PRs need to be based on the develop branch, not on the master.

The doc/scripting directory is depreciated and is only around to support older releases. No changes to that  will be made.

The new documentation is at https://github.com/autokey/autokey.github.io and I believe it is entirely generated from changes in metadata comments in the code base. Any documentation errors need to be fixed directly in the code.

I just got this documentation and it's not added to our wiki yet.
https://pastebin.com/fb7mW7jk

I haven't encountered the api.txt file before. It's used by scriptpage.py. It looks like something to do with editing scripts in the AutoKey script panel - maybe related to autocompletion. IDK what it does or whether changing it will affect the code in other places.

The change to window.py looks good (for develop).
While you're in there, "and" should be "any" on line 134.

K0stov:
Glad to hear from you, @josephj11.

The code works as it should, only the documentation is miswritten. Looks like this was fixed in all but one place in the `develop` branch. I've made [a new pull request](https://github.com/autokey/autokey/pull/941), where I've also changed _and_ to _any_ and edited the changelog as needed.

Keep up the good work.

Issue #964, opened 2024-11-08
Bump Python versions in workflow files + replace asyncore


Elliria:
### AutoKey is a Xorg application and will not function in a Wayland session. Do you use Xorg (X11) or Wayland?

Xorg

### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

Enhancement

### Choose one or more terms that describe this issue:

- [ ] autokey triggers
- [ ] autokey-gtk
- [ ] autokey-qt
- [ ] beta
- [ ] bug
- [ ] critical
- [X] development
- [ ] documentation
- [ ] enhancement
- [ ] installation/configuration
- [ ] phrase expansion
- [ ] scripting
- [ ] technical debt
- [ ] user interface

### Other terms that describe this issue if not provided above:

update

### Which AutoKey GUI did you use?

None

### Can you briefly describe the issue?

Update the Python versions in the [build.yml](https://github.com/autokey/autokey/blob/master/.github/workflows/build.yml) file, and the [python-test.yml](https://github.com/autokey/autokey/blob/master/.github/workflows/python-test.yml) file.

### Can the issue be reproduced?

None

<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #964</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @Elliria</code> (replace <code>20</code> with the amount, and <code>@Elliria</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

Elliria:
[...]

PR #967, opened 2024-11-09
Bump Python versions in workflow files to satisfy issue #964.

Elliria:
* Bump Python version in the `build.yml` file.
* Bump Python versions in the `python-test.yml` file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.10.1]
+        python-version: [3.12.7]
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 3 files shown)


Elliria:
[Python 3.12.7 is supported by GitHub in ubuntu-latest](https://github.com/actions/runner-images#available-images), so we should be able to support it.

This pull request causes [this test-install error](https://github.com/autokey/autokey/actions/runs/11757735866/job/32755337270?pr=967) in Python 3.12:

Error

It's complaining about not finding the **asyncore** module.

A search of the AutoKey code doesn't turn up any reference to **asyncore**, so I'm assuming, from the rest of the error text, that the `pyinotify.py` dependency is causing it.

A search for **pyinotify** in the AutoKey code turns up several references, but none of them refer to specific versions of it, so I'm not sure what the next step would be to solve this.

Any help in sorting this out would be appreciated.

Elliria:
More details about asyncore are on [python.org](https://docs.python.org/3.11/library/asyncore.html).

Elliria:
It looks like this test-failure may be an upstream issue and I've reached out to the author of `pyinotify.py` for suggestions. Meanwhile, since support for **asyncore** was removed in Python **3.12**, I'm closing this pull request and will bump the Python version up to **3.11** instead until we can sort this out.

PR #968, opened 2024-11-09
Bump Python versions in workflow files to satisfy issue #964.

Elliria:
* Bump Python version in the `build.yml` file to **3.11**.
* Bump Python version range in the `python-test.yml` file to **3.9** through **3.11**.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.10.1]
+        python-version: [3.11.10]
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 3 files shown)


BlueDrink9:
Was there a reason for 3.9 as the lower bound btw? I understand 3.11 is the latest that is currently passing tests. Are pythons < 3.9 still supported

josephj11:
AFAIK, 3.7 is depreciated, but 3.8 should still be valid, but that's a guess. Is there a reason to jump past 3.8? Does it matter much if we do jump anyway? 

AFAIK, this version spec has no effect on end users unless they are cloning the repo on GitHub for private modifications. The population doing that at all is pretty small and those doing so on an old version of Python of likely to be vanishingly small.

BlueDrink9:
I might be misunderstanding you Joe, but I think this test is valid for more people than cloners. Ultimately the versions affect what the tests run against, and therefore validates who can install and run autokey successfully. I think we should find out what the newest available version is on the oldest still-supported Ubuntu LTS, and use that as our yard stick?

BlueDrink9:
Looks like [3.8 is the oldest version I would keep support for](https://packages.ubuntu.com/focal/python3)

PR #971, opened 2024-11-17
Add the "Update workflows" section to the CONTRIBUTORS.rst file on the develop branch

Elliria:


Elliria:
I'm assuming these errors are because of the Wayland code in this branch. Should I cancel this and wait to do it until the Wayland code is finalized?

josephj11:
I glanced at this. Several thoughts.
On one error I saw the trigger abbreviation uinput which expands to  Sam. ;)

Maybe it's a chicken and egg problem, but I see tests against Python 3.7. I guess they're not removed yet.

I thought tox tests weren't active on develop.

Despite what I thought and promoted, I think develop needs to go away. Maybe all further PRs should be against master. This is difficult because I think develop has all the Wayland code in it. I'm having trouble remembering what's where. I don't like having a main branch that is substantially broken.

Hopefully, Sam will surface sometime soon to clarify and resolve some of this.

Elliria:
Ah, you're right. It's probably because I didn't update the action and Python versions on the **develop** branch yet. It turns out that `tox` tests run on all the branches, so all branches need their versions updated. At least that's nice and consistent. But in the event that I shouldn't currently be committing to this branch, I'll wait until we hear from Sam first.

josephj11:
Hopefully, @sebastiansam55 will show up fairly soon, or we'll have to find a way to work around his changes until he does.

Elliria:
If not, no worries. I can cancel this and try creating a PR that updates the action and Python versions on **develop** instead.

josephj11:
You mean Master? This is develop.
 We need to evaluate what, if anything, should get updated on develop. It's just more stuff to triage when merginginto master.

Elliria:
No, I mean **develop**. Cancel this PR, update the workflow versions (as long as that won't interfere with Sam's code), and then do this PR.

We need to keep updating **develop**, at least during this release-cycle, since that's where most of the changes have occurred. Then, we'd want to think long and hard before considering doing away with **develop** There are currently 191 forks of AutoKey. That's a lot of boats to rock and shouldn't be done lightly. 

I'm in favor of cleaning up our mess of aheads and behinds and trying to work at least one more release-cycle with the current structure used properly (i.e.: merge PRs onto **develop** instead of **master** except for hotfixes and workflows). I'm guilty of contributing to the mess and would be willing to try to keep it clean the next time around.

josephj11:
I clearly don't understand this well enough.

Would all those forks be of develop and not of main? How would they be impacted?

It makes sense to continue with develop for the next (major) release. It would be good to figure  out a path forward so that this is not perpetuated if others agree that a reliance on a develop branch is causing more problems than it solves.

I would like to see 0.96.1 assembled and released and then the Wayland changes (if good enough)... as 0.97.0 since that's a major change and not a patch/bug fix release.

BlueDrink9:
Just chipping in where I see an issue, in the linting I see
> lib/autokey/uinput_interface.py:824:79: F821 undefined name 'keyName'
            logger.warning("Error sending modified key %r %r: %s", modifiers, keyName, str(e))


I would expect that to cause a crash if it gets hit. Not sure what touched that code, but if you can figure out the PR that did, request that they fix the missing variable

Elliria:
They're forks of the AutoKey project and they'd be impacted if they interacted with AutoKey in any way, like by synchronizing their branches or doing a PR.

While waiting to hear whether it's okay to do PRs on **develop**, I've since updated the steps that this current PR would have added to the `CONTRIBUTORS.rst` file, so this PR will need to be canceled anyway so that a new one can be done with the new information.

I also took another look at the script errors above today and updating the workflows should solve the `pytest` error. I can try that tomorrow if we haven't heard from @sebastiansam55 by then. Meanwhile, if I do that and all we're left with is that lint error, are you okay with ignoring the lint error and putting PRs through anyway, @josephj11, since it's part of the still-evolving Wayland code?

josephj11:
@Elliria I'm good with a merge with a lint-only error, but if we can find and fix the code that caused it, that would be much better.

It sounds like, once found, we just have to add a parameter to a function. That could be easy if the correct parameter value is readily discernible and available in the offending code. Since it might cause a crash, we should fix it if possible.

Elliria:
Okay, I'm going to close this out and attempt to sort out the PR test errors. Then this can be resubmitted with the current copy of it that I've got here. For now, I'm leaving the lint error alone, because that's part of @sebastiansam55's code and may work itself out.

PR #976, opened 2024-11-29
Satisfy Python 3.12.7 compatibility in issue #964 by adding the pyasyncore dependency.

Elliria:
Use code from the patch [provided by @dlk3](https://github.com/autokey/autokey/issues/946#issuecomment-2479559325) to replace the deprecated `asyncore` module with the `pyasyncore` module as a dependency in the `[testenv]` section of the `setup.cfg` file to resolve tox errors and make AutoKey compatible with Python 3.12.7.

PR #982, opened 2024-12-06
gha: Add python 3.12 and 3.13 to build and test matrix

highb:
WIP, running tests on a PR on my fork: https://github.com/highb/autokey/pull/1

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.10]
+        python-version:
+          - 3.10.15
+          - 3.11.11
+          - 3.12.8
+          - 3.13.1
     steps:
     - uses: actions/checkout@v3
       with:
@@ -64,7 +68,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.10]
+        python-version:
+          - 3.10.15
+          - 3.11.11
+          - 3.12.8
+          - 3.13.1
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
@@ -110,7 +118,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.10]
+        python-version:
+          - 3.10.15
+          - 3.11.11
+          - 3.12.8
+          - 3.13.1
     needs: [build_pypi_wheel, build_deb]
     steps:
     - uses: actions/checkout@v3
@@ -164,4 +176,3 @@ jobs:
     # ../autokey-common_${{ github.ref }}_all.deb
       env:
         GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,7 @@ jobs:
   build_deb:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version:
           - 3.10.15
@@ -67,6 +68,7 @@ jobs:
   build_pypi_wheel:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version:
           - 3.10.15
@@ -117,6 +119,7 @@ jobs:
   release:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version:
           - 3.10.15
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
     tags:
       - 'v*.*.*'
     # Always manually run on CI
-    branches: [ CI ]
+    branches: [ CI, update-gha-to-py-312-313]
 
 jobs:
 
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -21,7 +21,13 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.7", "3.10"]
+        python-version:
+          - "3.7"
+          - "3.9"
+          - "3.10"
+          - "3.11"
+          - "3.12"
+          - "3.13"
 
     steps:
       - uses: actions/checkout@v3
@@ -48,7 +54,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install flake8 wheel
           pip install -r pip-requirements.txt
-          
+
       - name: Python Code Quality and Lint
         uses: ricardochaves/python-lint@v1.4.0
         with:
@@ -83,7 +89,13 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.7", "3.10"]
+        python-version:
+          - "3.7"
+          - "3.9"
+          - "3.9"
+          - "3.11"
+          - "3.12"
+          - "3.13"
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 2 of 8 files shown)


highb:
arg, I need to rebase not use github's UI here.

josephj11:
Great! This is an area we really need help with. @Elliria has been updating our GHAs, but we're learning as we go.

Is it appropriate to support 3.13 yet? It's very new and and appears to have discontinued a few things we use, so that will require some (hopefully minor) programming changes. My only concern is that if 3.13  tests fail, they may prevent releases from building. That might be a small inconvenience for us, but it could cause distro packagers to reject new AutoKey releases.

Elliria:
I love your enthusiasm and your help is very much appreciated. I see that you're in the process of changing this PR. I had created a list of comments and questions based on the original form of this PR and am posting them below in case they're still of use. My apologies for the amount of comments and questions, but you made a lot of changes and I'm a junior coder with some familiarity with this project, but still have a lot to learn.

* About the change to the `.github/workflows/build.yml file`:
  * We don't put the version tag onto the **develop** branch because the workflow looks for that tag to create the builds and we don't want **develop** to be built. The **beta** and **master** branches are the two that get built when there's a new release. This is already written up, but isn't documented yet since develop is currently unstable while the Wayland work is underway.
  * We don't include branches from forks in the CI. What's the purpose of doing that since the tests will run here as part of submitting the PR?
  * What does `fail-fast: false` do?
  * Our current goal is to support Python **3.9.20** through **3.12.7**:
    * The [lowest version of Python supported by GitHub for Ubuntu 20.04](https://github.com/actions/runner-images#available-images) that [isn't EOL](https://devguide.python.org/versions/) is **3.9.20**.
    * The [highest version of Python supported by GitHub for Ubuntu 24.04](https://github.com/actions/runner-images#available-images) is currently **3.12.7**. Those changes have been cherry-picked onto **develop** and are awaiting merge.
  * Why do you do ranges of Python versions in this file and why do we do only our highest version? What's the difference in behavior?
  * We prefer to handle white-space changes in their own PR rather than with changes to code. Since this seems to be a recurring "theme", once we get **develop** stable, I might assign myself the task of climbing through every single file and doing a white-space clean-up on the whole project so those won't exist any more.

* About the changes to the `.github/workflows/python-test.yml` file:
  * Same comments and questions as above.

* About the changes to the `lib/autokey/autokey_app.py` file:
  * Same white-space comment as above.
  * What kind of test can we do inside of AutoKey to test out the regex change?

* About the change to the `lib/autokey/model/abstract_abbreviation.py` file:
  * What kind of test can we do inside of AutoKey to test out the regex change?

* About the change to the `lib/autokey/model/constants.py` file:
  * What kind of test can we do inside of AutoKey to test out the regex change?

* About the changes to the `lib/autokey/model/key.py` file:
  * Same white-space comment as above.
  * What kind of test can we do inside of AutoKey to test out the regex change?

* About the change to the `lib/autokey/scripting/window.py white-space ONLY change` file:
  * Same white-space comment as above.

* About the change to the `pip-requirements.txt` file:
  * I recently made that change in [PR #976](https://github.com/autokey/autokey/pull/976) and it was determined that that was not the correct location for that change, so the [PR was reverted](https://github.com/autokey/autokey/pull/979) and I created a [new PR](https://github.com/autokey/autokey/pull/981) that was just merged that added it to the `setup.py` file instead. Note that this has only yet been done on the **master** branch and will now need to be cherry-picked onto **develop** as well.

Elliria:
And just a heads-up on all those failed tests, GitHub doesn't support **Python 3.7** for Ubuntu any more and doesn't yet support Python **3.13**** for Ubuntu. We also don't yet have **develop** set up to support Python 3.12. Those three versions are expected to fail here as a result and don't need to be debugged. The others will need to be figured out.

Elliria:
@josephj11, talking about Python versions, I'm about to do a sloppy bump on **master**:

I missed two files when I bumped the Python versions on **master**, so I'm about to update those to **3.11.10** so they'll match the rest of the already-bumped files.

After that, since the **pyasyncore** issue is fixed now, I'll try bumping the highest version of Python in all of them to **3.12.7**.

The reason I want to do it in two steps instead of one is that if the bump to **3.12.7** falls over and has to be reverted, then all the files (including the two I missed earlier) will be at **3.11.10**.

If it works, all of that can get cherry-picked onto **develop** and hopefully turn red all the red x's for valid code into green check-marks from now on.

Elliria:
I poked around in the test errors:
* The **3.7** lint error is because GitHub no longer supports that version. This will go away once the bumped workflow files are cherry-picked from **master** to **develop**.

* The **3.9**, **3.10**, **3.11**, and **3.12** lint errors are all pointing to the `lib/autokey/uinput_interface.py` file, which is one of the new Wayland files that are still being worked on by @sebastiansam55, so those should be sorted out eventually.

* The **3.13** lint error is because GitHub doesn't support that version yet. This can be avoided by not using that version in the workflow files.

* The **3.12** test-install error is likely because we haven't bumped the workflow versions on **develop** yet.

* The **3.13** test-install error can be avoided by not using that version in workflow files.

dlk3:
@josephj11 asked:
> Is it appropriate to support 3.13 yet? It's very new and and appears to have discontinued a few things we use, so that will require some (hopefully minor) programming changes.

Fedora 41, the current release, delivers Python 3.13 as its default so if you want AutoKey to continue to run on Fedora ...

Fedora 41 does make versions 3.10 - 3.14 available as optional add-ons so a user could install 3.12 and still run AutoKey but given that the required change I saw on the "develop" branch is so trivial I can't imagine why you wouldn't just go ahead add support for 3.13.  The only change I had to make was to replace the ***imghdr*** function that is removed in 3.13.  ([PR #992](https://github.com/autokey/autokey/pull/992))

josephj11:
Great! If the upgrade to 3.13 is that trivial, that's really good news. Hopefully, @Elliria will time to look at this aspect shortly because she's been doing all our GH actions...

Elliria:
We officially support the versions of Python that GitHub supports, but we try to help out when folks try to run AutoKey in unsupported environments. When it comes to support, you have to draw your lines somewhere and the way we do it is by using the Python versions listed in the **Python** section inside of the **Cached Tools** section inside of the **Included Software** link for the **most recent** Ubuntu release (the top release listed) on the [GitHub Actions Runner Images](https://github.com/actions/runner-images#available-images) page.

The versions are currently:
* 3.9.20
* 3.10.15
* 3.11.11
* 3.12.8

highb:
Hey all, this PR got lost in the shuffle of the holidays. I'll try again soon with all of the updated context above, unless someone else has already done the work.

Elliria:
AutoKey supports the Python versions that GitHub supports for the latest version of Ubuntu as listed on its [GitHub Actions Runner Images](https://github.com/actions/runner-images#available-images) page. If you check that page, it looks like they just recently updated it and these Python versions are now supported by GitHub:
* 3.9.21
* 3.10.16
* 3.11.11
* 3.12.8
* 3.13.1

With the code-base as it currently is, the **develop** branch can't be updated. When [pull request #1004](https://github.com/autokey/autokey/pull/1004) goes through, it should solve this and a few other issues automatically and the **develop** branch should be fully functioning again going forward. At that point, I can update the **develop** branch with the new Python versions.

highb:
Hey @Elliria , I haven't had time to focus on this PR and I probably won't for a while. Just too much going on at work and at home right now. You may want someone else to take a look.

Elliria:
No worries, @highb. We won't be able to update the **develop** branch until after [pull request 1004](https://github.com/autokey/autokey/pull/1004) is merged in anyway. Once you've got some time, maybe you could [test that pull request](https://github.com/autokey/autokey/discussions/1025) and provide feedback. The more people we have testing, the better.

Issue #709, opened 2022-06-28
Changes to autokey.log


josephj11:
### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

Enhancement

### Which Linux distribution did you use?

N/A

### Can you briefly describe the issue?

The AutoKey log lives in the AutoKey configuration directory tree where it is primarily unnoticed.

The existence of this log needs to be justified. I have used AutoKey for decades and have never used this log. Why is it there? What purpose does it serve?

If I need a trace, I run one.

If there is no good answer to this, the log should be eliminated.

This log grows forever. If it is good for something, it needs to have logrotate applied to it so the current version is relatively small and fresh and older versions are periodically removed.

### What should have happened?

This log should be used or eliminated. If used, then its size should be limited via log rotation and something should be added to our wiki describing the log and how to use it.

### What actually happened?

The log is relatively invisible and unused and grows forever wasting megabytes of space.

### Anything else?

The log can be manually deleted or otherwise edited from outside of AutoKey with no consequences (at least while AutoKey isn't actively in the process of writing to it).

BlueDrink9:
I think definitely logrotate. Some bugs are hard to reproduce - having a log file always there is pretty standard software practice

josephj11:
How does the log compare to running a trace? I've never used it.

BlueDrink9:
It's similar content, but you don't have to have been running in terminal to access it - so if it's a once-off bug or a user doesn't know how to get a trace, we can just ask for the log

dlk3:
The $HOME/.local/share/autokey/autokey.log file is rotated by AutoKey internally, as @BlueDrink9 indicated.  The file is limited to 5MB in size and three copies are kept - one active file and the two previous ones.

I also agree with @BlueDrink9 that there's potential value in this file for troubleshooting a difficult-to-catch bug.  It provides the user with a way to capture logging output without always having to run AutoKey from a terminal prompt.  

I am submiting PR #999 to change the on-disk logging log-level settings so that it will respect the --verbose and --mouse options, which it doesn't do now.  My change also puts a date/time stamp and the system component name on each message in the on-disk log just like the console output has.  

[...]

PR #994, opened 2024-12-19
Bump versions in workflow and other files on develop.

Elliria:
* Copy bumped action and Python versions from workflow and other files on **master** to workflow files on **develop**.
* Copy `pages.yml` workflow file from **master** to **develop**.
* Copy **pyasyncore** fix from `setup.py` on **master** to `setup.py` on **develop** as part of this update (from the fix for [issue #946](https://github.com/autokey/autokey/issues/946) and [issue #964](https://github.com/autokey/autokey/issues/964) that had already been done on **master**).

PR #999, opened 2024-12-23
Make on-disk logfile content match console message output

dlk3:
Adjust the format of the messages recorded in the $HOME/.local/share/autokey/autokey.log disk file so that they match the format of the messages output to the console.  (Adds time/date stamp and affected module name to the messages in the on-disk log file.)

Make it so that the disk file honors the --verbose and --mouse debugging options and contains the exact same messages as these options output to the console.

May resolve issue #709

### DIFF
--- a/lib/autokey/logger.py
+++ b/lib/autokey/logger.py
@@ -44,6 +44,7 @@ def configure_root_logger(args: Namespace):
         backupCount=MAX_LOG_COUNT
     )
     file_handler.setLevel(logging.INFO)
+    file_handler.setFormatter(logging.Formatter(LOG_FORMAT))
 
     stdout_stream_handler = logging.StreamHandler(sys.stdout)
     logging_level = logging.INFO
@@ -51,7 +52,8 @@ def configure_root_logger(args: Namespace):
         logging_level = logging.DEBUG
     if args.mouse_logging:
         logging_level = 9
-    
+
+    file_handler.setLevel(logging_level)
     stdout_stream_handler.setLevel(logging_level)
     stdout_stream_handler.setFormatter(logging.Formatter(LOG_FORMAT))
 


josephj11:
I looked briefly at your code.

It looks like AutoKey supports logging levels. This is something I've been asking for for years and was pretty sure we didn't have. 

Can you glance at the code that supports that and let us know functionally what it does?

In particular, we get a lot f traces that start with a huge number of key  grabbing log messages that are almost always useless. I'm wondering if I open another issue to get those suppressed by defaullt (but still available if needed), would it be easy to do or if is this only a mouse or no mouse logging toggle.

dlk3:
> It looks like AutoKey supports logging levels. This is something I've been asking for for years and was pretty sure we didn't have.

The logging module in Python provides this function and that's what AutoKey is using.  Each individual message that is generated by the code is assigned a given logging level, and then each logging destination, like a disk file or the console, is configured with a log level threshold value so that only  messages that are at or above that threshold will be visible in that log.  There are predefined levels in Python but you can define custom levels as well.

The predefined levels are named constants with integer values:
```
logging.DEBUG = 10
logging.INFO = 20
logging.WARNING = 30
logging.ERROR = 40
logging.CRITICAL = 50
```
Custom messaging levels are any integer in between these.  For example, AutoKey's mouse messages are issued at level 9.  If AutoKey's --mouse option is specified, then the log level threshold for log output is set to 9 and all messages issued at level 9 and above are sent out to the console.   The --verbose option sets the threshold to logging.DEBUG (10), a threshold one step higher, so the mouse messages are then suppressed.  The default threshold when no options are used is logging.INFO (20).

In order to do what you're asking we would need to go though the code and give all the message types you want to filter out a unique level number within the hierarchy of message levels.  We'd then need to design a scheme for the command line options that control which message types are output to the log.  Some applications do this with a single ```--verbose=N``` option where N is a value representing the desired level of verbosity.  Others do it the way AutoKey does it now, where separate options like ```--verbose``` and ```--mouse``` incrementally increase the verbosity level.

So the net of all this is that I can probably do something like this for you but I'd need you to define, in order, what the increasing levels of verbosity you want are, i.e. modify and or extend the following list:

Starting at the lowest level of verbosity and going up:
- Always show INFO, WARNING, ERROR and CRITICAL level messages (current default when no options are used)
- Show DEBUG level messages (the current --verbose option)
- Show mouse level messages (the current --mouse option)
- Show the key grabbing messages

Let me know what you think / want to do.

Elliria:
Thank you for a wonderful explanation of all of this, @dlk3. The man pages ([GTK](https://github.com/autokey/autokey/blob/develop/doc/man/autokey-gtk.1) and [Qt](https://github.com/autokey/autokey/blob/develop/doc/man/autokey-qt.1)) on **develop** contain the `-m` or `--mouse` option.

By the way, I thought I'd jump in and say that this LGTM except that it needs to just be on **develop** instead of **master** unless it's currently needed as a hot-fix. If you'd like it as a hot-fix on **master**, @josephj11, we can put it through right away once it has the accompanying `CHANGELOG.rst` entry, since **master** passes all tests.

I see now that you already added this same fix to your Wayland PR, @dlk3. If it gets submitted again on **develop** in a separate PR instead, perhaps we could put through my [bump of the versions](https://github.com/autokey/autokey/pull/994) on **develop** to try for all green check-marks (fingers crossed) and then put it through on **develop**, too.

josephj11:
Is there any reason this should have high priority?" It's nice, but it doesn't fix anything that's seriously broken, so I would think it should go into develop and not main. IIUC, every PR we add to main creates technical debt/complications for when we want to release develop as a new version. I want that situation to get better,  not worse.

This is the sort of change that could be added in 0.96.1, if we choose to create that release.

As far as changing how AutoKey works with logging levels, I'll open a new enhancement issue. That will require an evaluation of the work involved and how we (not just me) want it to work. I don't understand why the current levels for mouse, etc. were set the way they are. I don't want to arbitrarily change something that might have been done for a good reason. @BlueDrink9 ,  any thoughts?

I have a whole list of my pet issues/enhancements that I would rather see done first,  but that's just my opinion and developers are free to work on whatever they choose.

All of this is somewhat academic until we have someone willing and able to take on the release process so we can release anything. I'm not wild about having to support a lot of people using an unstable develop branch in order to get new features, etc.

dlk3:
The effort is trivial.  I have gone ahead and prototyped the change in line with the list in my previous comment.  I set up a separate logging level for the "grabbing hotkey" messages and an associated command line option to show them in the log.  (I haven't pushed this commit yet, it's only on my workstation.)

Here's what the AutoKey help looks like now with the new "--debug" option I added.  For now the phrase "and additional debugging output"  on the "--debug" option only refers to the "grabbing hotkey" messages.  Those messages only show up when "--debug" is used, otherwise they are suppressed.  My intent with "--debug" was to create a "show everything" option so that, in the future, it would be possible to create even more custom message levels and still have an option that would show all messages at all levels.  That seemed better to me than creating an option specific to just the "grabbing hotkeys" messages.
```
usage: autokey [-h] [-c] [-C] [-v] [-m] [-d] [-V]

desktop automation

options:
  -h, --help         show this help message and exit
  -c, --configure    open the AutoKey main window
  -C, --cutelog      connect to a local default cutelog instance to display the full program log
  -v, --verbose, -l  enable verbose logging (-l is deprecated)
  -m, --mouse        enable verbose logging including mouse button events
  -d, --debug        enable verbose logging including mouse button events and additional debugging output
  -V, --version      print the current AutoKey version to standard output
```
The code I used for this also puts custom labels on the custom message types in the log.  Instead of all of these messages having the same "DEBUG" label, now the mouse messages will have a "MOUSE" label and the "grabbing hotkey" messages will have a "GRABKEY" label to separate them out.

Here are some example messages that show what that looks like:
```
2024-12-24 19:10:49,826 GRABKEY - autokey.interface - grabbing hotkey: ['<ctrl>'] '<f7>'
2024-12-24 19:10:53,179 MOUSE - autokey.service - Mouse click at root:(2623, 167) Relative:(832,48) Button: 1 In window: WindowInfo(wm_title='dlk@fang:~ — bash', wm_class='gnome-terminal-server.Gnome-terminal')
2024-12-24 19:11:49,107 DEBUG - autokey.service - Window visible title: 'Make on-disk logfile content match console message output by dlk3 · Pull Request #999 · autokey/autokey — Mozilla Firefox', Window class: 'Navigator.org.mozilla.firefox'

**Comment by dlk3** [comment]
Oh, and I wanted to be sure everyone was aware ... There are no "grabbing hotkey" messages when AutoKey is running with Wayland.  These messages only come when you're in X11.

**Comment by josephj11** [comment]
I like the latest changes. Can we select grabs and not mouse, etc. They're more on the same level than they are hierarchical, but the standard logging probably doesn't support that.

**Comment by dlk3** [comment]
> Can we select grabs and not mouse, etc. They're more on the same level than they are hierarchical, but the standard logging probably doesn't support that.

Yeah, I think we could do something like that.  I'll let you know what I come up with.

**Comment by Elliria** [comment]
I'll add the **-d/--debug** option to the man pages once we put [PR #994](https://github.com/autokey/autokey/pull/994) through successfully (passing all tests).

**Comment by dlk3** [comment]
@Elliria 
> I'll add the **-d/--debug** option to the man pages once we put [PR #994](https://github.com/autokey/autokey/pull/994) through successfully (passing all tests).

Please don't add this.  This change is not part of PR #994 and @josephj11 has suggested a totally different approach to this enhancement which will probably result in a different option being used not "--debug"

**Comment by dlk3** [comment]
@josephj11 said:
> Can we select grabs and not mouse, etc. They're more on the same level than they are hierarchical, but the standard logging probably doesn't support that.

Ok, I've coded something like that.  It's still a relatively trivial change .  As before, it's only on my workstation right now.

With this approach, the idea is that you would use the -v (--verbose) option to get basic debug messages and then **add** either or both of the -m (--mouse) or -g (--grabkey) options if you wanted those types of messages to be included.  Without --verbose being there, using either of the --mouse or --grabkey options would produce an error message telling you they only work when --verbose is also used.

All the messages would be output with the same DEBUG label.  As an example, here's what a mouse click message would look like, exactly as it does now except for the "DEBUG" designation instead of the current "Level 9":
```
2024-12-26 12:41:35,947 DEBUG - autokey.service - Mouse click at root:(711, 52) Relative:(738,44) Button: 1 In window: WindowInfo(wm_title='dlk@fang:~ — bash', wm_class='gnome-terminal-server.Gnome-terminal')
```
Here's what the new help looks like:
```
$ autokey -h
usage: autokey [-h] [-c] [-C] [-v] [-m] [-g] [-V]

desktop automation

options:
  -h, --help         show this help message and exit
  -c, --configure    open the AutoKey main window
  -C, --cutelog      connect to a local default cutelog instance to display the full program log
  -v, --verbose, -l  enable verbose logging (-l is deprecated)
  -m, --mouse        add mouse button event messages to the --verbose output, only works when the --verbose option is
                     also used
  -g, --grabkey      add X11 "grabbing hotkey" event messages to the --verbose output, only works when the --verbose
                     option is also used
  -V, --version      print the current AutoKey version to standard output
```
Here's the error you'd get if you forgot to include the --verbose option:
```
$ autokey --grabkey
usage: autokey [-h] [-c] [-C] [-v] [-m] [-g] [-V]
autokey: error: The -g or --grabkey option only works when it's used together with the --verbose option
```

josephj11:
I like the new options.

We prefer to have such options listed in the help and man pages, etc.  in alphabetical order though.

BlueDrink9:
> Is there any reason this should have high priority?" It's nice, but it doesn't fix anything that's seriously broken, so I would think it should go into develop and not main. IIUC, every PR we add to main creates technical debt/complications for when we want to release develop as a new version. I want that situation to get better, not worse.
> 
> This is the sort of change that could be added in 0.96.1, if we choose to create that release.
> 
> As far as changing how AutoKey works with logging levels, I'll open a new enhancement issue. That will require an evaluation of the work involved and how we (not just me) want it to work. I don't understand why the current levels for mouse, etc. were set the way they are. I don't want to arbitrarily change something that might have been done for a good reason. @BlueDrink9 , any thoughts?
> 
> I have a whole list of my pet issues/enhancements that I would rather see done first, but that's just my opinion and developers are free to work on whatever they choose.
> 
> All of this is somewhat academic until we have someone willing and able to take on the release process so we can release anything. I'm not wild about having to support a lot of people using an unstable develop branch in order to get new features, etc.

No need to worry about historical features. The different levels were just because they were particularly noisy outputs, so it made sense to have them hidden by default.

Elliria:
@dlk3
 
> Please don't add this. This change is not part of PR #994 and @josephj11 has suggested a totally different approach to this enhancement which will probably result in a different option being used not "--debug"

Not to worry. I know it's not part of [PR #994](https://github.com/autokey/autokey/pull/994) since that's my PR and I wouldn't add an unrelated change to it. That PR should, however, make it so that we can put other PRs through. If you two decide on a different option or options, I'll be happy to add them to the man pages.

PR #1009, opened 2025-01-13
Fix "Modifier key (like <ctrl>) not released on releasing hotkey"

### LINKED ISSUE #1007: Modifier key (like `<ctrl>`) not released on releasing hotkey

Xorg

- [X] This is not a question.

Bug

- [x] autokey triggers
- [ ] autokey-gtk
- [ ] autokey-qt
- [ ] beta
- [X] bug
- [ ] critical
- [ ] development
- [ ] documentation
- [ ] enhancement
- [ ] phrase expansion
- [ ] scripting
- [ ] technical debt
- [ ] user interface

_No response_

- Lubuntu LTS 24.04
- Kubuntu LTS 22.04

Qt

Issue occurs on every version I tried. E.g. [`master branch`](https://github.com/autokey/autokey/commit/c1aa63573a918bd6d5ceeada828f7f0f2f9a8483)

Issue occurs on every version, whether with installs from the distribution, nor with `git clone`.

Using a hotkey with modifiers (e.g. `<ctrl>+<1>`) to trigger a script, when the hotkey is pressed, and the modifier key is released _before_ the non-modifier key, then the modifier key remains stuck in the 'down' position.

Always

1. Create a new Script (you can leave the script empty)
2. Register a Hotkey, e.g. `<ctrl>+<1>` and press "Save" to activate it.
3. Use a keyboard event viewer like <https://w3c.github.io/uievents/tools/key-event-viewer.html>
4. Press `<ctrl>`
5. Press `1` (at this point, the empty script gets triggered)
6. Release `<ctrl>` (make sure that `1` is still pressed!)
7. Release `1` afterwards
8. The focused application will not register the KeyRelease event of the `<ctrl>` key and now has the `<ctrl>` key stuck in the 'down' position

The modifier (`<ctrl>` key) should be released

### What actually happened?

The modifier (`<ctrl>` key) was not released

### Do you have screenshots?

_No response_

_No response_

### Anything else?

The problem can be tracked down to `autokey/lib/autokey/interface.py`, to the calls to `window.grab_key` which intercepts the hotkey. While the hotkey is down, modifier events are intercepted as well. Thus, the KeyRelease event of the modifier is intercepted and not passed through to the application. A possible solution is to handle modifier `X.KeyRelease` events in `__flushEvents` and to send them down to the focused window.
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #1007</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @BenjaminRi</code> (replace <code>20</code> with the amount, and <code>@BenjaminRi</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

BenjaminRi:
Fixes #1007.

The fix seems quite straightforward, we simply take the event and forward it to the focused window. I run this version of AutoKey now on my machines. I am not sure if there is a better way. X11/Xlib is incredibly convoluted and poorly documented. The best explanation on some of the involved moving parts can be found [here](https://github.com/zhanghai/xkeymacs/issues/1). The fact that this person abandoned the project speaks volumes.

There is one subtle change in this PR: We remove a 'magic sleep': [`time.sleep(1)`](https://github.com/autokey/autokey/commit/94e77ca02ef9cc7bca751a9cec0626a6ec082cdb#diff-29d8e7c5f4c99407f3acec433b0a8ba79c1b7b38fce6f04addf31a1cde5f1939L985)

It is completely unclear what this sleep does. It was introduced [here](https://github.com/autokey/autokey/commit/986c3a23ddbac3a193bbc6340329015e537fcdfd) and then the duration was increased [here](https://github.com/autokey/autokey/commit/82ce5a6926a2bc7684077d0128eaca23c084b84d). The commit messages don't tell us anything about it. It probably serves a purpose, but I cannot figure it out for the life of me. This sleep introduces an artificial delay upon receiving events, which is very undesirable. I removed it to make my solution snappier. This removal probably broke something, but everything seems to be fine from what I've seen 🤷‍♂️

### DIFF
--- a/lib/autokey/interface.py
+++ b/lib/autokey/interface.py
@@ -982,20 +982,68 @@ class XInterfaceBase(threading.Thread):
         while True:
             try:
                 readable, w, e = select.select([self.localDisplay], [], [], 1)
-                time.sleep(1)
                 if self.localDisplay in readable:
                     createdWindows = []
                     destroyedWindows = []
                     
                     for x in range(self.localDisplay.pending_events()):
-                        event = self.localDisplay.next_event()
-                        if event.type == X.CreateNotify:
-                            createdWindows.append(event.window)
-                        if event.type == X.DestroyNotify:
-                            destroyedWindows.append(event.window)
-                        if event.type == X.MappingNotify:
+                        evt = self.localDisplay.next_event()
+                        if evt.type == X.CreateNotify:
+                            createdWindows.append(evt.window)
+                        if evt.type == X.DestroyNotify:
+                            destroyedWindows.append(evt.window)
+                        if evt.type == X.MappingNotify:
                             logger.debug("X Mapping Event Detected")
                             self.on_keys_changed()
+                        if evt.type == X.KeyPress or evt.type == X.KeyRelease:
+                            keyCode = evt.detail
+                            rawKey = self.lookup_string(keyCode, False, False, False)
+
+                            logLevel = logging.DEBUG
+                            if logger.isEnabledFor(logLevel):
+                                # Only do all this extra work when we actually need it
+                                action = evt.__class__.__name__
+                                keySym = self.localDisplay.keycode_to_keysym(keyCode, 0)
+                                shifted = bool(evt.state & self.modMasks[Key.SHIFT]) ^ bool(evt.state & self.modMasks[Key.CAPSLOCK])
+                                numlock = bool(evt.state & self.modMasks[Key.NUMLOCK])
+                                altGrid = bool(evt.state & self.modMasks[Key.ALT_GR])
+                                key = self.lookup_string(keyCode, shifted, numlock, altGrid)
+                                modifiers = [k.value for k, v in self.modMasks.items() if evt.state & v]
+                                logger.log(logLevel, "Event type: {}, keyCode: {}, keySym: {}, key: {}, rawKey: {}, modifiers: {}".format(action, keyCode, keySym, key, rawKey, modifiers))
+
+                            if evt.type == X.KeyRelease and rawKey in HELD_MODIFIERS:
+                                # If we let go of the modifier key while the hotkey is pressed,
+                                # the KeyRelease event for the modifier ends up here and is lost
+                                # to the application. This results in stuck modifier keys.
+                                # We rectify this problem by sending the KeyRelease event to the focused window.
+
+                                logger.debug("Pass modifier key {} release event through to focused window".format(rawKey))
+
+                                focus = self.localDisplay.get_input_focus().focus
+
+                                new_event = event.KeyRelease(
+                                    detail=evt.detail,
+                                    time=evt.time,
+                                    root=evt.root,
+                                    window=focus, # Note: evt.window does not work here because X redirected the event to the window passed to the grab_key call
+                                    child=evt.child,
+                                    root_x=evt.root_x,
+                                    root_y=evt.root_y,
+                                    event_x=evt.event_x,
+                                    event_y=evt.event_y,
+                                    state=evt.state,
+                                    same_screen=evt.same_screen
+                                )
+
+                                self.localDisplay.send_event(
+                                    destination=focus, # Note: evt.window does not work here (see above)
+                                    propagate=True,
+                                    event_mask=X.KeyReleaseMask,
+                                    event=new_event
+                                )
+
+                                self.localDisplay.flush()
+
                             
                     for window in createdWindows:
                         if window not in destroyedWindows:
@@ -1346,7 +1394,7 @@ class AtSpiInterface(XInterfaceBase):
         return True
 
 
-from autokey.model.key import Key, MODIFIERS
+from autokey.model.key import Key, HELD_MODIFIERS, MODIFIERS
 import autokey.configmanager.configmanager as cm
 
 XK.load_keysym_group('xkb')


josephj11:
Thanks for this contribution. 

@BlueDrink9 I tagged you mainly to see if you have any input on the delay that was removed. The rest seems straight forward for our other devs to review.

BenjaminRi:
I just realized that perhaps the `send_event` and `flush` calls I added should not be executed on the event handler thread, but on the dedicated thread that calls Xlib functions. As far as I can tell, Xlib is not threadsafe.

Elliria:
I haven't had a chance to review this yet, but it looks like it will be over my head anyway, so someone else will need to look it over. I thought I'd provide a couple of comments, though:
* For this pull request to go through, it would need to be against the **develop** branch rather than the **master** branch.
* I'd prefer to see **event** rather than **evt**, since **event** is self-explanatory and the less cryptic code we have, the better.

Elliria (changes requested):
I haven't had a chance to review this yet, but it looks like it will be over my head anyway, so someone else will need to look it over. I thought I'd provide a couple of comments, though:
* For this pull request to go through, it would need to be against the **develop** branch rather than the **master** branch.
* I'd prefer to see **event** rather than **evt**, since **event** is self-explanatory and the less cryptic code we have, the better.

BenjaminRi:
> For this pull request to go through, it would need to be against the develop branch rather than the master branch.

I see, that can easily be rebased.

> I'd prefer to see **event** rather than **evt**, since **event** is self-explanatory and the less cryptic code we have, the better.

I agree. However, the reason why I changed it to `evt` was because `event` collided with `Xlib.protocol.event`, which is imported in the beginning of `interface.py`. So this is actually not an aesthetic choice, but one that was forced on me.

Yeah, I guess this PR needs some rework. Honestly, if nobody else has the issue I'm having, maybe it's not worth upstreaming. I consider the issue quite severe, but apparently it doesn't bother anyone else? I just don't wanna be that guy who is like "ok I solved it, bye" never to be heard from again.

Elliria:
Ah, then **evt** is fine.

This is definitely an issue worth correcting. I haven't had a chance to test it yet to see if I can repeat it at my end, but I know we have trouble with some keys not behaving as expected and this may well be the cause.

And there's no need for bye. We'd love it if you stuck around. We can use all the help we can get.

Elliria:
I tried out your code, but might be doing the test wrong, so you might want to look over what I did and let me know what to change.

## Test environment:
* **Operating system**: Kubuntu 22.04 LTS running inside of VirtualBox
* **Display server technology**: X11

## The test
1. Clone AutoKey and run it:
```bash
cd Desktop
git clone https://github.com/autokey/autokey.git
cd autokey
git fetch origin pull/1009/head:pull_1009
git checkout pull_1009
python3 -m venv venv
source venv/bin/activate
pip install -r pip-requirements.txt
pip install packaging
cd lib
python3 -m autokey.gtkui -cl
#...or...
python3 -m autokey.qtui -cl
```
2. Set `<ctrl>+1` as the hotkey for the first example phrase.
3. Save the change.
4. Open the **Kate** text editor.
5. Press the `<ctrl>+1` hotkey combination.
6. Observe **Kate** for changes.

The result was that nothing was pasted into **Kate**.

Here's the relevant part of the log that looks to me like AutoKey is doing what it's supposed to:
```bash
2025-01-18 06:56:39,687 DEBUG - autokey.iomediator.iomediator - Key.CONTROL pressed
2025-01-18 06:56:40,364 DEBUG - autokey.service - Raw key: '1', modifiers: [<Key.CONTROL: '<ctrl>'>], Key: 1
2025-01-18 06:56:40,365 DEBUG - autokey.service - Window visible title: 'Untitled  — Kate', Window class: 'kate.kate'
2025-01-18 06:56:40,576 DEBUG - autokey.iomediator.iomediator - Key.CONTROL released
```
I tested it with both the GTK and Qt toolkits and with and without the **venv** with the same result each time.

BenjaminRi:
> The result was that nothing was pasted into **Kate**.

Huh, I can't quite follow. Is that a problem? AutoKey without my changes also pastes nothing into Kate. So I didn't change anything regarding that.

The original bug I fixed is that a CTRL keyup event is caught within AutoKey and never reaches the underlying application (like Kate). For some programs, this leads to stuck modifiers which can be very annoying. The missing keyup event can be seen with a tool like <https://w3c.github.io/uievents/tools/key-event-viewer.html>. I recommend reproducing the issue in mainline AutoKey without my changes, then applying my changes and see how the behavior differs.

> 2025-01-18 06:56:40,576 DEBUG - autokey.iomediator.iomediator - Key.CONTROL released

AutoKey sees the the `KeyRelease` event of `<ctrl>`. The problem is that the underlying application doesn't see it, because it's been intercepted by `grab_key` and redirected to AutoKey.

mjeveritt:
I wondered where the 'bug' lied with this - as I use [Barrier](https://github.com/debauchee/barrier) to multiplex multiple systems together with the same keyboard/mouse shared by one system. When I use some Ctrl-Shift-X or Ctrl-Alt-Z shortcuts, they get 'stuck', and I have all sorts of chaos on the 'remote' system when the modifier is stuck exactly as described.

I look forward to the fix being incorporated, but happy to do some testing here also.. I think I can create myself a custom Debian package/source code install still .. (mostly-gone are my Gentoo compiling days!).

josephj11:
@mjeveritt This PR affects code in the core of AutoKey, so you should be able to test it in  any environment where AutoKey will run. Shouldn't be necessary to take any heroic measures...

BenjaminRi:
Here's what's gonna have to happen before we can move forward with this PR / bugfix:
- There needs to be at least one project maintainer who can understand and reproduce this bug and can verify that the bug is fixed by the patch. In my personal experience, this bug is extremely severe (it makes AutoKey almost unusable for me), so I am surprised that this bug wasn't already reported before. At this point, I am still not sure if anyone else has ever reproduced the problem I am describing. I have done everything I can to describe it all.
- If available, there needs to be further information about the thing I mentioned above, the removal of this `sleep(1)` call. Someone wrote this line of code, I would love to know why. Did its removal break something? Is there a test suite we can run? Do we accept the risk for potential breakage (who makes that decision?)
- If possible, some feedback and / or expertise on X11 (with respect to `grab_key`, forwarding the `KeyRelease` event and also the threading model) would be appreciated. I worked out some of those details and I can refine it, but I am not an expert in X11.
- If all of the above is done, I would be willing to do a rebase to the `develop` branch and refine the code further, if required.

mjeveritt:
I'll work on doing some methodical debugging time over the next week/by end of the month, as I'm pretty sure I have a reproducer (or a similar bug) but I cannot say Yet whether this patch (would) fix[es] it .. watch this space!™

Elliria:
My test results up there were invalid because they weren't done correctly.

I've redone the tests and have the results below. It's pretty big with all the screenshots, so I'm tucking it inside of an HTML **details** block below that you can click to open:

<details>
<summary>SEE THE TEST RESULTS</summary>

## Test results

### Test environment:
* **Operating system**: Kubuntu 22.04 LTS running inside of VirtualBox
* **Display server technology**: X11

### What was tested
The hotkey was used **correctly** and **incorrectly** on the the [W3C Keyboard Event Viewer](https://w3c.github.io/uievents/tools/key-event-viewer.html) page in each of these ways:
* Without AutoKey.
* In AutoKey 95.10 with both toolkits.
* In AutoKey from a clone of this pull request with both toolkits.

### The tests

* **Test the correct hotkey sequence**:
  1. Open AutoKey.
  2. Set `Ctrl+1` as the hotkey for an AutoKey phrase.
  3. Open the [W3C Keyboard Event Viewer](https://w3c.github.io/uievents/tools/key-event-viewer.html) page in your browser.
  4. Left-click the input field in the page.
  5. Use the hotkey **correctly**:
     1. Press **Ctrl**.
     2. Press **1**.
     3. Release **1**,
     4. Release **Ctrl**.

* **Test the incorrect hotkey sequence**:
  1. Open AutoKey.
  2. Set `Ctrl+1` as the hotkey for a phrase.
  3. Open the [W3C Keyboard Event Viewer](https://w3c.github.io/uievents/tools/key-event-viewer.html) page in your browser.
  4. Left-click the input field in the page.
  5. Use the hotkey **incorrectly**:
     1. Press **Ctrl**.
     2. Press **1**.
     3. Release **Ctrl**.
     4. Release **1**.

### The results

* **Without AutoKey:**
  * Correct hotkey sequence:
    * As expected, nothing was entered into the input field.
    * The key-presses and key-releases registered correctly.
without_autokey_correct_sequence
  * Incorrect hotkey sequence:
    * As expected, 11111111111 was entered into the input field.
    * The key-presses and key-releases registered correctly.
without_autokey_incorrect_sequence

* **AutoKey GTK 95.10**:
  * Correct hotkey sequence:
    * As expected, the phrase was entered into the input field in the browser.
    * The **Ctrl** key registered as being pressed **and** released **twice**.
autokey_gtk_95 10_correct_sequence
  * Incorrect hotkey sequence:
    * As expected, **foo111111** was entered into the input field.
    * The **Ctrl** key registered as being pressed **and** released **twice**.
autokey_gtk_95 10_incorrect_sequence

* **AutoKey Qt 95.10**:
  * Correct hotkey sequence:
    * As expected, the phrase was entered into the input field in the browser.
    * The **Ctrl** key registered as being pressed **and** released **twice**.
autokey_qt_95 10_correct_sequence
  * Incorrect hotkey sequence:
    * As expected, **foo111** was entered into the input field.
    * The **Ctrl** key registered as being pressed **and** released **twice**.
autokey_qt_95 10_incorrect_sequence

* **AutoKey GTK this pull request**:
  * Correct hotkey sequence:
    * As expected, the phrase was entered into the input field in the browser.
    * The **Ctrl** key registered as being pressed **and** released **three times**.
autokey_gtk_pr_correct_sequence
  * Incorrect hotkey sequence:
    * As expected, **foo1111** was entered into the input field.
    * The **Ctrl** key registered as being pressed **and** released **three times**.
autokey_gtk_pr_incorrect_sequence

* **AutoKey Qt this pull request**:
  * Correct hotkey sequence:
    * As expected, the phrase was entered into the input field in the browser.
    * The **Ctrl** key registered as being pressed **and** released **three times**.
autokey_qt_pr_correct_sequence
  * Incorrect hotkey sequence:
    * As expected, **foo111** was entered into the input field.
    * The **Ctrl** key registered as being pressed **and** released **three times**.
autokey_qt_pr_incorrect_sequence

### Summary
Although the key-presses that are being registered are odd, I don't see any issue in AutoKey's default behavior or on AutoKey's effect on the target window. As you can see, AutoKey did exactly what was asked of it each time and, in all cases, the phrase was pasted as expected.

What I found odd was that although in every test, I only pressed and released each key once, the presses and releases registered as happening more than once when AutoKey was involved. This could have something to do with key-repeat being on in the operating system (some operating systems don't have that on and it's something I prefer and always enable), causing it to register held keys as repeated presses and releases.

Also, in looking at those images some more, it looks like the press of the **1** key is what's not being registered when the hotkey sequence is entered correctly, although its release is registered. Interesting.

The extra presses of the **1** key that are entered into the input field in the browser when you do the hotkey out of sequence are expected since you're still pressing the **1** key after releasing the **Ctrl** key. If you press the **1** key without holding down the **Ctrl** key, you would want and expect to get a **1**. If a press of the **1** key after releasing the **Ctrl** key didn't result in **1** being entered into the active window, that would be a bug.

As a result, releasing the **Ctrl** key before releasing the **1** key is user-error if the **1** (or repeats of the **1**) aren't desired, isn't it?
</details>

mjeveritt:
> My test results up there were invalid because they weren't done correctly.
> 
> I've redone the tests and have the results below. It's pretty big with all the screenshots, so I'm tucking it inside of an HTML **details** block below that you can click to open:
> SEE THE TEST RESULTS

The problem as I understand it, and I believe the OP is experiencing .. is there is a set [or sets of] of circumstances, where despite the usual sequence of key presses/releases (as you detail nicely!) .. the release of the Modifier key is not getting Registered by Something. What that Something is, and what interferes, is still to be defined accurately, but in my case, it is also exacerbated by using a remote system, and the 'forwarding' of those keypresses across the remote 'bridge'.

Thank you for detailing your testing strategy, as it is something I will endeavour to replicate locally also.

Elliria:
Yeah, I wasn't able to replicate it at my end. As you saw, the only odd thing I found was that the registered key-presses and key-releases didn't accurately reflect what I was actually doing, but AutoKey did what I expected it to do as far as pasting the phrase.

I can see where using a remote system across a bridge could throw at least one additional monkey-wrench into the works.

Good luck with your tests. It will be interesting to compare them.

BenjaminRi:
@Elliria Very thorough work, thank you. It is odd that you get such different results from mine. I wonder if there's something funky going on with virtualization and VirtualBox. It might change the exact keyboard behavior when compared to running it natively. Other than that, I cannot really explain it. Is your host operating system Windows?

> As a result, releasing the Ctrl key before releasing the 1 key is user-error if the 1 (or repeats of the 1) aren't desired, isn't it?

Yes, if your finger remains on `<1>` for longer and you just start typing a row of ones, that is a user error.  Excess hotkey presses after releasing the modifier are not within the scope of this bug. It's really only about making sure that the application understands that the modifier was released.

Elliria:
@mjeveritt, I've been thinking about this:

> there is a set [or sets of] of circumstances, where despite the usual sequence of key presses/releases (as you detail nicely!) .. the release of the Modifier key is not getting Registered by Something. What that Something is, and what interferes, is still to be defined accurately,

That's probably what the delay mentioned above is for and it likely works in most cases, while you two are experiencing an edge-case that needs a bit more oomph. AutoKey does its thing at machine-speed while some programs expect and can only handle human-speed input.

To get around that, you can replicate human-speed by introducing a delay, making it several seconds long at first, to verify that it will work, and then whittling it down to its shortest possible length afterwards. And if you wanted to, you could go a step further by using a conditional block to check the active window's type and only send the delay to windows that actually need it.

Elliria:
@BenjaminRi,  
> It is odd that you get such different results from mine. I wonder if there's something funky going on with virtualization and VirtualBox.

I wouldn't think so since it's just emulating hardware, but out of curiosity, I did the test of pressing the hotkey in the correct sequence with **AutoKey GTK 95.10** on bare metal (on my production machine) as a comparison:
1. Press and hold the **Ctrl** key.
2. Press the **1** key.
3. Release the **1** key.
4. Release the **Ctrl** key.

<details>
<summary>SEE THE RESULTS</summary>

## Test the correct-sequence with **AutoKey GTK 95.10** in a VM:
VirtualMachine

## Test the correct-sequence with **AutoKey GTK 95.10** on bare metal:
BareMetal
</details>

As you can see, just as it did in the VM, it registers an extra press and release of a key that I didn't do. On bare metal, the extra press and release of the key is registered as **Unidentified**. In the VM, the extra press and release of the key is registered as the **Ctrl** key. In both cases, AutoKey performs as requested by pasting the phrase into the text-box. Note that I also did the test with the incorrect sequence on bare metal with the additional presses and releases of the **Ctrl** key being registered as **Unidentified** in the same way, but I didn't include a screenshot of that here.

> Is your host operating system Windows?

No, both the host machine and the VM use **Kubuntu 22.04 LTS** amd **X11**.

BenjaminRi:
@Elliria I found the reason for the discrepancy between our observations. In my bug repro, I wrote:

> 1. Create a new Script (you can leave the script empty)

However, you tested it with the first phrase:

> 2. Set <ctrl>+1 as the hotkey for the first example phrase.

The assumption that hotkeys for phrases and scripts work the same way is very reasonable, but it seems to be wrong. Earlier, I saw that you were using phrases but did not realize it could be the cause of the different behavior. **The described bug only affects scripts.**

In detail: The first example phrase does not work in tests because it has a window filter for `.*-gedit`, so it doesn't work on any window other than gedit, so the shortcut doesn't apply (this is also why it didn't paste anything into Kate). Even if you do use a phrase without window filter (e.g. the second example phrase), the behavior is different from scripts. This is actually interesting, because I did not expect the phrases to use an entirely different logic from the scripts (under the hood, they could use the exact same mechanism, but they don't). With phrases, I see the behavior you reported.

Of course, on an architectural level, this entire story also raises the questions:

- Why don't phrases and scripts work the same way under the hood?
- Should they be made to work the same way?

That discussion is way outside the scope of this bug report, but knowing the answers would potentially enable a better fix.

Elliria:
My mistake. I repeated all of the tests with a script. This time, I just put a quick bit of commentary at the top and have left the screenshots to do the rest of the talking without additional commentary.

Before I begin, though, it's important to note that since the pull request is done off of the **master** branch, it's an unfair test since **master** is far behind **develop**, and it's quite possible that that has an effect on the results.

<details>
<summary>CLICK TO SEE THE TESTS</summary>

## About the tests
* All tests were done in the same VM with the same operating system and display server technology as before.
* Each test was done with an "empty" script that just contained the default comment that's in any script since empty scripts are not allowed.
* Each test was **also** done with a script that contained this line, which pastes **foo** into the active window:
  ```python
  keyboard.send_keys("foo")
  ```
* AutoKey behaved perfectly in every case, giving me exactly what I asked for.
* The keypresses, however, regsitered in wildly different ways in each test.

## AutoKey 95.10 GTK correct hotkey with script:

95 10 GTK correct hotkey with script

## AutoKey 95.10 GTK incorrect hotkey with script:

95 10 GTK incorrect hotkey with script

## AutoKey 95.10 GTK correct hotkey without script:

95 10 GTK correct hotkey without script

## AutoKey 95.10 GTK incorrect hotkey without script:

95 10 GTK incorrect hotkey without script

## AutoKey 95.10 Qt correct hotkey with script:

95 10 Qt correct hotkey with script

## AutoKey 95.10 Qt incorrect hotkey with script:

95 10 Qt incorrect hotkey with script

## AutoKey 95.10 Qt correct hotkey without script:

95 10 Qt correct hotkey without script

## AutoKey 95.10 Qt incorrect hotkey without script:

95 10 Qt incorrect hotkey without script

## AutoKey PR GTK correct hotkey with script:

PR GTK correct hotkey with script

## AutoKey PR GTK incorrect hotkey with script:

PR GTK incorrect hotkey with script

## AutoKey PR GTK correct hotkey without script:

PR GTK correct hotkey without script

## AutoKey PR GTK incorrect hotkey without script:

PR GTK incorrect hotkey without script

## AutoKey PR Qt correct hotkey with script:

PR Qt correct hotkey with script

## AutoKey PR Qt incorrect hotkey with script:

PR Qt incorrect hotkey with script

## AutoKey PR Qt correct hotkey without script:

PR Qt correct hotkey without script

## AutoKey PR Qt incorrect hotkey without script:

PR Qt incorrect hotkey without script

</details>

BenjaminRi:
For comparison, here's what it looks like on my end.

Autokey `master` branch:

ctrl down -> 1 down -> 1 up -> ctrl up:

Screenshot_20250124_210939

ctrl down -> 1 down -> ctrl up -> 1 up:

Screenshot_20250124_211005

(Note the missing Control keyup event)

---

Auto key with this PR:

ctrl down -> 1 down -> 1 up -> ctrl up:

Screenshot_20250124_211240

ctrl down -> 1 down -> ctrl up -> 1 up:

Screenshot_20250124_211257

The crux is the last image, where we (correctly) forward that Control keyup event.

Anyway, you are free to use my PR or reject it, it is entirely up to the project maintainers. The time I've spent on this issue has already exceeded the alotted time, so I will excuse myself from this issue and PR.

Issue #686, opened 2022-04-04
Add AutoKey version and selected front end to the end of AutoKey tracing output


josephj11:
### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

Enhancement

### Which Linux distribution did you use?

N/A

### Which AutoKey GUI did you use?

Both

### Which AutoKey version did you use?

0.96.0 Beta 10

### How did you install AutoKey?

N/A

### Can you briefly describe the issue?

When problems are reported, there is often some question as to what version was being used and which front end. It would be useful for tracing output to provide this information.

It is already listed at the very beginning but not at the very end of the tracing output, It would help to also have it at the end (When AutoKey is terminated, so if partial traces are posted it will still be visible.

josephj11:
After I filed this, it was pointed out to me that the version is already displayed at the beginning the log.

```
2025-02-21 04:22:15,036 INFO - autokey.configmanager.version_upgrading - Checking if upgrade is needed from version 0.96.0b10
```

However, also having it at the end of the log (when the program exits) would be great because often, only the end of the tracer is added to bug reports.

I just edited the Issue to reflect this. Sorry I didn't do it sooner.

PR #1018, opened 2025-02-20
Print application version in log

Andrei9699:
Pr for https://github.com/autokey/autokey/issues/686.
Adding version information, application type is already printed in the log on develop branch.


josephj11 (changes requested):
Thanks for the PR!

I  got the original request wrong. It's already at the beginning of the trace. I just modified the request to reflect that it's stiill needed at the end.

Sorry I didn't update the request sooner.

The desired change should be very similar, but in a different spot in the code.

PR #1019, opened 2025-02-21
Improved README with more detailed sections and added Features section

SamiSepehri:
**Key Updates:**
Improved Introduction: Clearly explains what AutoKey does and why it's useful.
Added Feature List: Summarizes key capabilities such as text expansion, scripting, and automation.
Expanded Installation Instructions: Provides direct package installation commands for Debian, Arch, and Fedora-based distributions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,13 +24,48 @@ About
 
 **Important**: This is an X11 application and, as such, will not function correctly when Wayland is in use instead of Xorg.
 
+**What is AutoKey?**
+AutoKey is a powerful Linux desktop automation tool for `X11`` that allows users to:
+- Automate repetitive text entries using abbreviations.
+- Trigger scripts and macros with hotkeys.
+- Control application windows via Python scripting.
+- Streamline workflows across multiple programs.
+
+With AutoKey, you can save time and reduce repetitive strain by automating text, keyboard, and mouse inputs across any application.
+
 .. _AutoKey: https://github.com/autokey/autokey
 .. _Google: https://code.google.com/archive/p/autokey/
+
+Features
+============
+✔️ **Text Expansion:** Use abbreviations to insert frequently used text or code snippets.  
+✔️ **Keyboard Shortcuts (Hotkeys):** Automate actions with custom key combinations.  
+✔️ **Python Scripting:** Create powerful automation scripts using Python.  
+✔️ **Application Control:** Automate interactions with running applications.  
+✔️ **Dynamic Macros:** Insert the current date, time, or custom tags dynamically.  
+✔️ **Cross-Application Support:** Works across all applications in X11 environments. 
 
 Installation
 ============
 
 **Important**: Please remove previous installations of AutoKey fully before installing!
+
+**Install via Package Manager**
+For Debian-based systems (Ubuntu, Debian, etc.):
+`sh`
+`sudo apt update && sudo apt install autokey-gtk`
+
+For Arch-based systems (Arch, Manjaro):
+`sh`
+`Copy`
+`Edit`
+`sudo pacman -S autokey-gtk`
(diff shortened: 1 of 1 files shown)


josephj11:
@SamiSepehri Welcome to the AutoKey community!

Updating documentation is hard work and is often unappreciated, but it's the difference between mediocrity and excellence. Thanks.

Specific comments:

I cheated and read the version in your branch rather than the actual diff, so some of these suggestions may refer to things that were there when you got here.

P3: contrary to the rest of the world, in AutoKey, there are phrases and scripts. The term nmacro is reserved for for the special purpose tokens  that can be embedded in phrases (all the things like `<left>`, `<cursor>`, and `<script>`).

@Elliria will undoubtedly chime in (she and I do most of the documentation work), but usually PRs should be based on the develop branch.

P4: should say most applications, not all. Some applications (notably some with names that start with an x like xterm and xsane) ignore synthetic keypresses and so ignore most of what AutoKey does. Same for the Features paragraph.

Installation: doesn't mention autokey-qt.

Elliria:
I love the idea of fleshing out the **About** section and adding a **Features** section. You've made it much more clear what AutoKey is for.

Although I agree with @josephj11 that PRs should usually be based on the **develop** branch, we break the rules in the case of the `README.rst` file since despite the fact that a developer would know to check other branches, it's the first document any user sees if they scroll down after loading the [AutoKey main page](https://github.com/autokey/autokey), so we made the changes on all the branches.

Before I chime in with my suggestions for this PR, I thought I'd mention that RST is a bit of a beast to work with, so there are a couple of sections of the RST syntax you'll want to update in your copy of this PR.

RST requires a blank line before a bulleted list and only accepts either an asterisk or a dash for the bullets. Both of your bulleted lists will want a blank line before them and you'll want to choose a bullet-type for the **Features** list.

RST code-blocks are also rather awkward because they require an introductory line of code followed by a blank line followed by a specific amount of indentation depending on where they're inserted. Also, the indentation can be a bit finicky and sometimes needs to be adjusted if they either don't display at all or you get display errors when the document is rendered. I've sometimes had to add what I thought was too many spaces when indenting code-blocks to get them to display.

At any rate, those are done like this for a plain code-block:
```rst

.. code-block::

   foo
   bar
   baz
```
And like this for a Bash code-block:
```rst

.. code-block:: bash

   foo
   bar
   baz
```
And so on.

I hope you don't mind if I had a bit of a play with your file and came up with my own version of it, which includes the suggestions made by @josephj11. I originally started by listing off some suggestions and then decided to just edit the document and attach it here along with the original from the repository and along with a copy of yours, so you can do a comparison of all three in your favorite program (I can highly recommend **Meld** for that) and see if you'd like to use any of my changes.

I've also included the additions you made to the **Installation** section in my copy, but I'm thinking that those don't really belong in this document. Since there's already a link to the [Installing](https://github.com/autokey/autokey/wiki/Installing) page in the **Installation** section of this document, it seems like those would be more suited to a spot in that page. Despite that opinion, it's fine with me if you'd like to leave them in this document.

Without further ado, here are the copies of the original file from the repository, @SamiSepehri's file, and my file:

<mark>**IMPORTANT:** GitHub doesn't support the attachment of RST files in these comments, so I had to change the extension on these files to **txt** and you'll want to rename them with the **rst** extension if you'd like to view them in a document viewer to see how they render.</mark>

* <mark>[Original.txt](https://github.com/user-attachments/files/18988185/Original.txt)</mark>
* <mark>[Elliria.txt](https://github.com/user-attachments/files/18988186/Elliria.txt)</mark>
* <mark>[SamiSepehri.txt](https://github.com/user-attachments/files/18988187/SamiSepehri.txt)</mark>

PR #1028, opened 2025-03-10
Update CHANGELOG.rst - fix #1027 by updating section headers

### LINKED ISSUE #1027: CHANGELOG.rst on master branch is flagged

Xorg

- [x] This is not a question.

Documentation

- [ ] autokey triggers
- [ ] autokey-gtk
- [ ] autokey-qt
- [ ] beta
- [ ] bug
- [ ] critical
- [x] development
- [ ] documentation
- [ ] enhancement
- [ ] phrase expansion
- [ ] scripting
- [ ] technical debt
- [ ] user interface

_No response_

_No response_

None

_No response_

_No response_

The 

Image

 file on the **master** branch is flagged as failing. I see why it's flagged. It's got two section headers that are out of sequence (on [line 49](https://github.com/autokey/autokey/blob/87ace10981b00bee1e5069fb7b1b5da9770a80bb/CHANGELOG.rst#L49) and [line 223](https://github.com/autokey/autokey/blob/87ace10981b00bee1e5069fb7b1b5da9770a80bb/CHANGELOG.rst#L223). It's an easy fix.

Note that the flag is for **Python 3.13**, which we don't yet use. It may have been put there by our PR tests because of a recent attempt to use **Python 3.13** that was since undone or it was put there by Sphinx or it was put there by GitHub as a friendly alert.

The [CHANGELOG.rst](https://github.com/autokey/autokey/blob/develop/CHANGELOG.rst) file on the **develop** branch isn't flagged, but we do still make changes to the **master** branch that involve updating [CHANGELOG.RST](https://github.com/autokey/autokey/blob/master/CHANGELOG.rst) file on the **master** branch.

Should I fix the issue? Don't bother?

Always

1. Visit the [CHANGELOG.RST](https://github.com/autokey/autokey/blob/master/CHANGELOG.rst) page.
2. Look above the document for the red x (marked with a green arrow in the image below).

Image

No flag.

### What actually happened?

_No response_

### Do you have screenshots?

_No response_

```bash

```

### Anything else?

_No response_
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #1027</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @Elliria</code> (replace <code>20</code> with the amount, and <code>@Elliria</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

Elliria:
Fix the two out-of-sequence section headers in CHANGELOG.rst at [line 49](https://github.com/autokey/autokey/blob/87ace10981b00bee1e5069fb7b1b5da9770a80bb/CHANGELOG.rst#L49) and [line 223](https://github.com/autokey/autokey/blob/87ace10981b00bee1e5069fb7b1b5da9770a80bb/CHANGELOG.rst#L223) to satisfy issue #1027.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ Other
 +++++
 - Add `show_recent_script_errors_dialog` to display errors with millisecond precision.
 - Add **Environment** as an issue type in the `bug.yml` file.
+- Fix two out-of-sequence section headers in CHANGELOG.rst to satisfy issue #1027.
 
 Version 0.96.0
 ============================
@@ -46,7 +47,7 @@ Features
 ---------
 
 Scripting API
-^^^^^^^^^^^^^
++++++++++++++
 
 **engine API object**
 
@@ -220,7 +221,7 @@ Features
 ---------
 
 Scripting API
-^^^^^^^^^^^^^
++++++++++++++
 
 **engine API object**
 


Elliria:
Another test error, and this one is tricky [sigh]. I just checked the action and Python versions and they're up-to-date. If the discussion in [this issue in a different GitHub repository](https://github.com/beeware/toga/issues/3143) is any indication, this won't be an easy fix and will probably require the intervention of someone with more than my limited understanding of how to work on these things.

The [apt-requirements.txt](https://github.com/autokey/autokey/blob/87ace10981b00bee1e5069fb7b1b5da9770a80bb/apt-requirements.txt#L7C1-L7C23) file lists **libgirepository1.0-dev** package as a current dependency.

[The complaint](https://github.com/autokey/autokey/actions/runs/13769881479/job/38505591588?pr=1028#step:6:1337) is:
```
Run-time dependency girepository-2.0 found: NO (tried pkgconfig and cmake)
Not looking for a fallback subproject for the dependency girepository-2.0 because:
Use of fallback dependencies is disabled.
```

The problem is that it's looking for the **girepository-2.0** dependency, which isn't available in Ubuntu 22.04. Ubuntu 22.04 offers the **libgirepository-1.0-1** package, which we could try to put in as a dependency to see if it makes the tests happy, but I suspect that it won't because  of the two last lines of that error message.

Any help with this would be appreciated.

Elliria:
In looking at that other repository's issue discussion a bit more, the simplest solution to this may be to make Python **3.11** our highest-supported Python version until GitHub no longer supports **Ubuntu 22.04**.

josephj11:
Is there a way to set something like this as "expected to fail"? I've seen something like that around here, maybe in the tox stuff.

I think I also saw some sort of conditional statement somewhere that would make a test only apply if some condition was met.

IDK if either of these is relevant here. Claude.ai said this, but the question should probably be refined.

```
I have a github pr that is failing tests because it needs something that is not yet present in python 3.10 and .11 but is in 3.12. Is there a way to ignore the error with something like expected to fail or a conditional requirement on  the test

Edit
You can handle version-dependent tests in Python using several approaches:

Use unittest.skipIf or unittest.expectedFailure:
python

Copy
import unittest
import sys

class MyTestCase(unittest.TestCase):
    @unittest.skipIf(sys.version_info < (3, 12), "Requires Python 3.12+")
    def test_python312_feature(self):
        # Test code that uses Python 3.12 features
        
    @unittest.expectedFailure
    @unittest.skipIf(sys.version_info >= (3, 12), "Should pass in Python 3.12+")
    def test_expected_to_fail_before_312(self):
        # Test that will fail on 3.10/3.11 but pass on 3.12+
With pytest, you can use similar markers:
python

Copy
import pytest
import sys

@pytest.mark.skipif(sys.version_info < (3, 12), reason="Requires Python 3.12+")
def test_python312_feature():
    # Test using 3.12 features

@pytest.mark.xfail(sys.version_info < (3, 12), reason="Expected to fail before Python 3.12")
def test_expected_to_fail_before_312():
    # Test that should fail on 3.10/3.11
For conditional imports, you can use try/except:
python

Copy
import sys

if sys.version_info >= (3, 12):
    # Import modules or use features only available in 3.12+
    from some_module import new_feature
else:
    # Define alternative implementation or mock for older versions
    new_feature = None

# Then in your tests, check if the feature is available
Which approach works best depends on your testing framework and specific requirements. What testing framework are you using?
```

Elliria:
Maybe, but I'm not sure if it would be a good idea or how to implement it.
* The **test-install (3.10)** test is failing at the **Install OS dependencies only** step.
* The **pytest (3.12)** test is failing at the **Install dependencies** step.
* The **Lint (3.12)** test is failing at the **Install dependencies** step.

In each of those, that step comes before any of the tests are run, so it's failing at fetching what it needs so that it can run its tests.

Elliria:
Some new info for this:

Tests can be created with assertions in any file with a name that starts with **test_** (ours are in the [tests](https://github.com/autokey/autokey/tree/develop/tests) directory). Tests can be marked with **skip** or **xfail** wrappers. Marking a test with **skip** prevents that test from being run at all. Marking a test with **xfail** indicates that you expect it to fail (useful for a new feature or an unfixed bug). Here's an example of a wrapper in the [pytest documentation](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions). We've got several such wrappers already. There's always the hope that at least one of them will surprise you and pass as the result of some change that was made to the code. Meanwhile, the fact that those skips and xfails are there is an open invitation to developers to take them on and solve them.

Elliria:
I'm closing this one since it no longer needs to happen. The one on the **develop** branch looks good.

PR #1056, opened 2025-11-12
Create python-publish.yml

finaljrad-IR:


finaljrad-IR (commented):
approve

finaljrad-IR (commented):
2371d3a83c5a0dc6c4da192df9c1502205b8db53

Elliria:
Thank you for your interest in the AutoKey project and for submitting a pull request. If I understand this workflow correctly, it publishes AutoKey releases to [PyPI](https://pypi.org/). Our [build,yml](https://github.com/autokey/autokey/blob/master/.github/workflows/build.yml) workflow already does that and runs automatically whenever a new release is created, so I don't think this PR will be necessary.

This can be verified by examining [this section](https://github.com/autokey/autokey/blob/master/.github/workflows/build.yml#L142-L146) of our build,yml workflow file and by doing [a search for "autokey" on PyPI](https://pypi.org/search/?q=autokey) and clicking on the [autokey](https://pypi.org/project/autokey/) entry in the search results to visit AutoKey's PyPI project page.

Interestingly, there may be some learning from this PR. Our [build,yml](https://github.com/autokey/autokey/blob/master/.github/workflows/build.yml) workflow is authenticated by using a **repo-token** to verify that its deployment originates from the AutoKey GitHub repository, whereas this PR uses **id-token: write** to presumably do the same thing much more succinctly. My understanding of how those tokens work isn't solid enough to be sure whether those are two different things or the same thing written differently, but if they're the same, maybe we should use the more succinct version in the future.

PR #1068, opened 2026-02-06
Added ability to call autokey-run with script-arguments that are passed to autokey scripts

adventuretc:
I PR to master because I noticed the development branch hasn't been used for 3 years.

Also added a test for the new feature and a test runner script that runs all tests in one go.

Manually tested also Qt and Gtk version.

### DIFF
--- a/autokey-run
+++ b/autokey-run
@@ -21,6 +21,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import sys
+import json
 import dbus
 import argparse
 
@@ -81,18 +82,19 @@ if __name__ == "__main__":
             else:
                 script_args.append(raw_arg)
 
-        try:
-            dbus_args = dbus.Array([dbus.String(arg) for arg in script_args], signature='s')
-            dbus_kwargs = dbus.Dictionary({dbus.String(k): dbus.String(v) for k, v in script_kwargs.items()}, signature='ss')
-            dbus_service.run_script_with_arguments(arguments.name, dbus_args, dbus_kwargs, dbus_interface="org.autokey.Service")
-            sys.exit(0)
-        except dbus.DBusException as e:
-            if e.get_dbus_name() == "org.freedesktop.DBus.Error.UnknownMethod":
-                if script_args or script_kwargs:
+        # Send args as a single JSON string over DBus to avoid type marshalling
+        # issues between python-dbus (GTK) and Qt DBus backends.
+        if script_args or script_kwargs:
+            args_json = json.dumps({"args": script_args, "kwargs": script_kwargs})
+            try:
+                dbus_service.run_script_with_args_json(arguments.name, args_json, dbus_interface="org.autokey.Service")
+                sys.exit(0)
+            except dbus.DBusException as e:
+                if e.get_dbus_name() == "org.freedesktop.DBus.Error.UnknownMethod":
                     print("Connected AutoKey instance does not support script arguments; running without them.", file=sys.stderr)
-            else:
-                print(e.get_dbus_message(), file=sys.stderr)
-                sys.exit(1)
+                else:
+                    print(e.get_dbus_message(), file=sys.stderr)
+                    sys.exit(1)
 
         dbus_function = dbus_service.run_script
     elif arguments.mode == "phrase":
--- a/autokey-run
+++ b/autokey-run
@@ -5,6 +5,7 @@
 
 # Copyright (C) 2011 Chris Dekter
 # Copyright (C) 2017 Thomas Hess
+# Copyright (C) 2026 Bertalan Göller <intmianol@disroot.org>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
--- a/autokey-run
+++ b/autokey-run
@@ -81,7 +81,9 @@ if __name__ == "__main__":
                 script_args.append(raw_arg)
 
         try:
-            dbus_service.run_script_with_arguments(arguments.name, script_args, script_kwargs, dbus_interface="org.autokey.Service")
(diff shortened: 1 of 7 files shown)


josephj11:
Thanks so much for this PR, @adventuretc .

We're currently stuck without a lead developer, so we can't release anything.

The main roadblock is that develop is out of sync with main. Work was started on fixing that, but our former lead developer left the project to pursue other goals.

We don't want to make the sync problem worse, so we aren't currently merging anything new into master unless it's a tiny change like updating Python versions in tests...

The secondary problem is that I would like to create a maintenance release 0.96.1 with just fixes, but develop has major changes in it for Wayland support, so it's 0.97.0 and 0.96.1 would have to be cherry picked  out of there.

Since this PR is an enhancement - a new feature - it would probably go into 0.97.0.

With proper testing, I would be willing to merge it into develop because (hopefully) it will not increase our sync problems.

Again, this is mostly academic, because we can't currently release anything.

adventuretc:
I see, @josephj11.

Maybe we could set up a project on an open source funding platform to aid development and Wayland support. Have you considered this possibility yet?

Open Collective, Thanks.dev, and there seems to be a lot more.

https://viktornagornyy.com/funding-open-source-projects/

josephj11:
We're already on Opire and the Wayland issue had around $200 total pledged from a number of donors last I looked. That's the only issue with a bounty AFAIK. That amounts to a token of appreciation,  but only a few hours of paid time at best.

Elliria:
Thank you for the PR, @adventuretc. A couple of comments: I understand why you chose to do the PR against the **master** branch, but AutoKey only accepts PRs to the **develop** branch, so it will need to be rebased. Also, the additional copyright will need to be removed from the `lib/autokey/qtui/dbus_service.py` and `lib/autokey/dbus_service.py` and `tests/scripting_api/test_engine.py` and `autokey-run` files. The rest of it is beyond me. Hopefully, someone who understands the code will review this PR.

BlueDrink9:
Code itself looks good at a glance, although this should be checked out and tested locally by someone. The shell script feels unnecessary to me. The idea of redirecting the configuration directory is a very good one, but it should probably be in a test fixture rather than a dedicated bash script. 

However, since you are adding a new feature, I would like you to give a description of why you think it is useful or necessary. Every user of AutoKey uses the GUI as its primary entry point. I briefly added a headless scripting mode a while back, but shortly realized that other tools usually are better suited if you want a headless setup. It would be good to have an argument for why this would be worth the increased maintenance burden, and to double check that you couldn't do it another way.

Thanks for the PR!

josephj11:
This comes under the heading of nice, but not necessary because it's already easy to pass data/parameters to scripts using AutoKey stored local and global variables as well as using Python to interact with the rest of the Linux environment, e.g. files.

There's also a symmetry issue unless we also enhance the rest of AutoKey  to allow passing parameters to scripts somehow. How to do that is less than obvious. This leads me to think that maybe this isn't such a good idea.

If you're doing desktop automation that doesn't require AutoKey's main virtue, triggering from a hotkey or trigger phrase, it's way easiers to do that using xdotool (ydotool), etc. wthout using AutoKey at all and it's an order of magnitude easier to test and debug too.

adventuretc:
> However, since you are adding a new feature, I would like you to give a description of why you think it is useful or necessary. Every user of AutoKey uses the GUI as its primary entry point. I briefly added a headless scripting mode a while back, but shortly realized that other tools usually are better suited if you want a headless setup. It would be good to have an argument for why this would be worth the increased maintenance burden, and to double check that you couldn't do it another way.

I already had a long autokey script that handles file opening and wanted to call it from the outside, passing a parameter. The script issues keystrokes and reads the clipboard which I thought were more the strengths of autokey so I thought I would leave it up to it. Do you think writing a keyboard and clipboard handling "backend" with imports would be better?
I recognize this is a rare use case and respect if you deem it too heavy for the project's maintenance.

adventuretc:
Having that Opire is nice, but it's nowhere visible on the main page of this repo or otherwise, and, also, Autokey has no project-wide crowdfunding platform presence which I think is a lost opportunity.

josephj11:
@adventuretc I'm not an accountant. If you know, please tell me what AutoKey would have to do legally to be able to accept money from any source and not pay taxes on it and how much that would cost to setup and maintain?

How much would we have to receive in donantions just to break even?

Then, give me some ideas on exactly what to do with the money that's left. It would have to be a lot to actually pay a competent programmer to do anything significant. What about me and the other volunteers? How much is our work worth? (I have put in thousands of hours and others have done a lot.)

There are a couple of organizations setup to handle all the accounting,  etc., but naturally, they're not free.

Maybe if we got a corporate sponsor or other significant donor, this would all make sense, but that hasn't happened and we haven't had anyone make presentations at conferences, etc. to generate that sort of interest.

Anyone is free to setup a Patreon account or publish a PayPal account or Bitcoin address, etc. and offer their services. No one has done that so far.

This is not criticism. It's just to give an idea of how something that sounds simple really isn't.

adventuretc:
I understand.
Well, in my country Hungary receiving donations (legally *gifts*) is tax-free under a certain amount (most probably enough for this purpose ~470 USD), you just have to register that you received something and the name and data of the other party (person or, probably also when you receive from an organization like those platforms).

The distribution between contributors can also be solved this way, on a personal agreement basis.

josephj11:
@adventuretc That works great for you or me, but there isn't an "AutoKey" entity to receive the funds.

So, as it stands now, any donor... is free to give money, etc. to any person(s) on the project as long as they agree to accept it along with any terms that are specified by the donor. You just can't give it to the project itself.

My  only concern with that would be if someone gives to some people, but not to others, etc. I don't want to have to deal with hard feelings, etc.

PR #1070, opened 2026-02-10
Show a clear error on Wayland without X11

swjcpy:
Detect Wayland sessions without an X11 display and show a clear error explaining that AutoKey currently requires X11/XWayland. This avoids opaque startup failures.

### DIFF
--- a/lib/autokey/UI_common_functions.py
+++ b/lib/autokey/UI_common_functions.py
@@ -78,6 +78,21 @@ def checkRequirements():
     return errorMessage
 
 
+def get_wayland_error_message():
+    session_type = os.environ.get('XDG_SESSION_TYPE', '').lower()
+    has_wayland_display = bool(os.environ.get('WAYLAND_DISPLAY'))
+    has_x11_display = bool(os.environ.get('DISPLAY'))
+
+    if (session_type == 'wayland' or has_wayland_display) and not has_x11_display:
+        return (
+            "Wayland session detected but no X11 display is available. "
+            "AutoKey currently requires X11. Please run under an X11 session "
+            "or enable XWayland."
+        )
+
+    return ""
+
+
 def create_storage_directories():
     """Create various storage directories, if those do not exist."""
     # Create configuration directory
--- a/lib/autokey/gtkapp.py
+++ b/lib/autokey/gtkapp.py
@@ -78,6 +78,13 @@ class Application:
             Gdk.threads_leave()
             sys.exit("Missing required programs and/or python modules, exiting")
 
+        wayland_error = UI_common.get_wayland_error_message()
+        if wayland_error:
+            Gdk.threads_enter()
+            self.show_error_dialog("AutoKey requires X11 to run.", wayland_error)
+            Gdk.threads_leave()
+            sys.exit(1)
+
         try:
             create_storage_directories()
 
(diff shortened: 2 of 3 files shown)


josephj11:
Welcome to the AutokKey community, @swjcpy !

This addresses a pet peeve of mine. See #600.

I briefly reviewed the code and it looks OK to me, but I am not one of the devs.

This assumes that there will be a 0.96.1 release since 0.97.0 will support Wayland, so it may need to take that into account because there isn't a 0.96.1 branch and it's not clear that there will be one.

@Elliria and @dlk3 Please take a look at this if you  can. I'd love to merge it with your approval.  Could this be added to the master branch in a way that doesn't create technical debt in develop?

swjcpy:
Thanks! Feel free to let me know if questions.

Elliria:
Thank you for the PR, @adventuretc. A couple of comments: AutoKey only accepts PRs to the develop branch, so it will need to be rebased.  Also, although this has no effect on how the PR works, I'd replace the space after **session** in line 89 of `lib/autokey/UI_common_functions.py` with a period. Other than that, it looks like it's a good PR that will temporarily address the Wayland issue while the Wayland code is being worked on.  I haven't tested it, though, and it will need to be tested before being merged. My computer is currently in a state of transition, so it will be a while before I can test things.

@Joe, this would go onto the **develop** branch because it would definitely create technical debt if it went onto **master**. And I think you're right that since it's an enhancement, it should go into the next version release instead of being an update to the current release. If you'd like, we can choose **Feature release 0.97.0** from the **Milestone** drop-down on the right side of this PR so we'll know that that's where it should go.

dlk3:
A warning message would be a nice thing, it's true.  But I take @josephj11 point too, that the project maintainers see value in keeping the "master" branch frozen.  Beyond that I'd point out that the "develop" branch code supports both the X11 and Wayland environments already.  So there's no need for a warning message long term.

Also, I received notice that you, @swjcpy, have claimed the Opire reward for #87 on the strength of this PR.  I am one of the contributors to that award.  I do not consider the simple addition of a warning message to be sufficient to earn that reward so I won't be paying out my portion of the award.  It is my opinion that @sebastiansam55 has already earned that award with the code that he put into the "develop" branch several years ago .  As evidence of this, see [my fork](https://github.com/dlk3/autokey-wayland) of the "develop" branch which I've polished up and made to run.  That version runs on both X11 and Wayland desktops.

josephj11:
No one is giving the award to a poser, but there's the real problem that AFAIK, @sebastiansam55  dropped off GitHub entirely quite some time ago and we don't want to send a bounty into /dev/null. Since Opire doesn't collect the bounties in advance, it's not that big a problem though.

dlk3:
Yeah, I get it that @sebastiansam55 is not talking to us, but he's the one who did the work.  Looking at his GitHub profile shows him active in December.  That's a bit too recent for me to declare him dead.  Rather than just cancel my Opire offer, I prefer to leave it out there in case he ever comes back around.  Especially since, as you say, it costs me nothing to do so.

On the other hand, I could lobby for the reward myself, I guess, since I have made AutoKey work under Wayland and X11 over in [my fork](https://github.com/dlk3/autokey-wayland).  ```<grin>```

josephj11:
@dlk I totally understand burnout/wanting to leave a project and developers don't owe us anything, but I really dislike being ghosted.

@sebastiansam55 We're starting to get close to a release of AutoKey for Wayland. There's a small bounty for that on Opire and some of us think you ought to get most or all of it. Please let us know if you want it. We don't expect you to do anything further on AutoKey unless you want to.

Elliria:
From the comments above, it looks like this one probably should be closed since it would only be necessary if AutoKey wasn't Wayland-capable, which it now mostly is in the development code.

josephj11:
That's the best reason to close a PR! It addresses a deficiency that is no longer present.

PR #1071, opened 2026-02-12
Merge BlueDrink's PR 1004 first and then bring develop the rest of the way up to sync with master

dlk3:
Once again, the goal is to bring the develop branch into sync with master.  This PR takes a better approach than my previous submission (#1069), I think, because it leverages BlueDrink9's expertise with the code and the work that he's done in PR #1004.

This PR was created by first merging @BlueDrink9's PR #1004 into develop.  There were conflicts in doing that.  They were generally the same ones I saw when I did my own merge for #1069.  There were a couple that @BlueDrink9 had dealt with, but by-and-large the conflicts were the same.  I fixed them, same as before.

I then merged the master branch into this new version of develop.  There were a couple of trivial conflicts in this merge that I fixed.

This PR then, is @BlueDrink9's PR #1004, with all conflicts resolved, plus the latest few commits from the master branch.  This puts the two branches in sync with one another.

Here's a [log](https://github.com/user-attachments/files/25267706/merge.log) of the merge process that built this.

PR #1072, opened 2026-02-15
Merge BlueDrink's PR 1004 first and then bring develop the rest of the way up to sync with master

dlk3:
Ok, here's the same as #1071, but this time it's properly targeted at the develop branch.  

My apologies for my lack of paying proper attention to what I'm doing.


josephj11:
@dlk3 No problem. AFAIK, it's not that hard to fix such things in git/GitHub as long as they get detected before other changes get applied on top of them. 

God bless Linux... It always assumes you know what you're doing and helps you do it. ;)

josephj11:
@dlk3 I merged it anyway, but none of the tests in the action succeeded. Most were cancelled. I thought at least some of them were supposed to pass.

Given that this is develop, I'm not clear on why there are any tests or if they ever worked.

dlk3:
I have no expectation that the tests will work at this point.  I just took it for granted that, since I couldn't even get the develop branch to start, as @Elliria just found out, the tests would fail.  @sebastiansam55's changes in the develop branch made dramatic changes to the module content and layout in the underlying code that is used under both X11 and Wayland.  The tests are going to have to be rewritten to take those changes into account if you're going to continue to use them.  I have no experience with tox, but I did look at a couple of the tests that were failing to confirm that this is what was happening.

josephj11:
@dlk3 is there a way to disable the tests until they are refitted?

Are we supposed to have tests on develop and are they supposed to be the same tests as we have on the master?

Elliria:
We expected the tests to fail on this PR since the fixes for those aren't merged in yet.

You can disable the tests temporarily by opening [Actions](https://github.com/autokey/autokey/actions) in the repository's menu and selecting **Python test** in the left panel, which will put you [here](https://github.com/autokey/autokey/actions/workflows/python-test.yml). Then, click the **...** in the upper-right corner to the right of the **Filter workflow runs** text-box and choose **Disable workflow** from the drop-down menu that opens. Note that you can re-enable it from that same drop-down menu later. I wouldn't recommend doing that, though. It's better to see the tests failing as a reminder to fix them.

Elliria:
I suspect that the tests won't need to be rewritten, but rather will need to be satisfied. If my understanding of how they work is correct, they're designed to run the AutoKey code in each possible environment (specified by versions and dependencies), fail if it doesn't run, and give you the most cryptic possible reasons for the failures when they happen so that you have to bend and snap to try to figure them out. Basically, once the code is solid and behaves itself, the tests should (fingers crossed) all pass.

Elliria:
@josephj11:
 
> Are we supposed to have tests on develop and are they supposed to be the same tests as we have on the master?

We have three workflow files: [build.yml](https://github.com/autokey/autokey/blob/master/.github/workflows/build.yml), [pages.yml](https://github.com/autokey/autokey/blob/master/.github/workflows/pages.yml), and [python-test.yml](https://github.com/autokey/autokey/blob/master/.github/workflows/python-test.yml). The tests are configured to run on any event that occurs on **master**, **develop**, or **beta**.

I believe that the workflow files, if present on the branch that triggers an event, will be used, but some events require a workflow file to be on the default (**master** for us) branch.

In theory, you could have the same **python-test.yml** file on **master** and **develop** and update only the one on **develop** and then only the **develop** branch would use the updated tests.

On a related note, I'm a bit unclear on how **build.yml** works because it specifies the **CI** branch, which we don't have. I can't remember whether that was a deleted branch that needs to be recreated before doing a new build or whether the **CI** that it refers to is an elusive reference rather than a reference to an actual branch. We'll surely be finding out soon whether that just runs on its own when a release tag is created on the **beta** or **master** branch (which is what I was given to understand) or whether it will need some intervention from us in the form of restoring the branch first.

josephj11:
@Elliria  I have never heard of an AutoKey branch with CI (continuous integration) in the name. I would guess it "should" be something like  a symlink to develop.

Elliria:
Yeah, I might be getting the deleting and restoring of a branch mixed up with the **beta** branch, which could, conceivably, be deleted and restored as needed without doing any harm, but which we just leave dormant between uses. We'll find out soon enough.

josephj11:
@Elliria I'm not sure, but I vaguely recall something about Jenkins (CI/CD) being setup in the past.

dlk3:
> @dlk3 is there a way to disable the tests until they are refitted?
> 
> Are we supposed to have tests on develop and are they supposed to be the same tests as we have on the master?

Sorry @josephj11, these questions got past me.  I have no experience with the tox tool you're using, but I do know that some of your tests have been set to "expected to fail," which effectively turns them off.  The same could easily be done to the remaining failing tests as well, until such time as those tests can be fixed to work with the new code.

Should the tests be the same across develop and master?   Maybe, maybe not.  The commits coming into "master" need to be "perfect", so you want stringent testing there.   Do you need the same level of perfection in the commits that come into the develop branch?  Maybe you want to allow a little more leniency there.  Maybe test the commits to ensure they don't totally break the entire app, but don't block the introduction of new code or changes in functionality.

What you definitely don't want is to enforce a test routine that stifles development.  You're in kind of a unique situation because you no longer have a maintainer who has the knowledge necessary to do code reviews of new commits.  Unfortunately, in a dynamic development environment, a test suite can't substitute for a skilled developer.  It definitely doesn't work when you're faced with a significant restructuring of the application's architecture, like you are now with the introduction of Wayland.

josephj11:
@dlk3 I think @luziferius set up tox and @BlueDrink9 knew his way around them.

We'll have to do something about them before a new release because Debian was stuck for years on old releases because they wouldn't package a new release with failed tests. I still don't know how they got 0.96.0 because it had failed tests related to Python versions.

Elliria:
@josephj11:

> @Elliria I'm not sure, but I vaguely recall something about Jenkins (CI/CD) being setup in the past.

It looks like **CI** was a branch on an external repository. See the source of [this PR](https://github.com/autokey/autokey/pull/544) as an example during the time when its functionality was introduced to the AutoKey repository.

luziferius:
> @dlk3 I think @luziferius set up tox

This may have been the case. I'd have to dig through the history to confirm, but sounds likely.

I do use it in other Python+Qt projects for additional build steps, since pure Python build toolings, like `build` still cannot do external commands that are needed for Qt resource, localization and UI compilation.

This was the path of least resistance and most cross-platform approach that avoids having shell scripts laying around in the repo.

I had very little experience with tox back then, so if that's still in use, it may need a slight overhaul.

josephj11:
@luziferius Thanks for your input. I trust all is well with you.

PR #1075, opened 2026-02-15
Fix regex strings

### LINKED ISSUE #536: Update deprecated regex sequences
## Classification:

Maintenance

## Version

AutoKey version: develop  5963c7ceae668736e1b91809eff426a47ed5a507

## Summary

The CI tests are warning about several deprecated regex sequences. We should find alternatives and replace them.

The regex is one of the better-tested parts of the code in terms of unit tests, so we should be safe with what we change.

```
lib/autokey/model/key.py:122
  /home/runner/work/autokey/autokey/lib/autokey/model/key.py:122: DeprecationWarning: invalid escape sequence \+
    KEY_SPLIT_RE = re.compile("(<[^<>]+>\+?)")

lib/autokey/model/helpers.py:21
  /home/runner/work/autokey/autokey/lib/autokey/model/helpers.py:21: DeprecationWarning: invalid escape sequence \w
    DEFAULT_WORDCHAR_REGEX = '[\w]'

lib/autokey/model/abstract_abbreviation.py:143
  /home/runner/work/autokey/autokey/lib/autokey/model/abstract_abbreviation.py:143: DeprecationWarning: invalid escape sequence \s
    if len(stringBefore) > 0 and not re.match('(^\s)', stringBefore[-1]) and not self.triggerInside:
```

T0mstone:
I found these with `python -m compileall -q .`:
```
./lib/autokey/model/abstract_abbreviation.py:143: SyntaxWarning: "\s" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\s"? A raw string is also an option.
  if len(stringBefore) > 0 and not re.match('(^\s)', stringBefore[-1]) and not self.triggerInside:
./lib/autokey/model/helpers.py:21: SyntaxWarning: "\w" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\w"? A raw string is also an option.
  DEFAULT_WORDCHAR_REGEX = '[\w]'
./lib/autokey/model/key.py:122: SyntaxWarning: "\+" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\+"? A raw string is also an option.
  KEY_SPLIT_RE = re.compile("(<[^<>]+>\+?)")
```

Edit: Oh, there was already an issue for this...
Fixes #536

### DIFF
--- a/lib/autokey/model/abstract_abbreviation.py
+++ b/lib/autokey/model/abstract_abbreviation.py
@@ -140,7 +140,7 @@ class AbstractAbbreviation:
 
             # Check chars ahead of abbr
             # length of stringBefore should always be > 0
-            if len(stringBefore) > 0 and not re.match('(^\s)', stringBefore[-1]) and not self.triggerInside:
+            if len(stringBefore) > 0 and not re.match(r'(^\s)', stringBefore[-1]) and not self.triggerInside:
                 # check if last char before the typed abbreviation is a word char
                 # if triggerInside is not set, can't trigger when inside a word
                 return False
--- a/lib/autokey/model/helpers.py
+++ b/lib/autokey/model/helpers.py
@@ -18,7 +18,7 @@ import enum
 import os
 import re
 
-DEFAULT_WORDCHAR_REGEX = '[\w]'
+DEFAULT_WORDCHAR_REGEX = r'[\w]'
 JSON_FILE_PATTERN = "{}/{}.json"
 SPACES_RE = re.compile(r"^ | $")
 
--- a/lib/autokey/model/key.py
+++ b/lib/autokey/model/key.py
@@ -119,7 +119,7 @@ _ALL_MODIFIERS_ = (
 
 # Used to identify special keys in texts. Also include <codeXX> literals as defined in the _code_point_re.
 KEY_FIND_RE = re.compile("|".join(("|".join(Key), _code_point_re.pattern)), re.UNICODE)
-KEY_SPLIT_RE = re.compile("(<[^<>]+>\+?)")
+KEY_SPLIT_RE = re.compile(r"(<[^<>]+>\+?)")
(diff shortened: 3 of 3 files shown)


josephj11:
Welcome  to the AutoKey community @T0mstone ! Thanks for contributing to AutoKey.

All PRs (except for things like emergency fixes) need to be based on our develop branch, not on the master.

There is currently a lot of activity on develop in an effort to sync it to master and get it ready for testing on Wayland.
A big change just went in and there may be another one or two in the next few days, so you may want to wait a few days before rebasing this fix on develop. IDK if the files you are modifying are affected or not.

Once things stabilize, we look forward to reviewing your new PR and merging it if it is approved.

dlk3:
I found a number of these regex bugs in the develop branch.  I have fixes for them in the next PR I'm going to submit in a few days.

chrislee35:
Thanks for taking the time to track this down and put up a fix, @T0mstone — and sorry this sat without a follow-up for so long.

Checking in on this today, I don't think it makes sense to move forward with it as-is, for a couple of reasons:

1. **Base branch**: as @josephj11 noted back in February, PRs need to target `develop`, not `master`.
2. **Already resolved on `develop`**: I checked all three sites this PR touches against current `develop`, and the underlying warnings from #536 are already gone there, fixed independently by other changes:
   - `key.py` and `abstract_abbreviation.py` were fixed in a different commit, using an escaped backslash (`'\\s'`) rather than a raw string — same effect, different spelling.
   - The `helpers.py` line (`DEFAULT_WORDCHAR_REGEX = '[\w]'`) doesn't exist anymore; it was refactored away into a `make_wordchar_re()` function.

So even rebased onto `develop`, this PR wouldn't have anything left to change. I'll close this out along with #536 as already fixed. The fix itself was the right call (raw strings are the cleaner way to avoid this warning), just overtaken by other work on develop in the meantime — appreciate the contribution regardless.

PR #1083, opened 2026-03-07
fix: get version from common.py instead of git tags in debian/build.sh

### LINKED ISSUE #1082: Possible fix for issues in debian/build.sh

There are two possible conflicts between our [release-cycle strategy](https://github.com/autokey/autokey/wiki/Release-Cycle) and [line 6 in the debian/build.sh file](https://github.com/autokey/autokey/blob/87ace10981b00bee1e5069fb7b1b5da9770a80bb/debian/build.sh#L6) that may/will cause our local VM builds to fail if they aren't fixed when we run that script. This is the line:
```
VERSION=$(git describe --tags --abbrev=0 --match "v*.*.*")
```

The version is obtained by `git describe` looking for the most recent **tag** in the git history. Since our release cycle has us resetting **develop** to **master**, then adding new code to **develop**, and then resetting **beta** to **develop**, the most recent **tag** in the git history on **develop** or **beta** will be the **tag** for the old version on **master**. As a result, it will build a **.deb** using the old version, which we wouldn't want and which might prevent it from installing entirely or might overwrite our stable version. The fix for that could be to get the version from AutoKey's `common.py` file. This can optionally be followed by offering it the alternative of falling back to the tag that it used to get the version from if no version is specified in the `common.py` file, although the fallback is most likely unnecessary since the `common.py` file will always have a version in it.

Also, the `"v*.*.*"` portion of the line doesn't fully work with our naming conventions for releases. It **does** recognize our 3-part full-release versions (**vX.X.X**), but it **doesn't** recognize our 4-part **beta** versions (**vX.X.X-beta.X**), so those will be skipped and a beta won't be built. The fix for that could be to change the `"v*.*.*"` to `"v*"` to get it to match our **beta** tags, but then it would also match **any** tag starting with **v**, which could cause an accidental build. Instead, changing it to `"v[0-9]*"` will cause it to only do a build if there's a number after the **v**.

Always

The full release and the beta release should build.

### What actually happened?

Only the full release builds.

### Anything else?

Here are two possible ways to rewrite that line, with my preference being for the first one for the reason explained above:
* This is a replacement for the existing line that gets the version from the `common.py` file:
  ```
  VERSION=$(python3 -c "from lib.autokey.common import VERSION; print(VERSION)" 2>/dev/null)
  ```
* This adds to the existing line by getting the version from the `common.py` file while also offering the former way of getting the version number from the tag as a fall-back with the correction to match our naming conventions:
  ```
  VERSION=$(python3 -c "from lib.autokey.common import VERSION; print(VERSION)" 2>/dev/null || git describe --tags --abbrev=0 --match "v[0-9]*")
  ```

<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #1082</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @Elliria</code> (replace <code>20</code> with the amount, and <code>@Elliria</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

krishna3554:
Fixed version detection in debian/build.sh
The build script currently was using git describe --tags --abbrev=0 --match "v*.."

This can lead to incorrect version used because tags (vX.X.X-beta.X) are not matched properly.

This changes directly retrives the version from lib/autokey/common.py, ensuring correct version is been used while building debian package.

Fixes #1082

### DIFF
--- a/debian/build.sh
+++ b/debian/build.sh
@@ -3,7 +3,7 @@
 [ -f debian/build_requirements.txt ] && \
   cat debian/build_requirements.txt | xargs sudo apt-get install -y
 
-VERSION=$(git describe --tags --abbrev=0 --match "v*.*.*")
+VERSION=$(python3 -c "from lib.autokey.common import VERSION; print(VERSION)" 2>/dev/null)
 # Strip leading 'v' because that is invalid as a debian version number
 DEBVERSION="${VERSION#?}"
 uscan -dd


Elliria:
> [!CAUTION]
> This is code from a bot.

PR #1095, opened 2026-03-26
wayland: harden uinput path and avoid GNOME DBus hard-fail (#87)

vildanden-ai:
## Summary
Bounty-focused Wayland follow-up for #87.

This PR targets practical Wayland support behavior in real sessions by hardening the uinput path and removing hard failures when GNOME DBus window integration is unavailable.

### 1) Wayland uinput startup no longer hard-fails if GNOME DBus extension is missing
- In `IoMediator`, initializing `GnomeExtensionWindowInterface()` is now guarded.
- If GNOME DBus integration is unavailable, AutoKey falls back to a null window interface instead of failing startup.

### 2) Safe fallback window metadata path for Wayland
- Added `_NullWindowInterface` fallback implementation with safe defaults.
- Provides:
  - empty `WindowInfo` (`wm_title`, `wm_class`) instead of exceptions
  - safe screen size default
  - no active-window assumptions

### 3) Defensive handling in uinput event path
- `handle_keypress` / `handle_keyrelease` now guard window info retrieval and use empty `WindowInfo` on failure.
- `relative_mouse_location()` now handles missing active window safely and returns absolute coordinates instead of crashing.

### 4) GNOME window interface hardening
- `GnomeExtensionWindowInterface` now safely handles missing active-window state and returns empty metadata strings rather than throwing.

## Why this is relevant to #87
On Wayland, many users run into runtime failures due to compositor/desktop-specific window APIs. This PR makes AutoKey continue operating (key processing/uinput path) even when GNOME-specific window metadata integration is unavailable or transiently failing.

This is an implementation-focused reliability step toward real Wayland usability, rather than a warning-only change.

## Validation
- `python3 -m py_compile` passed for:
  - `lib/autokey/iomediator/iomediator.py`
  - `lib/autokey/uinput_interface.py`
  - `lib/autokey/gnome_interface.py`

Refs: #87

### DIFF
--- a/lib/autokey/gnome_interface.py
+++ b/lib/autokey/gnome_interface.py
@@ -50,21 +50,24 @@ class GnomeExtensionWindowInterface(DBusInterface, AbstractWindowInterface):
         """
         Returns a WindowInfo object containing the class and title.
         """
-        window = self._active_window()
-        return WindowInfo(wm_class=window['wm_class'], wm_title=window['wm_title'])
+        active = self._active_window()
+        if not active:
+            return WindowInfo(wm_class="", wm_title="")
+        return WindowInfo(wm_class=active.get('wm_class', ''), wm_title=active.get('wm_title', ''))
 
     def get_window_class(self, window=None, traverse=True) -> str:
         """
         Returns the window class of the currently focused window.
         """
-        return self._active_window()['wm_class']
-        
-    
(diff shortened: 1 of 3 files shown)


vildanden-ai:
Closing this PR. It does not fully implement the requested Wayland support scope for #87.

PR #1099, opened 2026-04-07
Use raw triple-quotes to protect escape-sequences

Elliria:
Replace the triple-quote in line 34 of the `window.py` file with a raw triple-quote to prevent the **SyntaxWarning** about invalid escape-sequences that gets triggered by the escapes in lines 40 and 41 of the same file when tests are run under **Python 3.12**. This will also future-proof AutoKey, because the current **SyntaxWarning** is scheduled to [eventually become a **SyntaxError** in a future Python version](https://docs.python.org/3/whatsnew/3.12.html).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -91,6 +91,7 @@ Other changes
 - Rename the bug.yaml file to bug.yml.
 - Update the contents of the `bug.yml` file to make it identical with its counterpart on the **master** branch.
 - Add the `config.yml` file to the `/.github/ISSUE_TEMPLATE` directory to match `its counterpart`_ on the **master** branch.
+- Use raw-string format in `window.py` to handle invalid escape sequences.
 .. _its counterpart: https://github.com/autokey/autokey/blob/master/.github/ISSUE_TEMPLATE/config.yml
 
 Version 0.96.0-beta.9
--- a/lib/autokey/scripting/window.py
+++ b/lib/autokey/scripting/window.py
@@ -31,7 +31,7 @@ XRANDR_MONITOR_REGEX = r" (\d{3,4}).*?x(\d{3,4})\/.*?\+(\d{1,4})\+(\d{1,4})"
 # this translates to monitor x and y size, and x and y offset for each monitor
 
 class Window:
-    """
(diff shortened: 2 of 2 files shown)


Elliria:
Same friendly PSA warning as before that this isn't expected to give us green check-marks and that's okay.

josephj11 (commented):
LGTM. Merge?

Questions: What's the r/raw thing.  Never saw that before.
Does it make a difference if it's three single quotes or three double quotes? I haven't seen double quote used before.

Elliria:
The Python team didn't elaborate on what **raw** means, but we can deduce it from the warning and the recommended fis. My read of it is that a raw area is like a code-block here on GitHub. It's a hands-off area for the interpreter.

When **pytest** is used to run tests on AutoKey, a **SyntaxWarning** is currently being kicked out letting us know that [lines 33 through 42](https://github.com/autokey/autokey/blob/55cf8ab980ed095ff7623df854936483b3dc73f6/lib/autokey/scripting/window.py#L33-L42) of the [lib/autokey/scripting/window.py](https://github.com/autokey/autokey/blob/develop/lib/autokey/scripting/window.py) file contain characters that will cause a problem in a future Python release because the current **SyntaxWarning** is [scheduled to become a **SyntaxError**](https://docs.python.org/3/whatsnew/3.12.html).

The "offending" characters are the two backslash escapes in that triple-quoted docstring. Although they're not currently a show-stopper, the warning is being kicked up to give us time to fix our docstring before it becomes a **SyntaxError** in an as-yet-unspecified future Python version.

[PEP 257](https://peps.python.org/pep-0257/) instructs us to use the **r-prefix** to turn the docstring into a raw docstring (replace `"""` with `r"""`) to "talk" to the interpreter and tell it specifically not to interpret backslash escapes within those docstrings.

As for triple single-quotes, you **can** use them, but the official standard is to use triple double-quotes, as demonstrated in the PEP that I linked to above.

Elliria:
I feel comfortable recommending this for a merge, but it wouldn't hurt for you or someone else to also run tests on AutoKey without and with that change to see that the warning disappears when the change is made.

PR #1100, opened 2026-04-07
Bump GitHub Action and Python versions

Elliria:
Bump action and Python versions in the workflow files using [these steps](https://gist.github.com/Elliria/2dc56b4b0316d71819666d7bf43b726a) that provide links to the references I used.

Special attention was given to these two points:
* Use 2-segment versions to prevent version patches from causing test-failures, with the exception of the **sys.version_info** version in the `setup.py` file, which should be 3 segments since it makes sense in the context of its use there and won't be impacted by version patches or updates.
* Use hard-coded version instead of **master** for the **lewagon/wait-on-check-action** in the `build.yml` file to prevent updates from causing test-failures.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,18 +16,18 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.12.9]
+        python-version: [3.14]
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python-version }}
 
     - name: Cache pip
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         # This path is specific to Ubuntu
         path: ~/.cache/pip
@@ -53,7 +53,7 @@ jobs:
         mv ../*.deb debs
 
     - name: Upload built debs as artifacts
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v7
       with:
         name: debs
         path: debs
@@ -64,16 +64,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.12.9]
+        python-version: [3.14]
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python-version }}
 
     - name: Cache pip
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         # This path is specific to Ubuntu
         path: ~/.cache/pip
@@ -99,7 +99,7 @@ jobs:
         --outdir dist/
 
     - name: Upload built dist as artifact
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v7
       with:
         name: dist
         path: dist/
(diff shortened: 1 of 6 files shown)


Elliria:
Fingers crossed for green check-marks, but if not, no worries. There are more changes to come.

Elliria:
Note that the **pyhamcrest** version in the `setup.cfg` file also needs to be updated, but I'm doing that in a separate PR that will make several additional changes to this file since this PR is strictly for GitHub Actions and Python versions.

Elliria:
It looks like dependencies will take care of our lacking green check-marks. Those will be coming a bit later.

Elliria:
I feel comfortable recommending this for a merge, but it never hurts to have other people's eyes on the code. The changes I made can be found in the [Files changed](https://github.com/autokey/autokey/pull/1100/changes) tab above and the [Gist I linked to above](https://gist.github.com/Elliria/2dc56b4b0316d71819666d7bf43b726a) takes you to a document with links you can follow to verify that the versions I replaced are the currently-supported versions, that the corrections were made in the specified way (the notes let you know if the lowest or highest version or all versions should be used), and in all of the correct locations. It should only take a few minutes to go through that document and verify whether each step was followed correctly or not.

Elliria:
And for further explanation on replacing **@master** with, for example, **@v123**, this [issue in another repository](https://github.com/actions/upload-artifact/issues/41) gives a description of why specifying versions is a better idea. We only had one **@master** version, so I updated that one with an actual version in this PR.

josephj11:
@Elliria  I was about to merge this after looking at each set of changes, but it says that changelog.rst has merge conficts that must be resolved.

Selection_014

Please advise or just do it if you want to.

Elliria:
That's what happens when an unstoppable force meets an immovable object. Heh.

It's kind of like going to the mechanic to get a shiny new $4000 catalytic converter bolted on and then going back a few minutes later to get a shiny new $500 heat-shield bolted on and they get all confused about whether you want to keep the catalytic converter and forget about the pesky heat-shield or get rid of that pesky catalytic converter and put the heat-shield on or keep the catalytic converter and put the heat-shield on or replace them both with this shiny new $1850 radiator or get rid of them both entirely. It's up to you.

GitHub only ever has the repository in its current state to work with and along comes you trying to give it something to add to it. Once in a while, you try to put something where something else is and GitHub needs help to figure out what you actually want to do, so it pops up a helpful collision message ("Incoming!").

In the case of this PR:
* You had just gotten done [asking GitHub to merge in the raw-string PR](https://github.com/autokey/autokey/pull/1099/files). One of the changes in that PR was the insertion of a new entry in the `CHANGELOG.rst` file just above the **its counterpart** line. GitHub did that with no problems.
* A few minutes later, you [asked GitHub to merge in the bump versions PR](https://github.com/autokey/autokey/pull/1100/files#diff-2c623f3c6a917be56c59d43279244996836262cb1e12d9d0786c9c49eef6b43c). One of the changes in that PR was **also** the insertion of a new entry in the `CHANGELOG.rst` file just above the **its counterpart** line. Since it's the exact spot where the raw-string line was inserted in the previous PR, GitHub is experiencing a merge-conflict, but I prefer to think of it as a collision.

When your new content collided with its existing content, it couldn't figure out whether you'd like to keep just the original content, replace the original content with the new content, keep them both, get rid of them both entirely, or replace them both with something else entirely, so it left it up to you to decide by handing you the entire file with the part it doesn't understand wrapped in markup, like this:
```
<<<<<<< beginning of conflict
new contents
======= separator
old contents
>>>>>>> end of conflict
```

When that happens, GitHub expects you to use the editor that it offers you, find that marked-up conflicted section, use the helpful markup lines to determine which of the conflicted content is old and new, tidy it up so that it looks the way you'd like it to, and, finally, remove GitHub's helpful markup lines. When you're happy with the file, you can click the **Commit merge** button and GitHub will accept it as unconflicted and proceed with the PR merge.

You'd think that GitHub would handle this elegantly by figuring things out on its own, but it cares very much about the surrounding content when it sees any change and is designed to protect your content by leaving such decisions up to you. In this case, the lines of existing content above and below the new line were identical in both PRs. Hence, the collision.

This sort of conflict can be avoided from now on by adding a blank line beneath the bullet-list so that the **its counterpart** line (which is an invisible link-line that's used by a link in the bullet-list, but isn't actually seen in the resulting document) is no longer considered to be part of the bullet-list. That would follow the convention of the other link-lines in that file, which are probably that way for that reason.

Ideally, all of the link-lines should be in a **Links** section at the bottom of the document so they can be easily found for reference and/or editing, but that's a conversation for another time.

Last, but not least, since this PR hasn't been edited yet, there's a **Resolve conflicts** button we can click to fix it. I'll do that now and also add in the blank line mentioned above to prevent this from happening again.

Elliria:
Gosh. The editor didn't work for me at all. It looked great, but when I submitted the change, it left a mess of markup in the file, so I edited it file in my fork instead and that worked beautifully. Both PRs are now mentioned in the `CHANGELOG.rst` file and the blank line beneath them has been added. This PR can now be merged, but you might want to take a look at it first to make sure you're happy with what I did.

PR #1101, opened 2026-04-10
Update pyhamcrest dependency version to 2.1.0

Elliria:
* See [the PyHamcrest PyPI page](https://pypi.org/project/PyHamcrest) to verify that **2.1.0** is the current version.


Elliria:
I'm 100% certain of this one, @josephj11 and it's a quick check of one page linked to above to verify it. The checks are still failing, but that's okay. We'll get to that.

PR #1102, opened 2026-04-10
Update setup.cfg white-space

Elliria:
* Add white-space around some delimiters for readability and consistency with the other delimiters in the file.
* Note that the **cov** lines in the **[tool:pytest]** section are an exception to this and cannot contain white-space since they're command-line options and white-space could lead to unexpected behavior.


Elliria:
I'm confident of this since we already have white-space around delimiters on several lines, but here are some references to verify that it's fine to do this:
* See the examples on the [Python configparser page](https://docs.python.org/3/library/configparser.html) page. You can also read through some or all of it and see that it likens these config files to Microsoft Windows INI files that follow the INI standard, which allows for white-space around delimiters.
* See the examples on the [Setuptools specification page](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html).

PR #1104, opened 2026-04-20
Port Qt UI from PyQt5 to qtpy (PyQt5/PyQt6 compatible)

bgermann:
Replace all `from PyQt5.X import Y` with `from qtpy.X import Y` across the entire Qt UI layer, making the application runnable with either PyQt5 or PyQt6 as the underlying binding.

Key changes:
- Replace PyQt5 imports with qtpy throughout all qtui modules, scripting, interface, and macro modules
- Replace deprecated `exec_()` calls with `exec()` (required by Qt 6)
- Guard `Qt.AA_X11InitThreads` attribute access behind try/except, as it was removed in Qt 6
- Import `Q_CLASSINFO` with a PyQt6 try/except fallback, as it is a SIP-level binding not wrapped by qtpy
- Add `qtpy` to install_requires (setup.py), pip-requirements.txt, setup.cfg tox deps, debian/build_requirements.txt, and debian/control
- Update module availability check in UI_common_functions.py to check for `qtpy` instead of `PyQt5`

### DIFF
--- a/debian/build_requirements.txt
+++ b/debian/build_requirements.txt
@@ -5,6 +5,7 @@ python3-all
 gir1.2-gtk-3.0
 python3-pyqt5.qsci
 python3-pyqt5.qtsvg
+python3-qtpy
 python3-xlib
 devscripts
 python3-pip
--- a/debian/control
+++ b/debian/control
@@ -11,6 +11,7 @@ Build-Depends: debhelper-compat (= 12),
                python3-gi,
                python3-pyqt5.qsci,
                python3-pyqt5.qtsvg,
+               python3-qtpy,
                python3-setuptools,
                python3-xlib,
                python3-pyinotify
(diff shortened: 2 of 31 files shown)


josephj11:
Welcome to the AutoKey community, @bgermann !

Thanks for your contribution.

I am not an an AutoKey/Python developer, so I can't evaluate the merits of this PR, but it looks interesting.

However, there are a few procedural problems with it.

First, all non-emergency PRs need to be based on our develop branch. The master is (hopefully) always released working code.

Second, while this is conceptually one small change, it needs to be split into a number of smaller PRs because reviewing multifile PRs is a nightmare on GitHub. We just went through this with a big technical debt PR and there was a long discussion thread with links to code in it that were almost impossible to follow.

It will also need to be noted in our changelog,rst.

Hopefully, @dlk3 and @Elliria  can take a look at the code before you work on it further and see if it fits in with our project.

Elliria:
This seems like a valid upgrade. If I understand it correctly, **qtpy** is an elegant and painless way to transition from **Qt5** to **Qt6**, which is what's happening in **Ubuntu 22.04** and **Ubuntu 24.04**. Also, the **python3-qtpy** dependency that it adds to AutoKey is available in the **Universe** repository in both Ubuntu releases.

This pull request and several of the other pull requests that are currently open may all be wonderful improvements for AutoKey, but I haven't had a chance to try them out yet. I'm still deeply in the midst of messing with the current AutoKey code to try to get it to reliably test, build, install, and run in **Ubuntu 22.04 LTS** and **Ubuntu 24.04 LTS** under both **Wayland** and **Xorg**. Those current efforts involve several changes to AutoKey that haven't yet been added to the code-base. Changes like those in this pull request (and several others that are currently open) might need to wait until we have a solid base to build them upon. Patience and/or assistance is always welcome.

PR #1106, opened 2026-04-26
Update setup cfg indents

Elliria:
* Remove indent from line 10 by moving it up onto line 9 for readability of the **testpaths** line.
* Fix missing indents in the **addopts** section.


Elliria:
This is a very minor set of changes that improve readability and fix the syntax (required indents) that set the file up for further changes I'd like to make next. This has no negative effect and the file works perfectly in my local tests.

PR #1108, opened 2026-04-26
Use distinct setup.cfg comment types

Elliria:
* Use hash-marks for commenting outtext or documentation.
* Use semi-colons for commenting out (disabling) configuration lines or code.

This improves maintainability by using the default behavior of the [Python configparser](https://docs.python.org/3/library/configparser.html#supported-ini-file-structure), which recognizes both of those characters as prefixes for comments.


Elliria:
This one just makes life easier for devs without changing the behavior of AutoKey in any way.

PR #1110, opened 2026-04-28
Add xfail marker for Wayland compatibility issue

Elliria:
Mark the **test_application_runs_without_errors** function as being expected to fail under Wayland to prevent the function from causing a **SystemExit: 1** error during tests under Wayland on the grounds that the test requires **X11/xhost**, making it incompatible with Wayland.


Elliria:
Note that this is one of the files I have to mark in this way every time in order to do successful local tests and this fix works like a charm. I look forward to this one being added so that I no longer have to do the manual edits each time I want to run tests.

I also snuck in the removal of an errant blank line in the [CHANGELOG.rst](https://github.com/autokey/autokey/blob/develop/CHANGELOG.rst) file that's not visible in the rendered document, but can be seen when looking at its raw code.

PR #1114, opened 2026-04-29
Fix Window import error on scripting tests

Elliria:
Update the final conditional block in the `lib/autokey/scripting/__init__.py` file to handle the Window import error:
* Add the full path to the **Window** import statement in the final clause so that **Window** can be found and imported under **Xorg**.
* Replace the `elif autokey.common.SESSION_TYPE == "x11":` line with the `else:` line in the final clause for broader scope to include not only access by a user in a Xorg session, but also headless access by the CI or local testing utilities, both of which are also reliant on Xorg.
* Remove the **pass** statements that do nothing from both clauses for cleaner code and to prevent confusion.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -102,7 +102,7 @@ Other changes
 - Mark the **test_application_runs_without_errors** function as being expected to fail under Wayland.
 - Sort some of the `setup.cf` sections for readability and maintainability.
 - Clean up local test-build version and entries.
-
+- Handle Window import error on scripting tests in `lib/autokey/scripting/__init__.py`.
 
 .. _its counterpart: https://github.com/autokey/autokey/blob/master/.github/ISSUE_TEMPLATE/config.yml
 
--- a/lib/autokey/scripting/__init__.py
+++ b/lib/autokey/scripting/__init__.py
@@ -46,7 +46,5 @@ elif autokey.common.USED_UI_TYPE == "headless":
 
 if autokey.common.SESSION_TYPE == "wayland":
     from .window_gnome import Window
-    pass
-elif autokey.common.SESSION_TYPE == "x11":
-    from .window import Window
-    pass
+else:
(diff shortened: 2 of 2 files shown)


Elliria:
This fix works beautifully in local tests with `tox` and `pytest`.

Without this fix in place, I get this error:
```
==================================== ERRORS ====================================
_____________ ERROR collecting tests/scripting_api/test_window.py ______________
ImportError while importing test module '/home/elliria/clones/autokey/tests/scripting_api/test_window.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/scripting_api/test_window.py:7: in <module>
    from autokey.scripting import Window
E   ImportError: cannot import name 'Window' from 'autokey.scripting' (/home/elliria/clones/autokey/.tox/coverage/lib/python3.12/site-packages/autokey/scripting/__init__.py)
```
With this fix in place, I get no errors in local tests.

josephj11:
Same won't merge error.

Elliria:
Yep. I'm waiting to fix it until after you merge the other formerly-conflicted pull request in since even if I fix this, the moment you merge the other one in, this one will be conflicted again. That's because I keep making changes to the exact same spot in the [https://github.com/autokey/autokey/blob/develop/CHANGELOG.rst](https://github.com/autokey/autokey/blob/develop/CHANGELOG.rst), but that spot keeps changing every time a pull request is merged in, so the little GitHub bot can't help but become confused, poor little feller.

Elliria:
This merge conflict was even more interesting because two other pull requests had been merged in before it, so there were two lines added to the file:

conflict

The new **Handle window** line is beneath those now, so it's good to go.

PR #1118, opened 2026-05-03
Enable coverage tests

Elliria:
Enable the **--cov-report=term-missing** line in the [setup.cfg](https://github.com/autokey/autokey/blob/develop/setup.cfg) file to enable **coverage** tests. This prevents the **autokey module not found error** when attempts are made to run **coverage** tests and, thus, allowing the tests to run.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -105,6 +105,7 @@ Other changes
 - Update `setup.cfg` comments (fix typo, succinctness, punctuation).
 - Add Wayland-related test-handling to `test_interface.py`.
 - Handle Window import error on scripting tests in `lib/autokey/scripting/__init__.py`.
+- Enable **coverage** tests.
 
 .. _its counterpart: https://github.com/autokey/autokey/blob/master/.github/ISSUE_TEMPLATE/config.yml
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,7 @@ norecursedirs = dist build .tox scripts
 addopts =
   ; --cov-report=annotate:test_coverage_annotated_source
   ; --cov-report=html:test_coverage_report_html
-  ; --cov-report=term-missing
+  --cov-report=term-missing
   ; --cov-report=xml
   # Summarize failed tests at the end, including xfails and skips:
   -r fa


Elliria:
Note that this works beautifully in local tests, making it possible to do **coverage** tests as described in the [CONTRIBUTORS.rst](https://github.com/autokey/autokey/blob/develop/CONTRIBUTORS.rst) file.

Elliria:
Thank you for merging this. It's missing its commit merge (there are only two commits in the [Commits](https://github.com/autokey/autokey/pull/1118/commits) tab above and there should be three since you merged the pull request. As a result, I asked the **GitHub Copilot bot** about it and we had [this chat](https://github.com/copilot/share/425f5100-4940-8455-a912-9408204a41ef) and it turns out that either **Rebase and merge** or **Squash and merge** was used instead of **Create a merge commit** when merging the PR. The fix is in place in the files, including in a local clone that I grabbed as part of sleuthing this, so the pull request is valid. I figured I'd mention this here, though, since the missing merge commit is odd.

Elliria:
Here's the [full GitHub commit history](https://github.com/Elliria/autokey/compare/develop...autokey%3Aautokey%3Adevelop) for posterity's sake in case this is what a **Rebase and merge** or **Squash and merge** looks like. This history does show one unexpected merge that was done to my fork as part of [pull request #1120](https://github.com/autokey/autokey/pull/1120) and seems to have affected this pull request for some reason. That's likely what caused the GitHub confusion by somehow obliterating the merge-commit while actually doing the work of the pull request: [Merge branch 'develop' into update-documentation-mouse-py](https://github.com/Elliria/autokey/commit/9b8ac78133e88a0dd21b734169bf7dd7535da85a)

josephj11:
My head is still spinning after reading that chat. Is the bottom line that I accidentally selected the wrong merge option?

Elliria:
No idea. Those two bots seem to think so, but it doesn't seem like a likely accident since you'd have to deliberately open the drop-down and choose it instead of just clicking the button. I suspect (and said as much to the bot) that it's a residual effect from that outage the other day. It affected more than one of these pull requests, although I suppose that could be explained by a **Squash and merge**. Even so, though, that **Squash and merge** was probably done by GitHub and not by you.

I tried to file a bug report, but the bots outflanked me, so I pasted it all in here because the missing merge commit is concerning and that info could be used for sleuthing if that's needed down the road.

If the first bot is right and the merge happened in a detached HEAD state, then that could mean that the change is a blob that will be garbage-collected at some point. It's not clear whether GitHub does garbage-collection every few days or every couple of weeks, so we'll just have to keep an eye on the files in the affected pull requests to see if any of the changes vanish. Heh.

Just curious, though. Why do we have **Squash and merge** enabled? Does it serve any useful purpose for what we do? We could disable it in the repository and maybe that would prevent the Gremlins from doing their thing.

josephj11:
Unfortunately, that's all beyond my understanding.

If you don't mind a little clutter in the repo, you could easily create a couple of commits/PRs that collide on purpose and do the merge yourself and see what happens. Then you could revert the whole thng if it wasn't real changes.

Elliria:
Everything is a commit. Every pull request is a collection of one or more commits created by us. Every pull-request merge that you trigger is a merge-commit created by GitHub. A **Squash and merge** combines a bunch of pull requests together so that they can all be included under the umbrella of one merge-commit created by GitHub. The bots think that that's what happened. It's possible, but I don't think you did it. I think it happened as a result of some residual mess from the massive outage that GitHub recently had that affected everything.

I had done several pull requests and you didn't necessarily merge them in the order that I did them in. If something went wrong and a merge that you did was interpreted as a **Squash and merge** by GitHub, then it would grab every unmerged pull request up to that point and bundle them all up into the current one as one single commit that's actually part of a totally different pull request. It looks like that's exactly what happened.

Should we disable **Squash and merge** in the settings or do we have a need for it? I'm thinking if we disable it, GitHub won't use it no matter what. Meanwhile, it only happened that one time, so it's either intermittent or part of that outage. I'm also currently doing only one PR at a time in case they still have kinks to work out at their end.

PR #1119, opened 2026-05-03
Clean up and expand .gitignore test and IDE exclusions

Elliria:
Clean up and expand the [.gitignore](https://github.com/autokey/autokey/blob/develop/.gitignore) list that prevents certain directories and files from being accidentally added to the repository:
* Change the **# build artifacts** section header to sentence-case for consistency with the others.
* Move the `build/*` entry to the **# Build artifacts** section.
* Replace the `build/*` entry with `build/` so that not only will the files in the directory be ignored, but the directory, itself, will also be ignored.
* Remove the apostrophe and s from **# IDE's** section header for readability.
* Create a **# Tests** section header for the two test entries.
* Add these new entries to the **# Tests** section so these artifacts will also be ignored:
  ```
  .coverage
  .coverage.*
  htmlcov/
  *.lcov
  .pytest_cache/
  ```
* Move the `.tox/` entry to the new **# Tests** section header.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,12 +4,11 @@
 __pycache__
 *.egg-info
 *.pyc
-.tox/
-build/*
 lib/autokey/qtui/resources/icons/
 
-# build artifacts
+# Build artifacts
 autokey-gnome-extension/autokey-gnome-extension@autokey.shell-extension.zip
+build/
 
 # Environments
 .env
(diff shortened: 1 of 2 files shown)


Elliria:
This is stuff that I'm noticing as I do local builds. It's always a good idea to put anything that can be created temporarily during testing into the `.gitignore` file so that it doesn't end up in the repository and since our testing ability is expanding, we've become a bit messier on the local end (which is a good thing!) and this keeps it from going any further than that.

PR #1120, opened 2026-05-03
Fix typos in mouse documentation

Elliria:
Replace 2 typos of **window** with **screen** in the **click_absolute** function in the [lib/autokey/scripting/mouse.py](https://github.com/Elliria/autokey/blob/master/lib/autokey/scripting/mouse.py) file.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -56,6 +56,7 @@ Important misc changes
 - Bump the AutoKey version to 0.96.1 in the **autokey.spec** file to satisfy part of issue #227.
 - Fix erroneous `window.close` in place of `window.resize_move` in documentation
 - Adds GNOME Window Extension for interacting with Windows on x11/wayland
+- Fix typos in mouse documentation (**window** --> **screen**).
 
 
 Features
--- a/lib/autokey/scripting/mouse.py
+++ b/lib/autokey/scripting/mouse.py
@@ -67,8 +67,8 @@ class Mouse:
 
         Usage: C{mouse.click_absolute(x, y, button)}
 
-        :param x: x-coordinate in pixels, relative to upper left corner of window
-        :param y: y-coordinate in pixels, relative to upper left corner of window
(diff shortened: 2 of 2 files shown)


Elliria:
For some reason, GitHub changed **line 231** (the last line of text in the file) and then flagged it as being different in the file comparison. I asked Co-pilot about it inside the GitHub file comparison interface and it says it's a harmless and beneficial change that GitHub just did, so it's nothing to worry about and I guess we should be grateful to GitHub for making the fix. Here's the bot's response:

> Copilot finished thinking
> Copilot said: I'll load the PR understanding ability and

> I'll load the PR understanding ability and then examine the changes in that PR to help you understand what's happening with line 231.Now let me fetch the PR details to understand what's happening with line 231:The red minus sign with a red circle (❌) indicates that line 231 had a "No newline at end of file" issue that you fixed in your commit.

> Here's what happened:

> Line 231 in lib/autokey/scripting/mouse.py originally had:
> Code

> self.interface.scroll_up(number)\ No newline at end of file

> The \ No newline at end of file marker in the diff means the file was missing a newline character at the very end.

> Your fix added the newline, so it now ends with:
> Code

> self.interface.scroll_up(number)

> The red indicator is GitHub's way of flagging that this line changed—even though you didn't explicitly edit the logic on line 231. It's flagged because you changed the whitespace/formatting at the end of the file, which is generally good practice (files should end with a newline character).

> This is a minor, positive fix that makes the code follow proper formatting standards.

Elliria:
~~Heads up that the [actions will need to be run manually](https://github.com/autokey/autokey.github.io#run-the-actions-after-documentation-changes) in the [AutoKey Documentation's repository](https://github.com/autokey/autokey.github.io) after this pull request is merged.~~

It turns out that that won't be necessary after all since this pull request was done against the **develop** branch instead of the **master** branch, so, it won't take effect in the actual documentation until the next release of AutoKey goes out.

PR #1122, opened 2026-05-05
Add enhancements to setup cfg

Elliria:
Add measurement and timeout and blame settings to `setup.cfg` for more robust testing:
* Add the **--cov=lib/autokey** line to the **pytest** section to tell the pytest-cov plugin to measure and report code coverage specifically for the autokey libraries (basically, you get more results) and to ensure that running **pytest** in a terminal outside of **tox** will work.
* Add a missing blank line between sections for consistency with the rest of the document.
* Add **pytest-timeout** as a **[testenv]** dependency to handle timing issues.
* Add the **[coverage:html]** and **[coverage:run]** sections to the bottom of the setup.cfg file to satisfy the .[coveragerc](https://github.com/autokey/autokey/blob/develop/.coveragerc) file which contains the **show_contexts = true** line that tells **coverage** to track which test affects which line of code: 
* Add the **--cov-context=test** option to the **commands** entry in the **[testenv:coverage]** section as an on-switch for measurement during execution and to provide extra data in the testing HTML report.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -109,6 +109,7 @@ Other changes
 - Handle Window import error on scripting tests in `lib/autokey/scripting/__init__.py`.
 - Clean up and expand test and IDE exclusions in the `.gitignore` file.
 - Enable **coverage** tests.
+- Add measurement and timeout and blame settings to `setup.cfg` for more robust testing. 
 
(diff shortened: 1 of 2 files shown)


Elliria:
This is the last major change to this file that makes testing better in several ways. There will be one final small (but vital) change to this file as part of the **green check-marks** pull request that will be happening soon. I didn't want to do it as part of this since the **green check-marks** pull request will involve changes to several files all related to one pesky dependency, so I'd rather lump those together.

PR #1123, opened 2026-05-08
Sort apt requirements txt

Elliria:
Update the [apt-requirements.txt](https://github.com/autokey/autokey/blob/master/apt-requirements.txt) file:
* Remove the trailing white-space from the end of line 112 (incidental fix I noticed while updating the file).
* Sort the packages in alphabetically for easier comparison on future changes.


Elliria:
Sorting the entries has no effect on AutoKey since the entire file is read before it's used. I'm about to make several changes to the file, a couple of which are replacements and the rest of which are additions, and having the packages sorted will make the comparison a much gentler experience.

PR #1125, opened 2026-05-08
Setup.cfg cross-version harmony

Elliria:
Update the [Setup.cfg](https://github.com/autokey/autokey/blob/develop/setup.cfg) file:
* Pin the **PyGObject** version in the **[testenv]** section to version **3.50.0** or lower to ensure cross-version harmony in both **Ubuntu 22.04** and **Ubuntu 24.04+**.


Elliria:
This is the final change to this file for the foreseeable future and lays the groundwork for getting those green check-marks back. As you can see, we still fail the tests, but that's only because this file can't fix it on its own. Others will follow.

PR #1126, opened 2026-05-09
Update apt-requirements.txt dependencies

Elliria:
Update dependencies in the [apt-requirements.txt](https://github.com/autokey/autokey/blob/develop/apt-requirements.txt) file:
* Add **gir1.2-ayatanaappindicator3-0.1** to ensure tray-icon support on newer distributions (**Ubuntu 24.04+**) that have transitioned away from the legacy **appindicator** package.
* Add **gobject-introspection** to provide the binary tools needed to process the interface definitions for **autokey-gtk**.
* Replace **libcairo-dev** with **libcairo2-dev** as an upgrade for the newer Python versions that's supported by both **Ubuntu 22.04 LTS** and **Ubuntu 24.04 LTS**.
* Add **meson** to create a build plan using the library paths and dependencies provided by **pkg-config**.
* Add **ninja-build** to provide the engine that executes the build plan's commands when requested by **pip**.
* Add **pkg-config** to help the build system find D-Bus and Wayland library paths and dependencies.
* Replace the **libpython3-dev** package with the **python3-dev** meta-package to ensures that headers match Python version used in GitHub Actions and local tests for compiling C-extensions.
* Add **python3-evdev** for low-level input monitoring on Wayland.
* Add **python3-tk** to resolve an **ImportError** during the unit-testing phase of the build.


Elliria:
This file is also now finished for the foreseeable future. We're still failing the tests, as expected, but this takes us another step in the direction of solving that and giving us our green check-marks back. More pieces to come later in other files.

PR #1127, opened 2026-05-11
Fix triggered abbreviation for immediate scripts

### LINKED ISSUE #948: get_triggered_abbreviation() Not Working

Xorg

- [X] This is not a question.

None

- [ ] autokey triggers
- [X] autokey-gtk
- [ ] autokey-qt
- [ ] beta
- [X] bug
- [ ] critical
- [ ] development
- [ ] documentation
- [ ] enhancement
- [x] phrase expansion
- [X] scripting
- [ ] technical debt
- [ ] user interface

_No response_

Xubuntu 22.04

GTK

0.96.0

*.deb package

I was working on an Autokey script and ran into trouble trying to use engine.get_triggered_abbreviation(), since I couldn't get any data out of either part of the tuple that it returns. So I created a simple script with just the following line of code and ran Autokey in verbose mode.

`abbreviation, trigger_character = engine.get_triggered_abbreviation()`

When I opened the text editor, typed my abbreviation, and hit enter, what I saw was the following:

```
DEBUG - autokey.service - Triggered a Script by an abbreviation. 
Setting it for engine.get_triggered_abbreviation(). abbreviation='', trigger=''
```

To sanity-check myself, I then added this second line to my script:

`print("abbreviation is" + abbreviation)`

Sure enough, in the terminal, I get

```
DEBUG - autokey.service - Triggered a Script by an abbreviation.
Setting it for engine.get_triggered_abbreviation(). abbreviation='', trigger=''
abbreviation is 
```

It doesn't appear that abbreviation and trigger are getting set. In the Set Abbreviations window, I have all of the boxes unchecked except "Omit Trigger Character". That might explain why trigger_character is not set, but why can't I get the triggered abbreviation?

Always

1. Create a script that contains the line 'abbreviation, trigger_character = engine.get_triggered_abbreviation()`
2. Assign an abbreviation to that script
3. Run Autokey in verbose mode
4. Open a text editor, type in the abbreviation, and hit enter
5. Note the output in the terminal

The "abbreviation" variable should have stored the abbreviation I used for the phrase expansion.

### What actually happened?

The "abbreviation" variable is empty

### Do you have screenshots?

_No response_

```bash
DEBUG - autokey.service - Triggered a Script by an abbreviation. 
Setting it for engine.get_triggered_abbreviation(). abbreviation='', trigger=''
```

### Anything else?

_No response_

ilostmyhandle:
## Summary
- Fix engine.get_triggered_abbreviation() for scripts triggered by immediate abbreviations
- Derive the typed abbreviation from the script's abbreviation matching helpers instead of slicing with an empty trigger character
- Add regression coverage for normal, prefixed, immediate, and case-insensitive abbreviation triggers

Fixes #948.

## Verification
- python -m py_compile lib\autokey\service.py tests\test_service.py
- Focused local logic check for immediate abbreviation extraction

I attempted to run pytest locally, but this Windows environment was missing pytest/PyHamcrest/packaging and dependency installation timed out. The added test is intended to cover the reported regression directly.

Disclosure: this PR was prepared with AI assistance and reviewed before submission.

### DIFF
--- a/lib/autokey/service.py
+++ b/lib/autokey/service.py
@@ -517,7 +517,7 @@ class ScriptRunner:
         backspaces, trigger_character = script.process_buffer(buffer)
         self.mediator.send_backspace(backspaces)
 
-        self._set_triggered_abbreviation(scope, buffer, trigger_character)
+        self._set_triggered_abbreviation(scope, script, buffer, trigger_character)
         if script.path is not None:
             # Overwrite __file__ to contain the path to the user script instead of the path to this service.py file.
             scope["__file__"] = script.path
@@ -577,11 +577,16 @@ class ScriptRunner:
         return script_code, script_name
 
     @staticmethod
-    def _set_triggered_abbreviation(scope: dict, buffer: str, trigger_character: str):
+    def _set_triggered_abbreviation(
+            scope: dict,
+            script: autokey.model.script.Script,
+            buffer: str,
+            trigger_character: str):
         """Provide the triggered abbreviation to the executed script, if any"""
         engine = scope["engine"]  # type: autokey.scripting.Engine
         if buffer:
-            triggered_abbreviation = buffer[:-len(trigger_character)]
+            configured_abbreviation = script._get_trigger_abbreviation(buffer)
+            _, triggered_abbreviation, _ = script._partition_input(buffer, configured_abbreviation)
 
             logger.debug(
                 "Triggered a Script by an abbreviation. Setting it for engine.get_triggered_abbreviation(). "
--- a/tests/test_service.py
+++ b/tests/test_service.py
@@ -1,5 +1,4 @@
 # Copyright (C) 2021 BlueDrink9
-
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
@@ -27,10 +26,13 @@ from tests.engine_helpers import *
 import tests.helpers_for_tests as testhelpers
 
 import autokey.model.folder
+import autokey.model.helpers
+import autokey.model.script
 from autokey.configmanager import configmanager
 from autokey.configmanager.configmanager import ConfigManager
 from autokey.service import Service
 from autokey.service import PhraseRunner
+from autokey.service import ScriptRunner
 import autokey.service
 from autokey.scripting import Engine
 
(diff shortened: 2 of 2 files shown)


ilostmyhandle:
Hi, just checking whether this looks like the right direction for #948. I kept the fix scoped to the abbreviation reported by the script's own matching helpers and added regression coverage for immediate abbreviations. Happy to adjust if you'd prefer the logic live somewhere else.

josephj11:
Welcome to the AutoKey community @ilostmyhandle !

Thanks for his PR. It looks interesting, but it's too far in the internals for me to evaluate. I'm not an AutoKey/Python developer.

I'm not sure we currently have a developer who can evaluate this. The those who were here recently have been pretty silent and I think this might be too deep for our junior developer to be sure about.

@dlk3, Can you take a look at this?

First, though:
All new PRs must be based on our develop branch. We are trying to keep the master as close to the current release as possible with only emergency changes. We hope to be able to release 0.97.0  this year and will do so at our earliest opportunity once it is completed and tested. We'll start with a beta release.

ilostmyhandle:
Thanks for the welcome and the guidance. I rebased the branch onto current `develop` and retargeted this PR accordingly. I also resolved the test-file overlap from `develop`; the local syntax check still passes with `python -m py_compile lib/autokey/service.py tests/test_service.py`.

dlk3:
@josephj11  I am unable to evaluate this PR because I cannot reproduce the original problem from #948 in my Wayland/Gnome environment using [my AutoKey for Wayland fork](https://github.com/dlk3/autokey-wayland).  The "engine.get_triggered_abbreviation" method referenced in #948 appears to work fine for me.

Per the comments in #948, I created the following test script and ran it using the abbreviation ";test" with space/enter as the trigger characters.

    logger = __import__("autokey.logger").logger.get_logger(__name__ + ".test.script")
    abbreviation, trigger_character = engine.get_triggered_abbreviation()
    logger.info(f'abbreviation = {abbreviation}, trigger_character = {trigger_character}')

My script wrote the following message in the log, which appears to me to qualify as "working as expected."

    2026-05-12 00:10:58,454 INFO - autokey.service.test.script - abbreviation = ;test, trigger_character =

ilostmyhandle:
Thanks for checking this. That result makes sense to me: the space/enter-triggered path already works with the current code.

The case this PR is trying to cover is the immediate-trigger path, where the script abbreviation is configured with "Trigger immediately" and there is no ending character. In that path `trigger_character == ""`, so the current code does:

```python
triggered_abbreviation = buffer[:-len(trigger_character)]
```

Because `len("") == 0`, that becomes `buffer[:-0]`, which is the same as `buffer[:0]`, so the abbreviation passed to `engine.get_triggered_abbreviation()` becomes an empty string.

A closer repro would be the same script, but with the `;test` abbreviation set to trigger immediately instead of using space/enter as the trigger character. The regression test in this PR covers both cases:

- `("abbr ", "abbr", False, False, ("abbr", " "))` keeps the already-working normal trigger behavior
- `("abbr", "abbr", True, False, ("abbr", ""))` covers the immediate-trigger behavior this PR changes

So I do not expect your space/enter test to fail before this PR; the intended difference should show up when the abbreviation triggers immediately.

josephj11:
@dlk3. Thanks for looking. We thought you might have left us. I invited you to our developers room on Gitter a while ago and never heard back. I will invite you again if you are interested.

@ilostmyhandle Your last response is very difficult to read. You need to use the Preview tab before posting anything other than plain text. GitHub supports markdown, so you can get things to display as intended without a lot of effort. You can still edit your post.Looks like you wanter tripple single quotes insteadof braces.

ilostmyhandle:
Thanks, Joseph. You're right. I posted that through a CLI command and accidentally sent the API JSON wrapper instead of the rendered Markdown. I've edited the comment so it should be readable now.

dlk3:
@ilostmyhandle Thanks.  I retested without a trigger and was able to reproduce the bug.

@josephj11 This PR does correct the problem in #948.  I'd recommend it be merged.

dlk3:
@josephj11 I'm not big on chat rooms, so I haven't rejoined the AutoKey room on Gitter.  I don't generally monitor chats anyway, so they're not a great way to reach me.  The best way to get my attention is to use the "mention" (@dlk3) tag feature here.

josephj11:
@ilostmyhandle Thanks for the great submission. It will be merged sometime soon. We have a lot of internal PRs going through ATM.

If you're not already familiar with our resources, be sure to check out our [wiki](https://github.com/autokey/autokey/wiki) and join us on [Gitter](https://gitter.im/autokey/autokey) to discuss anything AutoKey-related. Gitter tends to be pretty quiet, but I check it almost every day and there are a number of other users who show up occasionally.

Elliria:
I don't fully understand all of the code, but from what I do understand, it looks solid and it will solve a pesky problem.

PR #1129, opened 2026-05-13
Harden Wayland preflight edge cases

mt211211:
## Summary
- Treat missing or non-Wayland XDG_SESSION_TYPE as a no-op for the Wayland preflight check.
- Avoid KeyError when an unsupported Wayland session has no XDG_SESSION_DESKTOP value.
- Use non-interactive-safe user lookup and report a missing input group instead of crashing.
- Add focused coverage for the new preflight edge cases.

## Why
This is a focused follow-up for #87 against the current develop branch. It improves the existing Wayland path without replacing the broader implementation work, especially for run-from-source, CI, and partially configured sessions where startup should fail predictably or continue when not actually under Wayland.

## Testing
- git diff --check
- python -m py_compile lib/autokey/wayland_checks.py tests/test_wayland_checks.py
- manual wayland_checks smoke script on Windows with a stubbed Linux grp module

Not run locally: pytest tests/test_wayland_checks.py --no-cov, because pytest is not installed in this environment.

/claim #87

### DIFF
--- a/lib/autokey/wayland_checks.py
+++ b/lib/autokey/wayland_checks.py
@@ -2,6 +2,7 @@
 #  desktop environment and that the prereqs that desktop environment needs are 
 #  in place.
 
+import getpass
 import grp
 import os
 import subprocess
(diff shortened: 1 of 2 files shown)


josephj11:
@mt211211 Welcome to the AutoKey community!

Thanks for this PR. It looks interesting, but the details are mostly over my head. I'm not an AutoKey/Python developer.

I reviewed the code and one thing popped out at me. When testing for the user group, it looks like you added an allowance for root (UID 0). This doesn't seem necessary or a good idea because AutoKey is designed as an end user tool and probably should never be run as root. AFAIK, we have never tested it running as root and have no idea what might break or what else it might break.

I'll leave the rest for our other developers to review.

@dlk3 Do you see these new PRs when they get submitted or should I tag you on them when they require more technical expertise?

mt211211:
I appreciate the feedback @josephj11 If there's additional feedback from the wider team as well, I'll try to adjust as required. Feel free to let me know 👍

dlk3:
@josephj11 This PR looks fine to me.  The root uid check is inconsequential, in my opinion.  I put it in because I don't see any reason why AutoKey couldn't/shouldn't be usable by the root uid, as unlikely as that might be.  But it could just as easily be dropped.

I'm not reading every PR that comes through.  If you want my reaction to something, it would be best to tag me.

Elliria:
The **sudo** command is innocent and is needed for Wayland. It adds your user account to the **input** group and the **sudo** part just lets the system knows that you really mean it, since nobody but you should be adding users to a group.

josephj11:
@mt211211 Just merged it. Hope you'll have more to contribute!

Issue #961, opened 2024-11-05
Fedora 41 compatibility — imghdr removed with python 3.13.


markallasread:
### AutoKey is a Xorg application and will not function in a Wayland session. Do you use Xorg (X11) or Wayland?

Xorg

### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

None

### Choose one or more terms that describe this issue:

- [ ] autokey triggers
- [X] autokey-gtk
- [X] autokey-qt
- [ ] beta
- [ ] bug
- [ ] critical
- [ ] development
- [ ] documentation
- [X] enhancement
- [X] installation/configuration
- [ ] phrase expansion
- [ ] scripting
- [X] technical debt
- [ ] user interface

### Other terms that describe this issue if not provided above:

fedora python fc41 imghdr

### Which Linux distribution did you use?

Moving to fedora 41 introduced a fresh error for me: ModuleNotFoundError: No module named 'imghdr'

Python 3.13 no longer includes imghdr which at least in my context, prevented autokey-gtk / autokey-qt from starting. I suspect any autokey install on a distribution that's moving towards python 3.13 will want to implement this (with some research, there may be better ways. I'm a novice). 

For my own setup this worked:

I've swapped out the imghdr-reliant code for imagesize — specifically within: 
/usr/lib/python3.13/site-packages/autokey/scripting/highlevel.py:

Lines 66=>78:

I switched out:

```
def get_png_dim(filepath: str) -> int:
    """
    Usage: C{get_png_dim(filepath:str) -> (int)}

    Finds the dimension of a PNG.
    @param filepath: file path of the PNG.
    @returns: (width, height).
    @raise Exception: Raised if the file is not a png
    """
    if not imghdr.what(filepath) == 'png':
        raise Exception("not PNG")
    head = open(filepath, 'rb').read(24)
    return struct.unpack('!II', head[16:24])
```

to:
```
def get_png_dim(filepath: str) -> tuple[int, int]:
    """
    Usage: get_png_dim(filepath:str) -> (width: int, height: int)
    Finds the dimension of a PNG.
    @param filepath: file path of the PNG.
    @returns: tuple of (width, height).
    @raise Exception: Raised if the file is not a png
    """
    # Check if file exists
    if not os.path.isfile(filepath):
        raise Exception("File does not exist")
        
    # Check if file is PNG using python-magic
[...]

PR #1076, opened 2026-02-17
Fixes needed to make the develop branch useable on Wayland and X11 systems

dlk3:
This PR makes the develop branch functional on Gnome/Wayland and on X11.  See the [CHANGELOG.rst](https://github.com/dlk3/autokey/blob/test-merge-into-develop/CHANGELOG.rst) it contains for the changes I made to the develop branch to get things running.

I have tested this version of AutoKey on my main Fedora 43 (Wayland) workstation, on a Ubuntu 24.04 (Wayland) VM, on a Kubuntu 24.10 (X11) VM, and on a Fedora 42 laptop running the XFCE desktop over X11.  I tested by running my personal collection of phrases and scripts.

The focus of this PR is on enabling Wayland, but it still runs fine on X11 systems too.  Under Wayland, this version still only works in the Gnome desktop environment.  For an explanation of why this is the case, see [this comment](https://github.com/autokey/autokey/issues/1042#issuecomment-3905240349), where I discussed this in the context of enabling it to also run under KDE on Wayland.

I have replaced the file called <code>wayland_install.md</code> with this [wiki page](https://github.com/autokey/autokey/wiki/AutoKey-develop-Branch-Installation-Instructions) that has the current instructions for installing this version under Wayland.  For X11 system testing, I hope/expect that you can use whatever manual installation procedure you've used in the past.  In my case, I have not done any manual installations on X11 systems. Instead, I have been using the DEB or RPM packages I've built, which are also mentioned on that same [wiki page](https://github.com/autokey/autokey/wiki/AutoKey-develop-Branch-Installation-Instructions), to do X11 installs.  I'd appreciate information on any problems you see doing manual installations under X11.

PR #1133, opened 2026-05-20
fix: remove imghdr dependency from PNG size helper

MINBBBIGcode:
Refs #961.

Summary:
- Remove the deprecated Python 3.13-incompatible imghdr dependency from the PNG size helper.
- Parse PNG dimensions directly from the PNG header.
- Add focused scripting API regression tests.

Validation:
- python -m pytest -o addopts= tests\scripting_api\test_highlevel.py -q
- python -m py_compile lib\autokey\scripting\highlevel.py tests\scripting_api\test_highlevel.py
- rg -n "imghdr" lib tests setup.py pip-requirements.txt apt-requirements.txt

### DIFF
--- a/lib/autokey/scripting/highlevel.py
+++ b/lib/autokey/scripting/highlevel.py
@@ -6,7 +6,6 @@ import time
 import os
 import subprocess
 import tempfile
-import imghdr
 import struct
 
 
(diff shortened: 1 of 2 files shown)


josephj11:
@MINBBBIGcode Welcome to the AutoKey community! This one looks interesting. AFAIK, we solved it another way, but this might be simpler.

However, all new PRs must be based on the develop branch and you need to add the appropriate entry in CHANGELOG.rst.

You may want to wait until one of our devs looks at this to see if it really is better before doing extra work.

Elliria:
Yep. This was fixed in [pull request #1076](https://github.com/autokey/autokey/pull/1076) in the development code and will be in the next release, so it doesn't need to be done over on the **develop** branch, but can just be closed. I removed the **help wanted** tag from that issue so that this isn't likely to happen again.

PR #1145, opened 2026-05-26
Update predefined_user_script README.txt

Elliria:
Update the grammar, punctuation, formatting, and wording for clarity and readability.


josephj11:
I don't think making those few items hyphenated makes sense, but it's not enough to stop the PR.

PR #1147, opened 2026-05-31
Add referral section to issue-report form

Elliria:
Add the **referral** section to the bottom of the `bug.yml` issue-report form so users can share how they found the **AutoKey** project.


Elliria:
Note that once this is merged, it will get cherry-picked over to the **develop** branch.

PR #1148, opened 2026-06-01
Cherry-pick update of bug.yml from master

Elliria:
Cherry-pick the updated `bug.yml` file from the **master** branch that added the referral section to the issue-report form.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -239,7 +239,18 @@ body:
       render: bash
     validations:
       required: false
-  
+
+  # HOW DID YOU FIND THE AUTOKEY PROJECT?
+  - type: textarea
+    id: referral
+    attributes:
+      label: How did you find the AutoKey project?
+      description: >
+        If this is your first AutoKey pull request, please
+        describe how you found the AutoKey project.
+    validations:
+      required: false
+
   # ANYTHING ELSE?
   - type: textarea
     id: notes
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -115,6 +115,7 @@ Other changes
 - Pin **PyGObject** version in `setup.cfg` to ensure cross-version harmony.
 - Update the dependencies in the `apt-requirements.txt` file.
 - Fix typos and formatting and wording in the ``_README.txt` file.
+- Cherry-pick the the `bug.yml` file from **master** to get the update that added a referral question to the issue-report form.
 
 .. _its counterpart: https://github.com/autokey/autokey/blob/master/.github/ISSUE_TEMPLATE/config.yml
 .. _PEP 440: https://peps.python.org/pep-0440/


Elliria:
Notice that this did put the branches out of sync:
* On the **master** branch, the update of the `bug.yml` file is a commit, the update of the `CHANGELOG.rst` file is a commit, and the PR is a commit. That's three right there.
* On the **develop** branch, the cherry-pick is a commit, the update of the **CHANGELOG.rst** file is a commit, and the PR is a commit. That's three more.

We don't care about the three on **develop** because we expect **develop** to be ahead (and it already was). We do, however, care about the three on **master** because that's messy. We do know  that they're perfectly innocent and don't need further intervention when it comes time to do the next release. As long as we're aware of that with any changes that have to happen on **master** and as long as we always bring a change immediately over to **develop**, it won't matter how many commits **master** is ahead of **develop** and we can safely ignore those.

My question, at this point, is whether we should try to keep it at the lowest possible number (the current three plus any version bumps that are needed before the next release) and do the documentation fixes on **develop** or go for it and do the documentation fixes on **master**? If it helps any, as long as we always, always, always make sure to either cherry-pick or copy a change from **master** to **develop** so that there's no doubt that that was done and we don't have to wade through them all later to verify it.

Initially, I'd say that once there's a release, **master** should be completely frozen until it's time to wipe it and replace it with **beta**, but there's no elegant way to do that. After all, **master** contains the **GitHub Actions** workflows and some vital configuration files, all of which need to be updated when versions are bumped.

Thoughts?

josephj11:
@Elliria My initial take is to minimze this sort of thing so that if we make a mistake/oversight and create a "real" sync problem, it isn't buried among a ton of "innocent" ones.

Whatever we do shouldn't make project/repository management any more difficult than necessary.

Elliria:
Agreed. And perhaps from version **0.97.0** forward, we could simply make **master** completely hands-off for the duration of that release. When you think about it, if that was our policy, it wouldn't matter that its versions didn't keep up, because there would be no activity that would need them. All activity would be happening on **develop** and eventually **beta**, both of which have their own configuration files and GitHub actions, which can be kept up-to-date.

When it comes to fixing the current documentation errors, do you think it would be better to do those on **develop** instead, with the understanding being that the official documentation will always be in whatever state it happens to be in at release time?

The way I'm picturing that, if all documentation updates are done on **develop**, we could point users at our guides that show how to build the documentation locally. Since the guides already include steps for various ways to do the clones, we could suggest that they choose to clone the **develop** branch when doing the build so they can see the updates.

Elliria:
I've found a solution to our branch mess with a nifty little command known as a **graph sync merge**, which tells Git that we **know** there's a deviation, but we'd like Git to consider the branches to be synchronized **anyway** even though Git and we **know** they're not. It's so very tiny, yet such a **very** powerful command.

When a **graph sync merge** is used to tell Git that we know about the deviation (our bad!), but we're okay with it (no worries!) and would like it to shove our mess under the rug and make like the branches are synced anyway (what mess?), GitHub will consider the conflicts "resolved" and the pull request will show as ready to merge.

If you look at the **Files changed** tab in the pull request, it will only show the **graph sync merge** commit and its accompanying `CHANGELOG.rst` commit. Once the pull request is approved and merged in, both of those commits will be made on the **develop** branch and the **develop** branch will now have a **behind** count of **0** while its **ahead** count of welcome additions or changes (that are currently at well over six hundred) will still safely be there untouched.

And, for our peace of mind, the actual commits that created the mess (in this case, those 3 commits that are currently ahead on **master** that happened when we updated the `bug.yml` file) will all still remain completely intact in the Git (and GitHub) history, but we just won't be **told** about them any more and Git won't get worked up about them as being aheads or behinds, either. All we've done is tell Git that, from **this** moment forward, we'd just like for **master** to not be **considered** to be **ahead** of **develop** despite the fact that Git and we **know** that they are (because those are strictly **master** commits and we don't want them on **develop**).

This is **not** a rebase, which would be an aggressive thing to do to a production branch and is, thus, frowned upon for this type of purpose. And since I know you're going to ask (I did!), the only way that this could **ever** come back to bite us is if we merged **develop** into **master**, which is not a part of our [release cycle](https://github.com/autokey/autokey/wiki/Release-Cycle) and is something we would **never** do. Instead, when it comes time for a new release, we use a hard reset to **obliterate** master by overwriting it with **beta**. As a result, we're completely safe.

This is so incredibly glorious! It allows us to not only do things like the `bug.yml` update that we just did in this PR, but also version updates, workflow edits, documentation fixes/edits/additions, or even hotfixes, without **ever** worrying about desychronizing those branches. All of those tasks would normally be **offenders** that would up the **behind** count on **develop**, but with **this** nifty trick, each time we do **any** of those things, we can tidy it up immediately afterward since we'll know for sure that we've already safely cherry-picked the change over to **develop** and would like to no longer have to **think** about that excess cruft from **master**. It's simply breathtaking!

I've already successfully done it locally (wheee!). Would you like me to do it in a PR so you can see it, too?

P.S. I realize that I'm rather excited about this one, but it's truly geekly heaven at its best. 😁❤️

josephj11:
This is a bit confusing to me, especially since i'm not down in the trenches with you. What I understand this to mean is that it only affects the relationship between develop and master. It would not, for instance, cover up some change by two separate PRs to the same code, causing a real conflict that needs resolution. If that's the case, great.

But, if it's so great, why isn't it the way things work by default? There must be a potential downside somewhere.

We discussed the documentation before and I wanted it to reflect the current release in production. If there's a new feature in develop, it should not be documented as in production as that would mislead or confuse users. So if we can release documentation at the same time as a release, that would also be great.

OFC, if we found errors or ommissions in the current documentation, it would be fine to fix them in the best way when they were discovered.

Elliria:
There is **one** potential down-side to it that we're immune to, as I pointed out above:

>And since I know you're going to ask (I did!), the only way that this could **ever** come back to bite us is if we merged **develop** into **master**, which is not a part of our [release cycle](https://github.com/autokey/autokey/wiki/Release-Cycle) and is something we would **never** do. Instead, when it comes time for a new release, we use a hard reset to **obliterate** master by overwriting it with **beta**. As a result, we're completely safe.

Also, it **is** the way things work by default. In other words, this magical command is a default offering from Git. And the reason that it isn't **the** default is that we usually **want** to know when branches deviate in **either** direction -- ahead **or** behind. In the case of necessary changes on **master** that we **know** we don't want on **develop**, knowing about those changes is a **problem** that will continually haunt us until the next release by being a big, ugly **behind** on **develop** and on **any** feature-branch that's **based** on **develop**, including **beta**. If you have a **bunch** of the wrong kinds of  deviations (as we did before), you end up with a huge laundry-list of those problematic deviating commits (as we did before) to chase down and sort out before the next release. This gets rid of the laundry-list.

When you run the `git merge -s ours master` command on **any** branch other than **master**, you're saying to Git:

<mark>"Please ignore **any** commits that happened on **master** before **this** commit that **aren't** on **this** branch.</mark>

It's a **get out of jail free** card that lets you deliberately make **master** deviate without penalty by making the other branch ignore the deviation. In other words, it lets you get stuff done.

And a silly little **TLDR** to go along with it:
* **You:** "Git?"
* **Git:** "Yeah?"
* **You:** "Uhm. hey... You weren't looking when I made those three messes over there on **master**, were you?"
* **Git:** "Yeah. Yeah, I actually was. Those are nasty. I'm gonna have to hold them against you in the **behind-count**."
* **You:** "How about I slip you a little bit of this **Graph Sync Merge** and you forget all about those messes from now on?"
* **Git:** "What messes?"
* **You:** "Thanks, Git."

Elliria:
I forgot to reply to this part:

> We discussed the documentation before and I wanted it to reflect the current release in production. If there's a new feature in develop, it should not be documented as in production as that would mislead or confuse users. So if we can release documentation at the same time as a release, that would also be great.

Understood, and not a problem. All of that can be done.

At this point, we don't have a plan for how to separate the documentation for each release. We just have the [legacy documentation](https://autokey.github.io/autokey/index.html) and the [current documentation](https://autokey.github.io/index.html). We didn't version them when we created locations for them, so, adding new versions will be an issue. In other words, when **0.97.0** is released, the [current documentation](https://autokey.github.io/index.html) will then be for that release.

That would leave the documentation for **0.96.0** unaccounted for and no longer available. The [build-pdf.md](https://github.com/autokey/autokey.github.io/blob/master/guides/build-pdf.md) guide can can be used to handle that issue by creating a PDF file of each [release](https://github.com/autokey/autokey/releases) and adding it to its downloadable artifacts.

If that's a welcome change, we can automate that in the [build.yml](https://github.com/autokey/autokey/blob/develop/.github/workflows/build.yml) file so that that artifact is created and added as artifacts automatically when that **GitHub Action** is run as part of our standard workflow. If there are any updates to the documentation after the fact, we can use the guide to manually generate a fresh version of the PDF and easily replace the artifact with the updated one (or add it - whichever is preferred).

Last, but not least, we can put a small note on every page of all of the documentation that lets users know that **this** documentation is for **release X** and if they're looking for documentation for **another** release, please click **here**. I already have code that automatically does that in my local builds.

josephj11:
@Elliria I don't like "big, ugly behinds"! LOL

I like the note per page - maybe a footnote, not in your face.

Should we be versioning our docs? if not, why not?

How do I know these "artifacts" exist  and that I might need to look at them?

Maybe README.rst could point to the docs/artifacts and how to find the version to match what the user has installed.

Elliria:
Same here on the big, ugly behinds. Would you like me to get rid of the three we currently have?

The note needs to be in your face. If it's in the footer, users may never notice it and then become frustrated when the information they find in the documentation doesn't seem accurate and they don't understand why. It would be in the form of a prologue.

We probably should have versioned our documentation to begin with, but we didn't. I'll start a discussion on it since we need to explore it.

To see the downloadable artifacts for any release, go to AutoKey's [Releases](https://github.com/autokey/autokey/releases) page and scroll down to the bottom of any release. They'll be in the **Assets** section. If, for example, you scroll down to the **Assets** section for version **0.96.0**, you'll see links to these five files:
```text
autokey-common_0.96.0_all.deb
autokey-gtk_0.96.0_all.deb
autokey-qt_0.96.0_all.deb
Source code (zip)
Source code (tar.gz) 
```
If I were to use our [build-pdf.md](https://github.com/autokey/autokey.github.io/blob/master/guides/build-pdf.md) guide to create a PDF archive of **AutoKey 0.96.0** and attach that downloadable artifact to the **Assets** section, you'd see the **autokey-docs-0.96.0.pdf** file in that list and could download it.

Elliria:
I forgot to add that since we already have a [Documentation](https://github.com/autokey/autokey/blob/develop/README.rst#documentation) section in the `README.rst` file, there's already a note (with links) about the different versions and I agree that if we add any downloadable artifacts, we should mention them there.

josephj11:
I suppose it could be in the page  header.

It would be much better if we could connect the documentation to each version, even if it's a zip or tarball. That way, if someone hasn't or can't upgrade, they can still get accurate documentation for the version they have.

IMO, this isn't a super high proiority if it takes a lot of work.

Elliria:
We can definitely connect it to each version in any or all of those formats. The older versions will be certain amounts of work since we didn't use Sphinx back then, but I can poke around to see how difficult it would be to just scoop up what exists and shove it into those types of archives.

josephj11:
@Elliria That makes sense for 0.95.10 and 0.96.0. I see no reason to go back further.

Elliria:
Alas, we're [stuck going back further](https://github.com/autokey/autokey/discussions/1152).

Elliria:
By the way, did you want me to go ahead and do that **graph sync merge** to get the two branches in sync again?

josephj11:
> Alas, we're [stuck going back further](https://github.com/autokey/autokey/discussions/1152).

So the 0.94 stuff was pre-Sphinx? Is that why we have it?

Personally, I don't want to support anything older than 0.95.10.

At least now, if we go ahead with the documentation archives as discussed in that link. we won't have that problem going forward.

josephj11:
> By the way, did you want me to go ahead and do that **graph sync merge** to get the two branches in sync again?

If that's the thing that makes the forward and back PR conflicts between develop and master go away, then fine. If not, bear with me and explain it again.

Elliria:
It was that thing and it worked. Wheee! 😁

Issue #646, opened 2022-01-02
AttributeError when trying to set hotkey


sdanil-ops:
## Classification:
Bug

## Reproducibility:
Allways

## AutoKey version:
0.95.10-4

## Used GUI:
gtk

## Installed via:
yay

## Linux distribution:
5.15.7-1-MANJARO #1 SMP PREEMPT Wed Dec 8 10:09:19 UTC 2021 x86_64 GNU/Linux

## Summary:

Hi! Getting AttributeError when trying to set hotkey

## Steps to reproduce:
-Set HotKey

## Expected result:
Hotkey sets 

## Actual result:
Nothing but AttributeError

## Verbose output (if applicable):
```
2022-01-02 14:35:02,912 ERROR - interface - Error in X event loop thread
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/autokey/interface.py", line 242, in __eventLoop
    method(*args)
  File "/usr/lib/python3.9/site-packages/autokey/interface.py", line 674, in __grab_keyboard
    focus.grab_keyboard(True, X.GrabModeAsync, X.GrabModeAsync, X.CurrentTime)
AttributeError: 'int' object has no attribute 'grab_keyboard'
```

Elliria:
Interesting. A couple of questions:
- Does it happen with one specific hotkey or any hotkey you try to set?
- Does it happen on all of your scripts or just one?

josephj11:
A few more questions:
- Did this work for you before or is this a new install?
- If it worked before, what has changed/been updated since then.

Note: I didn't immediately find the focus module in our source code. Is it possible that its calling sequence has changed in a new version of an external package? Manjaro (Arch) gets bleeding edge updates.

- While we are investigating this..., is it possible to use the Qt version instead?

   On Arch, the two versions of AutoKey (GTK and Qt) are packaged separately, so this may require some extra effort to test. If you have any scripts and phrases already defined, be sure to make a backup of $HOME/.config/autokey (including hidden files) before doing any package management.

mikemp7:
I had this happening to me with `/usr/lib/python3/dist-packages/autokey/gtkui/dialogs.py`

This is what I did to fix it 
Replace `text = model.get_value(i.next().iter, 0)` with `text = model.get_value(next(i).iter, 0)` 

`i.next()` is Python 2 syntax.

Elliria:
Excellent find! It's added to the list of things to do once we can edit the code again.

josephj11:
[...]

PR #1164, opened 2026-06-25
fix(gtkui): save abbreviation text when cell editing is cancelled

### LINKED ISSUE #667: Setting an abbreviation still requires pressing Enter

- [X] This is not a question.

UI/Usability

Kubuntu 18.04

GTK

0.96.0 Beta 10

From GitHub debs

When setting a new abbreviation, one must press Enter or click the mouse in the abbreviations panel or the new abbreviation is lost.

Always

Select any script or phrase
Click Set Abbreviation
 Click + to add an abbreviation
Type any abbreviation string
Select OK

The abbreviation should be added. Or, OK should be disabled.

### What actually happened?

Two cases:
If there is no abbreviation, an error message is displayed saying an abbreviation must be added.
If there is already an abbreviation, nothing seems to happen and it looks like it worked.
In both cases, the new abbreviation is discarded and the user is clueless as to why this occurred and how to fix it.

### Do you have screenshots?

_No response_

_No response_

### Anything else?

This is an old bug. Both #220 and #539 are closed.

Autokey-qt  still requires pressing Enter or clicking in the abbreviations panel, but has the OK button disabled if the abbreviation has not been "finalized" and there are instructions detailing this process at the top of the dialog.

Apparently, these enhancements never made it into autokey-gtk.

Ideally, if an abbreviation has been typed (in either front end) and OK is selected, it should just work, but at a minimum, we need the changes made to autokey-qt duplicated in autokey-gtk.

kesmeey:
## Summary

Fixes #667 - Setting an abbreviation still requires pressing Enter.

## Problem

When editing an abbreviation in the Abbreviation Settings dialog, pressing Enter triggers the "edited" signal which saves the new text. However, clicking away or pressing Tab triggers the "editing-canceled" signal, which was restoring the OLD text — losing the user's work.

## Fix

- Connect the "editing-started" signal to capture a reference to the GtkEntry editable widget
- In the "editing-cancelled" handler, read the current text from the editable and commit it to the model if it's non-empty and different from the original
- Preserves existing behavior for empty/whitespace-only fields (removes the row)

## Test plan

- [ ] Open abbreviation settings for any phrase/script
- [ ] Edit an existing abbreviation, click away (or press Tab) — text is saved
- [ ] Add new abbreviation, type text, click away — text is saved
- [ ] Edit abbreviation to empty, click away — row is removed
- [ ] Press Escape to cancel editing — does not save (GTK default)

### DIFF
--- a/lib/autokey/gtkui/dialogs.py
+++ b/lib/autokey/gtkui/dialogs.py
@@ -232,6 +232,7 @@ class AbbrSettingsDialog(DialogBase):
         textRenderer.set_property("editable", True)
         textRenderer.connect("edited", self.on_cell_modified)
         textRenderer.connect("editing-canceled", self.on_cell_editing_cancelled)
+        textRenderer.connect("editing-started", self.on_cell_editing_started)
         column1.pack_end(textRenderer, True)
         column1.add_attribute(textRenderer, "text", 0)
         column1.set_sizing(Gtk.TreeViewColumnSizing.FIXED)
@@ -363,10 +364,20 @@ class AbbrSettingsDialog(DialogBase):
 
     # Signal handlers
 
+    def on_cell_editing_started(self, renderer, editable, path, data=None):
+        self._current_abbr_editable = editable
+
     def on_cell_editing_cancelled(self, renderer, data=None):
         model, curIter = self.abbrList.get_selection().get_selected()
         oldText = model.get_value(curIter, 0) or ""
-        self.on_cell_modified(renderer, None, oldText)
+        newText = None
+        if hasattr(self, '_current_abbr_editable') and self._current_abbr_editable is not None:
+            newText = self._current_abbr_editable.get_text()
+            self._current_abbr_editable = None
+        if newText is not None and newText != oldText and not EMPTY_FIELD_REGEX.match(newText):
+            model.set(curIter, 0, newText)
+        elif EMPTY_FIELD_REGEX.match(newText or "") and EMPTY_FIELD_REGEX.match(oldText):
+            self.on_removeButton_clicked(renderer)
 
     def on_cell_modified(self, renderer, path, newText, data=None):
         model, curIter = self.abbrList.get_selection().get_selected()


josephj11:
@kesmeey Welcome to the AutoKey community!

Thank you! This fix adrresses an annoying long-term problem that I'd really like to see fixed.

We'll have to wait for one of our devs to review it.

However, (assuming the review is positive), a couple of changes will be needed.

All new changes need to be based on our develop branch. We only make emergency changes to master.

An entry to CHANGELOG.rst is needed to document the change.

Elliria:
Thank you for your interest in AutoKey, @kesmeey. Unfortunately, we don't accept pull requests that are based on the **master** branch and ask that all contributions be based on the **develop** branch.

The **develop** branch has made significant changes to the `dialogs.py` file (specifically updating iterator syntax to Python 3, as seen in issue #646 and fixed in [pull request $1076](https://github.com/autokey/autokey/pull/1076)) and many other files.

Also, it seems that you and @awen11123 both submitted pull requests for the same issue at about the same time, with @awen11123's being [pull request #1165](https://github.com/autokey/autokey/pull/1165). You two might want to get together and either figure out which of the two pull requests would be more beneficial for AutoKey or see if you can work together to do a joint pull-request.

kesmeey:
> Thank you for your interest in AutoKey, @kesmeey. Unfortunately, we don't accept pull requests that are based on the **master** branch and ask that all contributions be based on the **develop** branch.
> 
> The **develop** branch has made significant changes to the `dialogs.py` file (specifically updating iterator syntax to Python 3, as seen in issue #646 and fixed in [pull request $1076](https://github.com/autokey/autokey/pull/1076)) and many other files.
> 
> Also, it seems that you and @awen11123 both submitted pull requests for the same issue at about the same time, with @awen11123's being [pull request #1165](https://github.com/autokey/autokey/pull/1165). You two might want to get together and either figure out which of the two pull requests would be more beneficial for AutoKey or see if you can work together to do a joint pull-request.

Hi, thanks for the guidance. I'll withdraw my PR and let @awen11123 take it forward. Appreciate the help!

PR #1165, opened 2026-06-25
fix: show instructions and disable OK until abbreviation is finalized in GTK dialog

### LINKED ISSUE #667: Setting an abbreviation still requires pressing Enter

awen11123:
## Summary

Prevents loss of newly entered abbreviations when users click OK without first pressing Enter to finalize the cell edit. Mirrors the autokey-qt behavior.

Fixes #667

## Changes

- **abbrsettings.xml**: Add a `GtkLabel` instruction at the top: "Type a new abbreviation and press Enter to add it, then click OK."
- **dialogs.py**: Disable the OK button until at least one abbreviation exists in the list. Wire `_update_ok_button()` into add/remove/edit handlers.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -50,6 +50,7 @@ Important misc changes
 - Update the help menu, deprecating one entry, adding several entries, updating existing wording, and sorting the entries.
 - Update the logger by removing an unneeded space and making the **cutelog** reference match the new command-line switch for it in the help menu.
 - Remove special handling of ignoreCase and matchCase options in abbreviation settings dialogs, allowing phrases to trigger on any input case while matching input case in the output (see #588).
+- Add instruction label and disable OK button in GTK abbreviation settings dialog until an abbreviation is finalized, to prevent silent data loss and match the Qt behavior. Fixes `#667 <https://github.com/autokey/autokey/issues/667>`__.
 - Update the GTK and Qt man pages.
 - Update date, formatting, and NAME section in the GTK and Qt man pages.
 - Fix typo: Replace all occurrences of "they key" with "the key" in the AutoKey documentation.
--- a/lib/autokey/gtkui/data/abbrsettings.xml
+++ b/lib/autokey/gtkui/data/abbrsettings.xml
@@ -64,6 +64,20 @@
           </packing>
         </child>
         <child>
+          <object class="GtkLabel" id="instructionLabel">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="label" translatable="yes">Type a new abbreviation in the entry and press Enter to add it to the list. Then click OK.</property>
+            <property name="wrap">True</property>
+            <property name="xalign">0</property>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="position">0</property>
+          </packing>
+        </child>
+        <child>
           <object class="GtkHBox" id="hbox1">
             <property name="visible">True</property>
             <property name="can_focus">False</property>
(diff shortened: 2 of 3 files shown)


josephj11:
See my comments in the iissue.

Elliria:
Thank you for your interest in AutoKey, @awen11123. Unfortunately, we don't accept pull requests that are based on the **master** branch and ask that all contributions be based on the **develop** branch.

The **develop** branch has made significant changes to the `dialogs.py` file (specifically updating iterator syntax to Python 3, as seen in issue #646 and fixed in [pull request $1076](https://github.com/autokey/autokey/pull/1076)) and many other files.

Also, it seems that you and @kesmeey both submitted pull requests for the same issue at about the same time, with @kesmeey's being [pull request #1164](https://github.com/autokey/autokey/pull/1164). You two might want to get together and either figure out which of the two pull requests would be more beneficial for AutoKey or see if you can work together to do a joint pull-request.

chrislee35:
Thanks for this, @awen11123 — nice, focused fix, and matches what autokey-qt already does per josephj11's screenshot in the issue.

I checked the diff against current develop: the concern raised earlier about this being based on master doesn't seem to hold up in practice — the touched methods (on_cell_editing_cancelled, on_cell_modified, reset_focus, etc.) match develop's current dialogs.py line-for-line, and this applies cleanly.

One gap worth calling out for review: this gates the OK button on "the list is non-empty," which correctly covers the issue's literal repro (adding a first abbreviation to an empty list) and matches Qt's behavior. It doesn't cover editing an already-existing abbreviation — if the list already has an entry and a user edits it, then clicks OK without pressing Enter or clicking away first, that specific edit can still be lost silently, since has_items stays True throughout. @kesmeey's now-withdrawn #1164 handled that case too (by committing the live editor text on cancel for any row, not just gating on list-emptiness). Not a blocker — this does what it says and matches Qt parity — but worth knowing before merge, in case it's worth folding in that part of #1164's approach as a follow-up.

josephj11:
@chrislee35 @awen11123 Seems like this works. Is the recommendation to proceed with the merge or to fold in the additional desired changes first?

chrislee35:
I recommend proceeding as is and fixing the pending edit issue in a separate PR.

josephj11:
@chrislee35 Can you file the new issue so it gets described correctly...

chrislee35:
Filed: #1185

Issue #732, opened 2022-09-15
Error message should pop up in a dialog window


mapsedge:
### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

UI/Usability

### Which Linux distribution did you use?

Operating System: Kubuntu 22.04
KDE Plasma Version: 5.24.6
KDE Frameworks Version: 5.92.0
Qt Version: 5.15.3
Kernel Version: 5.15.0-47-generic (64-bit)
Graphics Platform: X11
Processors: 4 × Intel® Xeon® CPU E31225 @ 3.10GHz
Memory: 15.5 GiB of RAM
Graphics Processor: NV167

### Which AutoKey GUI did you use?

Qt

### Which AutoKey version did you use?

0.95.10

### How did you install AutoKey?

Discover software center

### Can you briefly describe the issue?

Instead of showing a notification and making the user go through a menu selection to see the error, Tools -> "Show last script error", [have the option to] just pop the message up into a dialog. I've never gotten a notification where I didn't go see what the problem was.

### Can the issue be reproduced?

Always

### What are the steps to reproduce the issue?

1. create a script with an error
2. observe the notification 
3. wish there was a better way
4. make coffee and weep into your cup

### What should have happened?

There's a problem, here's the problem, done.

### What actually happened?

A notification and extra effort.

### Do you have screenshots?

**Bad.**
![Screenshot_20220914_195233](https://user-images.githubusercontent.com/41483625/190288370-df11b274-9936-4501-b471-0acfa6bf0952.png)

**Good**
![Screenshot_20220914_195332](https://user-images.githubusercontent.com/41483625/190288397-ac445594-1082-44c5-8f41-8b5320427f3d.png)

### Anything else?

That about covers it.

Elliria:
I heartily agree that having the details in the notification would be so much better than having to manually go and click a menu entry to get at them, especially considering how often I like to break scripts with all sorts of experimentation.

josephj11:
This sounds interesting. There would have to be a little parsing of the traceback so that just relevant part is displayed and not the whole long thing.

[...]

Issue #699, opened 2022-06-14
Main window list columns cannot be managed individually


ineuw:
## Classification:
Enhancement, UI/Usability

## Reproducibility:
Always

## AutoKey version:
0.96.0

## Used GUI:
Gtk

## Installed via:
pip3

## Linux distribution:
Linux Mint Cinnamon 20.3
Cinnamon 5.2.7

## Summary:
The left panel of the main window lists three columns but it's impossible adjust their widths and sort them as well. 

## Expected result:
Wanted to sort and change the width of the abbr.  and hotkey columns individually

## Actual result:
Cannot be adjusted

## Screenshot (if applicable):
![220614_frozen columns](https://user-images.githubusercontent.com/10367016/173507327-339e35da-abbb-4adc-ae2c-3a31c77e549d.jpg)

josephj11:
I didn't get this at first, but being able to sort by abbreviation and hotkey would be very useful for managing/avoiding conflicts and deciding what to use for a new entry.

The use of window filters complicates this a bit because it can eliminate potential conflicts, but this isn't exposed in the panel. Fully supporting this might be difficult, but it might help to colorize or otherwise highlight entries which have window filters set.
.
A (sortable) windows filters column might be useful in this regard. It could be hidden by default if you don't want to sacrifice the screen space it would consume. This might make it useful to be able to sort by hotkey or abbreviation as a primary key with window filters as a secondary key. Alternatively, the window filter could display in a tooltip on hover.

Being able to "print"/export a sorted list could be useful as well - especially for users with large numbers of entries.

Screen real estate is always limited, so being able to resize the columns is not a bad idea either.
Some thought should be given to field values that are larger than the displayed field width (scrollable or wrapped fields).

Sorry for complicating this request, but it got me thinking. :)

Elliria:
I'll join in with a +1 on the ability to rearrange and resize those columns. This is especially important when you don't open AutoKey in full-screen mode. I actually make mine bigger than I'd like it to be so that the columns show.

ineuw:
Sorting the individual columns are far less important than the ability to adjust the column width of "Abbr".  I am never sure if I see everything typed.
P.S: Elliria's suggestion of rearranging the columns with sorting of the first column only, might be a solution.

Elliria:
I can't take credit for that one. It was Joe's idea. I'm all for it, though. Being able to choose how to sort things gives us power and you never know which kind of power you might want at any given moment.

ineuw:
[...]

Issue #666, opened 2022-02-11
Phrase editor panel theme problems


josephj11:
### Has this issue already been reported?

- [X] I have searched through the existing issues.

### Is this a question rather than an issue?

- [X] This is not a question.

### What type of issue is this?

Bug

### Which Linux distribution did you use?

kubuntu 18.04

### Which AutoKey GUI did you use?

GTK

### Which AutoKey version did you use?

0.96.0 Beta 10

### How did you install AutoKey?

From GitHub debs

### Can you briefly describe the issue?

Phrase editor panel ignores dark theme - shows black text on white background
and
Text cursor is invisible. The current character blinks sometimes.

### Can the issue be reproduced?

Always

### What are the steps to reproduce the issue?

Open the main menu and select a phrase, Try to modify the phrase text.

### What should have happened?

Text cursor should be thin vertical line or block depending on insert/overwrite setting

### What actually happened?

Text cursor is invisible. Makes editing difficult.

### Anything else?

Works perfectly in autokey-qt.

Elliria:
The Qt version works perfectly for me as well, with a dark background, white text, and a blinking white cursor.

I can't replicate this, however, in the GTK version in this environment:
* Kubuntu version: Kubuntu 20.04 LTS running inside of VirtualBox
* KDE Plasma version 5.18.8
* KDE Frameworks version: 5.68.0
* Qt version: 5.12.8
* Kernel version: 5.13.0-28-generic
* OS Type: 64-bit
* Global theme: Breeze Dark

In my case, the GTK version completely ignores my dark theme and uses the light theme with black text and a blinking black cursor. That could be due to some system setting not being toggled properly. This is a default installation of Kubuntu in a VM that I haven't tinkered with, so please let me know if there's a setting I should try.

josephj11:
I use Breeze dark as well, but generally tinker with the settings (without taking notes :(  ), so I don't know what to tell you to try other than maybe temporarily switching to the Oxygen (or another) theme to see what happens.

Elliria:
I just tried AutoKey GTK and AutoKey Qt with Oxygen and both worked fine with the icon turning red each time I tried to run the "whoops" script.

StowasserH:
I have the same problem. There are also some scripts in the editor that are very difficult to read. Because dark blue syntax highlighting on black background is not good.
Switching the KDE system theme to "light" helped. Even if it's not pretty :-(

josephj11:
[...]

Issue #632, opened 2021-11-19
AutoKey process not killed by panel icon context menu Quit option


Elliria:
## Classification:
Bug

## Reproducibility:
Always

## Autokey version:
autokey-gtk 0.96.0-beta.9

Used GUI:
Gtk

Installed via:
pip3

Linux distribution:
- Kubuntu 20.04 LTS running inside of VirtualBox
- KDE Plasma Version 5.18.5
- KDE Frameworks Version: 5.68.0
- Qt Version: 5.12.8
- Kernel Version: 5.11.0-40-generic
- OS Type: 64-bit
- Memory: 3.2 GiB of RAM
- Processor: 1 x Intel Core i5-8600K CPU @ 3.60 GHz

## Summary:
The AutoKey process is not killed by the panel icon's context menu "Quit" option.

## Steps to reproduce:
1. Launch autokey-gtk.
2. Check your system for any autokey processes: ```pgrep -c "autokey"```
    - Expected result: Result should be 1.
    - Actual result: Result is 1.
5. Right-click the AutoKey icon in the panel and choose "Quit".
6. Check your system for any autokey processes: ```pgrep -c "autokey"```
    - Expected result: Result should be 0.
    - Actual result: Result is 1.
 
## Expected result:
Multiple expected results. See above.

## Actual result:
Multiple actual results. See above.

## Notes:
This issue doesn't affect closing AutoKey from the "Quit" option inside the "File" menu in the AutoKey main window. That one properly kills the AutoKey process.

BlueDrink9:
I can't reproduce with gtk. Can you please check again after a restart, and this time first do `pkill autokey; pgrep -c autokey` to ensure there aren't any other processes triggering a false positive?

BlueDrink9:
Also note that autokey does take some time to shut down, so leave 30 seconds in between

BlueDrink9:
I'm going to close for now, but I will re-open if you can confirm it is occuring. Provide logs, too.

Elliria:
I'm attaching a log of the output.

## The steps that I took:
This is exactly what I did to create that log:
1. I booted my VirtualBox virtual machine.
2. I launched AutoKey in a terminal window from a shortcut on my panel.
3. I right-clicked the AutoKey panel icon and chose "Quit" from the menu.
4. I checked for the AutoKey process in another terminal window.
5. The AutoKey process still existed.
6. I waited 30 seconds.
7. I checked for the AutoKey process in another terminal window.
[...]

PR #1175, opened 2026-07-17
Qt UI improvements: sortable/resizable columns, dark theme editor, search, inline editing, persisted window state

theRoadLessOrdinary:
Two related sets of Qt UI improvements to the tree/editor:

**Sortable/resizable columns, theme, persisted state, shutdown (#699, #168, #666, #632, #732)**
- Sortable tree columns (phrases and scripts intermix alphabetically)
- Editor theme auto-detects light/dark system theme
- Window size, position, panel sizes, and tree expand state persist between sessions
- Tray icon quit now fully terminates the process
- Script errors open the dialog directly instead of a tray notification

**Resizable columns, search, and inline editing**
- Fixes a header resize mode bug: a bitwise-OR of two `QHeaderView` enum values collapsed to `ResizeToContents`, silently disabling column dragging
- Adds a single search box (filters name, abbreviation, and hotkey together) below the tree, with per-parent-folder auto-expand on match
- Adds double-click shortcuts on the Hotkey and Window Filter values that skip the modal "Set..." dialogs (inline capture with Save/Escape/Enter/Tab handling, hotkey-conflict checking, live window-detection crosshair for filters)
- Fixes a header-click event filter that was swallowing column-resize-grip clicks meant for `QHeaderView`'s own drag handling
- Fixes `DetectDialog` being parented to a dialog that's never shown in the inline window-filter flow, which confused the window manager

Happy to split into separate PRs if reviewers would prefer that, or adjust based on feedback.


josephj11:
Welcome back!

This looks quite interesting, but the code is way over my head.

Before it can be reviewed, it needs to be rebased on the develop branch. We do not enhance the master branch which is intended to remain as close to the current release as possible.

Once it is rebased, it will eventually need to be split into separate PRs so we can manage it better, but I think it can be preliminarily reviewed before that. We'll see what the reviewer(s) say.

We currently have one member who might review this (after rebase), but we currently have no developers who can approve the merge and make a release that incorporates it. So we're stuck ATM.

chrislee35:
Hi @theRoadLessOrdinary — per josephj11's note above, I've rebased this onto develop: #1184.

Most of it applied cleanly (7 of the 12 files hadn't diverged from master at all), but qtapp.py needed real reconciliation — develop restructured Application.__init__/shutdown independently since this was written (part of the Wayland-support work). I merged both sides' intent rather than picking one; details are in #1184's description.

One thing worth flagging here directly: while testing the rebase by actually launching the app under Xvfb, I hit a NameError: name 'cm' is not defined in the new show_configure() geometry-restore code — cm/cm_constants were imported on master for unrelated reasons and this PR's diff never needed to add them itself, but develop had dropped those same imports in its own refactor. I've added them back in #1184, but figured you'd want to know in case it affects how you think about this on your end too.

I haven't done a full pass of the new interactions (live search, column drag-resize, inline hotkey/window-filter capture) — planning to review that behavior in detail as a follow-up. Thanks for the work here, this is a nice set of improvements.

PR #1178, opened 2026-08-05
fix: read NumLock and CapsLock state from the X server

### LINKED ISSUE #1177: numpad numbers are seen as directional keys

Xorg

- [x] This is not a question.

UI/Usability

- [ ] autokey triggers
- [ ] autokey-gtk
- [ ] autokey-qt
- [ ] beta
- [ ] bug
- [ ] critical
- [ ] development
- [ ] documentation
- [ ] enhancement
- [x] phrase expansion
- [ ] scripting
- [ ] technical debt
- [ ] user interface

_No response_

Linux mint cinnamon 21.3

GTK

0.95.10

From mint install software

I can create triggers with numpad numbers

But on expansion, I've to use numbers on the main keyboard to trigger the expansion. 

Using numbers on the numpad triggers nothing, bc numbers on numpad are seen as directional keys. 

Always

1. Insert 1/2 to trigger ½
2. Type 1/2 on the numpad
3. Nothing will expand 

1/2 should expand to ½, no matter if we type numbers on the main keyboard or numbers on the numpad. 

### What actually happened?

No expansion

### Do you have screenshots?

_No response_

```bash

```

### How did you find the AutoKey project?

_No response_

### Anything else?

No
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #1177</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @Jopp-gh</code> (replace <code>20</code> with the amount, and <code>@Jopp-gh</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

RawNuke:
Fixes #1177

## Problem

Phrase triggers created with numpad digits (example: abbreviation `1/2`) expand when typed on the main keyboard, but never when typed on the numpad. AutoKey sees the numpad digits as directional keys.

## Root cause

`XInterfaceBase.__init__` initialized the tracked NumLock/CapsLock state from the core X protocol `led_mask`:

```python
ledMask = self.localDisplay.get_keyboard_control().led_mask
```

Under XKB, which is the default keyboard extension on every modern distribution, the core `led_mask` frequently reads `0` while the NumLock LED is on. AutoKey then tracks NumLock as off, `lookup_string()` skips the `XK_TO_AK_NUMLOCKED` branch, and numpad digits decode to the directional names (`Key.NP_END`, `Key.NP_DOWN`, ...). Those names are non-simple keys, so `__updateStack()` clears the input stack and no abbreviation matches.

## Fix

Query the true lock state from the X server instead of trusting the LED mask:

- `keysym_to_keycode(XK_Num_Lock)` / `keysym_to_keycode(XK_Caps_Lock)` locate the lock keys.
- `get_modifier_mapping()` maps each modifier index to its keycodes.
- `query_pointer().mask` on the root window reports the current locked-modifier state.
- The LED-mask path is preserved as a fallback.

The new `query_lock_state()` helper is pure and unit-testable without an X server.

## Tests

- 8 new unit tests in `tests/test_interface.py` (fake display, no X server): both locks on/off, non-standard modifier index, keycode absent from mapping, XError on the pointer query.
- Full suite: 327 passed vs 319 baseline, same 10 pre-existing environmental failures, zero new failures.

Author: RawNuke

### DIFF
--- a/lib/autokey/interface.py
+++ b/lib/autokey/interface.py
@@ -252,6 +252,35 @@ class XWindowInterface(AbstractWindowInterface):
             return None
 
 
+def query_lock_state(display, root_window):
+    """
+    Query the current lock state of CapsLock and NumLock.
+
+    The core GetKeyboardControl led_mask is unreliable under XKB,
+    which is the default on modern distributions. Query the modifier
+    mapping and the current pointer state instead.
+
+    @param display: The X display to query.
+    @param root_window: The root window of the display.
+    @return: A tuple (capslock_on, numlock_on) with the current states.
+    """
+    numlock_on = capslock_on = False
+    numlock_keycode = display.keysym_to_keycode(XK.XK_Num_Lock)
+    capslock_keycode = display.keysym_to_keycode(XK.XK_Caps_Lock)
+    if numlock_keycode or capslock_keycode:
+        modifier_keycodes = display.get_modifier_mapping()
+        try:
+            pointer_mask = root_window.query_pointer().mask
+        except error.XError:
+            pointer_mask = 0
+        for index, keycodes in enumerate(modifier_keycodes):
+            if numlock_keycode and numlock_keycode in keycodes:
+                numlock_on = bool(pointer_mask & (1 << index))
+            if capslock_keycode and capslock_keycode in keycodes:
+                capslock_on = bool(pointer_mask & (1 << index))
+    return capslock_on, numlock_on
+
+
 class XInterfaceBase(threading.Thread, AbstractMouseInterface):
     """
     Encapsulates the common functionality for the two X interface classes.
@@ -570,9 +599,16 @@ class XInterfaceBase(threading.Thread, AbstractMouseInterface):
         self.join()
 
     def __set_lock_keys_state(self):
-        ledMask = self.localDisplay.get_keyboard_control().led_mask
-        self.mediator.set_modifier_state(Key.CAPSLOCK, (ledMask & CAPSLOCK_LEDMASK) != 0)
-        self.mediator.set_modifier_state(Key.NUMLOCK, (ledMask & NUMLOCK_LEDMASK) != 0)
+        try:
+            capslock_on, numlock_on = query_lock_state(
+                self.localDisplay, self.rootWindow)
+        except Exception:
+            logger.exception("Failed to query lock state; falling back to LED mask")
+            ledMask = self.localDisplay.get_keyboard_control().led_mask
+            capslock_on = (ledMask & CAPSLOCK_LEDMASK) != 0
+            numlock_on = (ledMask & NUMLOCK_LEDMASK) != 0
+        self.mediator.set_modifier_state(Key.CAPSLOCK, capslock_on)
+        self.mediator.set_modifier_state(Key.NUMLOCK, numlock_on)
 
     def __eventLoop(self):
         while True:
--- a/tests/test_interface.py
+++ b/tests/test_interface.py
@@ -139,3 +139,129 @@ class TestXrecord():
             # Need to cancel early. But cancel in tearDown as well in case this test fails.
             self.cancel()
         assert_that(self.ec.get_result(), is_(equal_to(expected)), failmsg)
+
+
+# Tests for the lock-state query helper (query_lock_state) added for #1177.
+# These use fake display objects. They do not need an X server.
+from Xlib.error import XError
+from autokey.interface import query_lock_state
+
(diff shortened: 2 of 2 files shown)


josephj11:
@RawNuke Thanks for researching and solving this!

However, we do not accept non-emergency PRs against the master branch. It needs to be rebased on the develop branch.

We do not currently have any developers to test and approve this (or any PR), so it may be a while before it can be merged.

AutoKey is strictly volunteer and has no formal organization let alone funds. Individuals can post bounties on Opire, but, so far, only one issue has a bounty and the people who solved it won't claim it. Go figure.

I am not an AutoKey/Python developer, but I just glanced at the code. One thing I noticed was something about hamcrest. AFAIK, we use that for development testing, but I don't think it's a dependency for normal users/installs, so that might be an issue.

RawNuke:
Thanks for the review and for pointing me to `develop`.

A heads-up on the rebase: `develop` has substantially refactored `lib/autokey/interface.py` (~850 lines changed versus `master`), so this fix does not replay cleanly onto it. The NumLock/CapsLock handling would need to be re-implemented against the new structure rather than rebased as-is, and `tests/test_interface.py` has also diverged (an add/add conflict).

Since there arent reviewers available right now, I will hold off on a blind re-implementation and leave this PR as-is against `master` for reference. I am happy to redo it properly on top of `develop` whenever someone is available to test and review, and I will account for your note about `hamcrest` being development-only in that redo.

josephj11:
@RawNuke Sorry I didn't have better news for you. However, you have still identified a problem and an approach to solving it, so that is much appreciated. I agree about leaving this open for same reason as you stated.

RawNuke:
Rebased onto `develop` as requested. A plain rebase conflicted because `develop` had refactored the startup lock-state code into `__set_lock_keys_state`, so I ported the change onto that method: it now calls the new `query_lock_state()` helper and falls back to the old `led_mask` on any error. `self.rootWindow` is set by `__initMappings()` before this runs, so the query path is the one actually used.

On the hamcrest note: `develop` already ships `tests/test_interface.py` using hamcrest, so I left those tests untouched and appended the lock-state tests in the same style. They use fake display objects and need no X server. The diff is now +165/-3 across `interface.py` and the test file, one commit against `develop`. Ready for another look when you have time.

chrislee35:
Verified this thoroughly, including live testing:

**Code-level verification (worktree, before/after):**
- On unpatched `develop`, the new tests fail to even collect
  (`ImportError: cannot import name 'query_lock_state'`) -- confirms
  they exercise genuinely new behavior, not something already covered.
- With this patch applied, all 8 new tests pass, and the rest of
  `test_interface.py` is unaffected (same 10 passed / 4 xfailed
  baseline as before).
- Traced the integration by hand: `self.rootWindow` (needed by
  `query_lock_state()`) is set in `__initMappings()`, which runs
  before `__set_lock_keys_state()` -- confirmed against the current
  `develop` code, not just the PR description.

**Live X11 testing (Linux Mint 22.3, Cinnamon):**
I have a VM for exactly this kind of testing. I directly compared the
old `led_mask` read against the new `query_lock_state()`, cross-checked
against `xset q`'s XKB indicator state as ground truth, across three
scenarios: live NumLock/CapsLock toggles via XTest (5/5), immediate
reads on a brand-new `Display()` connection with NumLock already on
(30/30), and reads after replaying AutoKey's actual init sequence
(30/30). In all of these, `led_mask` matched ground truth every time
on this VM -- I could not reproduce the described unreliability here.

That doesn't mean the bug isn't real -- the original report was Mint
21.3, and XKB/LED-mask behavior is known to vary by distro, XKB
ruleset, or keyboard model, so this may simply not reproduce on this
particular setup. But I can't personally confirm the root-cause
mechanism the way I've been able to on some other PRs.

Net: the fix itself is correct, safe (has a working fallback to the
old behavior), and well-tested, and querying real X state is a more
robust approach regardless of whether led_mask happens to be reliable
on any given system. I'd recommend merging on those grounds, while
flagging that I couldn't independently reproduce the original symptom
to confirm the root cause.

PR #1198, opened 2026-09-10
docs: align autokey-run man page with autokey-gtk/qt (closes #869)

### LINKED ISSUE #869: Update the autokey-run man page

Xorg

- [X] This is not a question.

Documentation

- [ ] autokey triggers
- [ ] autokey-gtk
- [ ] autokey-qt
- [ ] beta
- [ ] bug
- [ ] critical
- [X] development
- [X] documentation
- [ ] enhancement
- [ ] phrase expansion
- [ ] scripting
- [X] technical debt
- [ ] user interface

_No response_

_No response_

None

_No response_

_No response_

Update the [autokey-run](https://github.com/autokey/autokey/blob/develop/doc/man/autokey-run.1) man page format to be similar to the one that's being used in the in-process updates to the [autokey-gtk](https://github.com/autokey/autokey/blob/develop/doc/man/autokey-gtk.1) and [autokey-qt](https://github.com/autokey/autokey/blob/develop/doc/man/autokey-qt.1) man pages (including the new **SEE ALSO** section that they use). 

Also add this to the **script** option since it's a new feature that was added in AutoKey 0.96.0:

Accepts full paths to Python scripts. If no full path is given, will treat the name as an existing AutoKey script name instead.

Always

Examine the existing man page.

The updates should be there.

### What actually happened?

They're not there yet.

### Do you have screenshots?

_No response_

_No response_

### Anything else?

_No response_

muhammadbinriaz:
Resolves #869.

Rewrites `doc/man/autokey-run.1` to match the format used by
`autokey-gtk.1` and `autokey-qt.1`:

- Modern `\fB...\fR` markup and a consistent
  NAME / SYNOPSIS / DESCRIPTION / OPTIONS / AUTHOR layout.
- Adds the full **SEE ALSO** section the other two pages carry
  (sibling commands, related software, similar tools, and the
  AutoKey web pages).
- Documents that `--script` accepts full paths to Python scripts
  (added in AutoKey 0.96.0); without a full path the name is treated
  as an existing AutoKey script name instead.

Targeting `develop` since that's where the current man pages live.

### DIFF
--- a/doc/man/autokey-run.1
+++ b/doc/man/autokey-run.1
@@ -1,9 +1,14 @@
 .\"                                      Hey, EMACS: -*- nroff -*-
+.\" Please adjust this date whenever revising the man page:
+.TH AUTOKEY-RUN "1" "September 10, 2026"
+.\"
 .\" First parameter, NAME, should be all caps.
 .\" Second parameter, SECTION, should be 1-8, maybe w/ subsection.
 .\" Other parameters are allowed: see man(7), man(1).
-.TH AUTOKEY-RUN "1" "April 19, 2023"
-.\" Please adjust this date whenever revising the man page.
+.\"
+.\" TeX users may be more comfortable with the \fB<whatever>\fR and
+.\" \fI<whatever>\fR escape sequences to invoke bold-face or italics,
+.\" respectively.
 .\"
 .\" Some roff macros for reference:
 .\" .nh        disable hyphenation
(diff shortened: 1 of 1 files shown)


josephj11:
Welcome to the AutoKey community, @muhammadbinriaz !

Thanks, this looks useful.

@chrislee35 Can you take a look at this? I got as far as rendering it locally and it looks good, but I didn't verify the content.

chrislee35:
Verified this against the actual code, not just the PR description:

- `autokey-run`'s real argparse options are -s/--script, -p/--phrase,
  -f/--folder (plus argparse's automatic -h/--help) -- matches the new
  SYNOPSIS exactly. Bonus: the old SYNOPSIS omitted -h even though
  OPTIONS already documented it; this fixes that too.
- The --script full-path behavior is accurate: Service.run_script()
  (lib/autokey/service.py:257) checks for an absolute, existing path
  and falls back to looking up an AutoKey script by name otherwise --
  exactly what the new doc text says.
- Format and SEE ALSO content closely mirror autokey-gtk.1/autokey-qt.1,
  as #869 asked for.

Looks good to merge.

muhammadbinriaz:
I've rebased this onto the current `develop`. The earlier CI failures were from an outdated base (before the xvfb/pytest CI fixes landed), not from this change — it only touches `doc/man/autokey-run.1`. Could a maintainer approve the workflow run when you get a chance? Thanks!

PR #1200, opened 2026-09-11
fix: import TriggerMode from correct module in engine and test_service

### LINKED ISSUE #1199: Engine.create_abbreviation()/create_hotkey() raise AttributeError: module 'autokey.model.helpers' has no attribute 'TriggerMode'
`lib/autokey/scripting/engine.py`'s `create_abbreviation()` and
`create_hotkey()` (both public scripting API methods) reference
`autokey.model.helpers.TriggerMode`. That attribute doesn't exist --
an earlier refactor moved `TriggerMode` into its own module
(`autokey.model.triggermode`) without updating these two call sites.

Any script calling `engine.create_abbreviation()` or
`engine.create_hotkey()` hits an `AttributeError`. `tests/test_service.py`
has the same stale reference in its `create_script()` test helper,
which is why `test_set_triggered_abbreviation_uses_typed_abbreviation`
fails.

This has been silently broken since the refactor -- it was only
caught because CI has been failing at an earlier "Install dependencies"
step (missing `libgirepository-2.0-dev`, see #1182) since well before
this bug was introduced, so the test suite never actually ran.

Fix: import `TriggerMode` from `autokey.model.triggermode` in both
files, matching the pattern already used elsewhere (e.g.
`lib/autokey/model/folder.py`, `tests/test_phrase.py`).

<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #1199</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @chrislee35</code> (replace <code>20</code> with the amount, and <code>@chrislee35</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

ghzhost:
## Description

Fixes #1199.

An earlier refactor moved `TriggerMode` into its own module (`autokey.model.triggermode`), but `lib/autokey/scripting/engine.py` and `tests/test_service.py` still referenced `autokey.model.helpers.TriggerMode`, causing an `AttributeError` on `engine.create_abbreviation()`, `engine.create_hotkey()`, and `test_set_triggered_abbreviation_uses_typed_abbreviation`.

This PR updates the imports and references to `autokey.model.triggermode.TriggerMode` in both files.

/claim #1199

### DIFF
--- a/lib/autokey/scripting/engine.py
+++ b/lib/autokey/scripting/engine.py
@@ -24,6 +24,7 @@ from typing import Tuple, Optional, List, Union
 import autokey.model.folder
 import autokey.model.helpers
 import autokey.model.phrase
+import autokey.model.triggermode
 import autokey.model.script
 from autokey import configmanager
 from autokey.model.key import Key
@@ -299,7 +300,7 @@ Folders created within temporary folders must themselves be set temporary")
 
         self.monitor.suspend()
         p = autokey.model.phrase.Phrase(description, contents)
-        p.modes.append(autokey.model.helpers.TriggerMode.ABBREVIATION)
+        p.modes.append(autokey.model.triggermode.TriggerMode.ABBREVIATION)
         p.abbreviations = [abbr]
         folder.add_item(p)
         p.persist()
@@ -339,7 +340,7 @@ Folders created within temporary folders must themselves be set temporary")
 
         self.monitor.suspend()
         p = autokey.model.phrase.Phrase(description, contents)
-        p.modes.append(autokey.model.helpers.TriggerMode.HOTKEY)
+        p.modes.append(autokey.model.triggermode.TriggerMode.HOTKEY)
         p.set_hotkey(modifiers, key)
         folder.add_item(p)
         p.persist()
--- a/tests/test_service.py
+++ b/tests/test_service.py
@@ -27,6 +27,7 @@ import tests.helpers_for_tests as testhelpers
 
 import autokey.model.folder
 import autokey.model.helpers
+import autokey.model.triggermode
 import autokey.model.script
 from autokey.configmanager import configmanager
 from autokey.configmanager.configmanager import ConfigManager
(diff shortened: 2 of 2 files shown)


chrislee35:
Thanks for jumping on this so quickly, @ghzhost!

Heads up: this overlaps with #1201, which I'd gently recommend using
instead -- it's the counterpart to #1182 (the CI pipeline fix). #1182
had to temporarily mark this exact test `xfail` (referencing #1199) so
its own pipeline could go green while this fix was pending, and #1201
both applies the fix and removes that marker in the same commit. Since
this PR was branched before #1182 merged, it doesn't have that marker
to clean up -- merging it as-is would fix the bug but leave the xfail
in place, showing as a silent XPASS instead of a normal pass.

To be clear, this is a special, one-time situation: normally a fix
like this would just be one PR. It only ended up split across two
(#1182 and #1201) because #1182 needed the pipeline unblocked before
the real fix could land, so the temporary workaround and its cleanup
were sequenced separately.

Two smaller things, in case #1201 doesn't end up being used: this PR
leaves `import autokey.model.helpers` in both files even though
nothing else uses it after the fix (won't be caught by CI -- the
flake8 config here doesn't check for unused imports), and the rest of
the codebase imports TriggerMode as `from autokey.model.triggermode
import TriggerMode` rather than the fully-qualified form used here.

josephj11:
@chrislee35 Your argument makes sense to me.

Too bad it needs to supersede a new new contributors efforts.

This is the kind of problem I like to have though!

chrislee35:
I agree.  It hurt when I put myself in @ghzhost's shoes: jumping on a problem quickly and applying such a great patch, only to have the issue submitter already have a patch ready.  I apologize for not being clearer in issue #1199.  Again, this was a special situation to get the pipeline back to health.

josephj11:
I'm going to close this because it got done in the other PR. This was a valid PR and we appreciate getting it.

Issue #1189, opened 2026-09-07
Window detection (crosshair click-to-select) silently fails on native-Wayland windows


chrislee35:
### Has this issue already been reported?
- [x] I have searched through the existing issues.

### Is this a question rather than an issue?
- [x] This is not a question.

### What type of issue is this?
Bug

### Which AutoKey GUI did you use?
Qt

### Can you briefly describe the issue?
The "detect window properties" feature (used by the Window Filter "Set..." dialog, and by #1184's new inline double-click shortcut on the Window Filter value) silently does nothing when the user clicks a window that isn't running via XWayland.

Root cause: window-click detection goes through `iomediator.windowgrabber.WindowGrabber`, which registers as an `IoMediator.listeners` entry and waits for `handle_mouseclick` — driven entirely by `XRecordInterface`'s use of the X11 RECORD extension (`lib/autokey/interface.py`). RECORD only observes X11 protocol traffic, so it's blind to clicks on genuinely native-Wayland client windows (GTK4 apps, many current GNOME apps, Firefox/gnome-terminal in Wayland-native mode, etc.) — anything not running through XWayland. There's no error, no timeout, no feedback of any kind: the crosshair cursor (or the modal dialog, for the old "Set..." flow) just sits there forever with nothing to show for the click.

Confirmed on a real Wayland VM (GNOME): `xlsclients` showed only AutoKey's own windows running via XWayland — gnome-terminal and Firefox were both native-Wayland and clicking on either produced no reaction whatsoever, no log output, nothing.

This isn't new to #1184 — the old modal "Set Window Filter" dialog's "Detect Window Properties" button uses the identical `WindowGrabber` mechanism and has the same limitation. #1184 just made it easier to trigger accidentally via a double-click shortcut, with no dialog open to make the "waiting for a click" state visible at all.

### What are the steps to reproduce the issue?
1. On a Wayland session, open AutoKey's Window Filter settings for any phrase/script (old modal dialog or #1184's inline double-click shortcut).
2. Click "Detect Window Properties" (or double-click the Window Filter value, in the new inline flow).
3. Click on a native-Wayland application window (e.g. a GTK4 app, or Firefox/GNOME Terminal running Wayland-native rather than via XWayland — check with `xlsclients` to see which of your running apps are XWayland-backed).

### What should have happened?
Either window detection should work regardless of whether the target is XWayland or native Wayland, or the user should get clear feedback (an error, a timeout message) that detection isn't available for that window / isn't supported on this session type.

[...]

PR #1208, opened 2026-09-13
fix: Wayland window detection silent fail (WindowGrabber / uinput) (#1189)

elecelias-ux:
## Summary

Addresses #1189: **Detect window properties** (WindowGrabber crosshair / click-to-select) hangs forever on native Wayland windows with no feedback.

On `develop` under Wayland, AutoKey uses `UInputInterface` (not XRecord). Mouse `BTN_*` presses were read from evdev but **never forwarded** to `IoMediator` listeners, so `WindowGrabber` never completed. This PR:

1. Forwards mouse button presses from `UInputInterface` to `IoMediator.handle_mouse_click`, resolving window info via the existing GNOME extension / KWin window interface (focused window after click — best-effort on Wayland). When absolute/relative coords cannot be resolved, the button event is still forwarded without inventing `(0, 0)`.
2. Adds an **8s timeout** (`DEFAULT_DETECT_TIMEOUT_SECONDS`) on `WindowGrabber` plus Qt/GTK warning dialogs so detection cannot fail silently.
3. Adds unit tests (`tests/test_windowgrabber.py`) for grabber timeout/click and uinput button → mediator forwarding (mocked desktop deps).

Follow-up for @chrislee35 review: timeout test waits on dialog callback (not listener membership); BTN-up swallow kept + unit-tested (`_consume_mouse_button_event`); GTK timeout dialog uses `MODAL | DESTROY_WITH_PARENT`.

### Honesty

**Not runtime-tested on a real Wayland session** — the agent environment has no Wayland (`WAYLAND_DISPLAY` unset). Please treat this as draft until someone verifies on GNOME/KDE Wayland using the checklist below.

## Manual test checklist (Wayland)

- [ ] GNOME Wayland + AutoKey GNOME extension enabled
- [ ] KDE Wayland (KWin scripting)
- [ ] Detect → click native Wayland app → DetectDialog shows title/class
- [ ] Detect → wait with no click → timeout warning ~8s, button re-enabled
- [ ] Detect → click XWayland app still works
- [ ] X11 session regression: Detect still works

## Test plan (automated)

- [x] `PYTHONPATH=lib pytest tests/test_windowgrabber.py -o addopts=` → 11 passed (with stubs for dbus/gi/evdev where needed)

Refs #1189


elecelias-ux:
@coderabbitai review

elecelias-ux:
Friendly ping on #1208 (Wayland WindowGrabber / uinput silent fail). Ready for review whenever a maintainer has a slot — note: not yet runtime-tested on a real Wayland session from my side. Happy to iterate on feedback.

josephj11:
@chrislee35 can you review this (haven't checked if you did already)?

Initial notes:

lib/autokey/qtui/dialogs/windowfiltersettings.py 108

AutoKey does not have any mouse properties in its configs AFAIK

15 second timeout mentioned in the general description sounds a bit long

lib/autokey/uinput_interface.py 956 962

is setting coords to (0,0) if not retrieved

elecelias-ux:
Thanks @josephj11 — noted, happy to adjust while @chrislee35 reviews.

1) windowfiltersettings.py ~108: I’ll drop any mouse-property wiring that isn’t a real AutoKey config. Window detect should only use the existing grabber / window interface.
2) Timeout: I’ll cut 15s → 8s (still enough to click, less “hung” feeling). Easy to make it a constant if you prefer another value.
3) uinput_interface.py ~956/962: (0,0) was a fallback when evdev didn’t give abs coords. I’ll stop inventing a click position — forward the button event and let WindowGrabber resolve the focused Wayland window instead.

I’ll push that in a follow-up commit. Still not runtime-tested on a live Wayland session; checklist in the PR stands.

josephj11:
As far as 2 and 3, those were questions, not recommendations. Maybe wait until Chris comments.

chrislee35:
Thanks for the PR, and welcome — looks like this is your first contribution.

Reviewed this against current `develop` (no drift — base commit diffs clean against HEAD for all touched files) and traced the root cause: on `develop`, `UInputInterface.handle_mouseclick` is dead code (its only call site is commented out, and its body calls `mediator.handle_mouse_click` with the wrong signature anyway), so `WindowGrabber` genuinely never sees a click under uinput/Wayland. The fix direction — forward `BTN_*` presses and resolve the window via the existing GNOME-extension/KWin `get_window_info()` — matches what #1189 itself suggested, and the three notes from earlier in the thread (mouse-config tip, 15s timeout, invented `(0, 0)`) all look addressed in the second commit.

I ran the test suite (`PYTHONPATH=lib pytest`) and found one issue:

**Flaky test:** `test_windowgrabber_timeout_notifies_dialog` fails intermittently (~1/20 runs, even in isolation; also failed once in a full-suite run). Root cause: `WindowGrabber._on_timeout()` removes itself from `IoMediator.listeners` (via `cancel()`) *before* calling `dialog.receive_window_detect_timeout(...)`. The test polls until the grabber leaves `listeners` and then immediately asserts the dialog callback ran, but there's a small window between delisting and the callback where the assertion can fire first. Harmless in production, but will occasionally flake CI as-is — worth waiting on the callback itself (or a short poll on the mock) rather than on list membership.

**Question on scope:** the `__flush_events` restructuring, besides adding button-down forwarding, also now fully swallows mouse-button *up* events (they hit the new `if mouse_button is not None:` branch and match neither of its `keystate == 1` case nor any of the old `elif`s). Previously a `BTN_LEFT` release fell through to `handle_keyrelease` → `mediator.handle_keypress(...)`. That old path looks like it was already producing garbage rather than doing anything useful, so dropping it is plausibly fine — but it's a behavior change beyond what's described in the PR and isn't covered by a test. Was this intentional?

Minor nit: the new GTK timeout dialog uses `Gtk.MessageDialog(..., flags=0, ...)`, while the other warning/error dialogs in `gtkui/dialogs.py` and `configwindow.py` use `Gtk.DialogFlags.MODAL | Gtk.DialogFlags.DESTROY_WITH_PARENT`. Cosmetic, but worth matching for consistency.

elecelias-ux:
Thanks @chrislee35 — appreciated the thorough review.

1) Flaky `test_windowgrabber_timeout_notifies_dialog`: agree. I'll wait on the dialog callback/mock instead of listener membership so the race between `cancel()` delisting and `receive_window_detect_timeout` can't flake CI.

2) Button-up swallowing in `__flush_events`: intentional cleanup of the old dead path (BTN release → `handle_keyrelease` → garbage `handle_keypress`). Happy to (a) keep the swallow + add a unit test documenting it, or (b) restore an explicit no-op with a comment — I prefer (a) unless you want the old fallthrough preserved.

3) GTK flags: will match `Gtk.DialogFlags.MODAL | Gtk.DialogFlags.DESTROY_WITH_PARENT`.

I'll push a follow-up once you confirm (a) vs (b) on the release path. Still no live Wayland runtime from my side; checklist in the PR stands.

elecelias-ux:
Follow-up pushed (a8de001) addressing your three notes:

1) Timeout test now polls the dialog mock callback, not listener membership.
2) Kept intentional BTN-up swallow (option a) via `_consume_mouse_button_event` + unit tests documenting no `handle_keyrelease` / `handle_keypress` on release.
3) GTK timeout dialog flags → `MODAL | DESTROY_WITH_PARENT`.

`PYTHONPATH=lib pytest tests/test_windowgrabber.py -o addopts=` → 11 passed (timeout test also 25/25 in isolation).

chrislee35:
verified, LGTM. Thanks for the follow up. 
@josephj11, it's ready for merge.

josephj11:
Merging. Curious: what's the impact on all of this for users with fancy mice pressing additional buttons?

PR #1212, opened 2026-09-14
Fix KDE Wayland window property response scope

keeltrace:
## Summary

Fix a KDE Wayland KWin scripting bug in `KdeWindowInterface.get_properties()`.

`KWinInterface._load_script()` appends `callDBus(..., result)` after the supplied JavaScript. `get_properties()` previously declared `result` with block-scoped `let` separately inside its `if` and `else` branches, so the appended callback referenced an out-of-scope identifier and could not return the current window-property response.

This change declares `result` once in the outer script scope and assigns to it inside either branch.

## Regression coverage

Adds a focused regression test that verifies the generated `get_properties` script keeps `result` in the outer scope and does not reintroduce branch-local declarations.

Verification on Python 3.14:

* regression test copied onto the unfixed parent commit: **fails as expected**
* same regression test on this commit: **passes**
* `tests/test_kde_interface.py`: **26 passed**
* project CI flake8 subset: **0 findings**
* `git diff --check`: **passes**

The final composed JavaScript, including the callback appended by `_load_script`, was also executed in a JavaScript runtime for both window-found and window-not-found branches; both produce the expected `get_properties` payload.

## Scope

This is deliberately narrow and does not change the KWin transport architecture, timeout behavior, or other Wayland backends.

Related to #87.

## Manual testing

I did not have an authorized graphical KDE Wayland session available for live desktop verification, so I am not claiming live KDE runtime testing. The defect itself is deterministic JavaScript scope behavior and is covered by the regression and composed-script checks above.

## AI assistance disclosure

This patch was developed with AI assistance. I reviewed the resulting code and verified the underlying failure mode directly: `result` was declared with block-scoped `let` inside both branches while `_load_script()` appends a callback that reads `result` outside those blocks. I also verified the unfixed regression fails, the fixed regression passes, both composed-script branches produce the expected payload, and the focused KDE interface tests pass.


keeltrace:
/claim 87 

This fixes a concrete defect in the Wayland support associated with #87. It does not implement wayland supprt its self. Partial claim is requested

josephj11:
@chrislee35 This looks interesting.

@keeltrace Thanks. We'll take a look at this.
Your issue comments make several references to JavaScript while this project is entirely in Python AFAIK. 

The bounty on #87 is not available in practical terms. The work on it is almost done and the two users who deserve it IMO have declined to claim it. That's why we closed the issue.

keeltrace:
> @chrislee35 This looks interesting.
> 
> @keeltrace Thanks. We'll take a look at this.
> Your issue comments make several references to JavaScript while this project is entirely in Python AFAIK. 
> 
> The bounty on #87 is not available in practical terms. The work on it is almost done and the two users who deserve it IMO have declined to claim it. That's why we closed the issue.


I was wondering, I only submitted this because the bounty was increased, sorry if I've over stepped. I'm new to git.

Feel free to consider this a donation, no obligation for the claim. 

Feel free to suggest any improvements and I can update that into the pr. 

Best, 
Keeltrace 

Edit: On the JavaScript point: kde_interface.py generates .js scripts for KWin. This PR fixes result being block-scoped inside that generated script, which prevented the appended D-Bus callback from accessing it.

josephj11:
@keeltrace No problem! We're glad to get any valid contributions of code or other efforts.

Opire doesn't seem to adequately address the case where someone has made a partial contribution to an issue with a bounty and doesn't expect to get the whole thing.

Ultimately, bounties are resolved directly between claimants and donors. AutoKey as a project is not involved with the process.

Obviously, we each can have opinions on claims and our maintainers can choose whether or not to accept the actual work product and that may influence the decisions of the donors.

I didn't realize there was a JavaScript component. Always something new to learn!

chrislee35:
Thanks for the fix and the regression test — verified this independently.

The root cause checks out: `_load_script()` appends `callDBus(..., result)` after whatever script text it's given when `response_expected=True`, and `get_properties()` was the only method in the file declaring `result` with block-scoped `let` inside `if`/`else` branches rather than the outer scope — every other `response_expected=True` call site (`get_active_window`, `get_window_list`, `get_screen_size`, `get_active_desktop_index`) already declares `result` once at the top level, so this was an isolated miss, not a systemic pattern.

I copied the new regression test onto the parent commit — it fails exactly as described (`'let result;' not found in ...`) — and confirmed it passes on this commit. Also ran the full suite: 444 passed, 9 skipped, 18 xfailed, no failures.

Scope is appropriately narrow. LGTM.

dalecannegie-commits:
That's how we learn, together. One happy branched family.

josephj11:
@dalecannegie-commits Yep. Winning friends and influencing people. (off by one letter) :)

PR #1221, opened 2026-09-17
Fix hotkey resurrection on items with the hotkey trigger disabled

bjohas:
### Problem

An item that has a hotkey recorded in its JSON fires regardless of its trigger modes. Un-ticking the hotkey trigger in the GUI has no lasting effect, because `"modes": []` cannot survive deserialization.

`AbstractHotkey.set_hotkey()` force-adds `TriggerMode.HOTKEY` whenever a hotkey is recorded:

```python
def set_hotkey(self, modifiers, key):
    ...
    if key is not None and TriggerMode.HOTKEY not in self.modes:
        self.modes.append(TriggerMode.HOTKEY)
```

`inject_json_data_base()` assigns `item.modes` from the file *before* calling `AbstractHotkey.load_from_serialized()`, which routes into `set_hotkey()`. The force-add therefore overwrites the modes that were just loaded. The item lands in `ConfigManager.hotKeys` and dispatches in `Service.handle_keypress()`.

This is a regression against 0.95.10, whose `set_hotkey()` only assigned `modifiers` and `hotKey`.

### Reproducer

A phrase saved with the hotkey trigger disabled but a hotkey still recorded:

```json
{"modes": [], "hotkey": {"modifiers": ["<hyper>"], "hotKey": "a"}, ...}
```

```python
p = Phrase("", "", path=".../testphrase.txt")
p.load_from_serialized()
print([m.name for m in p.modes])
```

- before: `['HOTKEY']` — the item fires
- after: `[]` — the item stays disabled

Practical symptom: a disabled sample phrase silently shadows a real script bound to the same hotkey, and because `Service.handle_keypress()` breaks on the first match in `ConfigManager.hotKeys`, which one wins depends on load order.

### Fix

Assign `item.modes` last in `inject_json_data_base()`, so the file's value is not clobbered.

The force-add in `set_hotkey()` is deliberately left alone: every other caller (`UI_common_functions.py`, `predefined_user_files.py`, `configmanager.py`) already pairs `set_hotkey()` with an explicit `set_modes()`, but `scripting/engine.py` may rely on it, so removing it would be a wider behaviour change than this bug needs.

Also present on `master` / v0.96.0, where the same assignment is duplicated in `model/phrase.py`, `model/script.py` and `model/folder.py`. Happy to send that as a follow-up if it is wanted.


josephj11:
@bjohas Thanks for your contribution!

This is a pretty subtle bug. How did you discover it?

I'm quite confused as to how this is even possible.

From your discussion, this appears to be in common code, so it  affects AutoKey running under both front ends.

It's a bit unclear how this situation could occur under normal use unless the user is directly editing the .json metadata files.
I don't see any "untick" for hotkeys - at least in the Qt front end.

AFAIK, the only place AutoKey keeps action metadata is in the  .json companion files. I don't see where any other state information/metadata could be stored for a conflict to arise.

Are you saying that when a user Clears/resets the hotkey in the GUI, the metadata file doesn't get modified?

Does this occur just in one running session of AutoKey or persist after quit and restart?

IIUC, this bug is in 0.96.0 (and carried over to develop), but not in 0.95.10.

@chrislee35 This looks like it might be important to fix. Depending on what was "shaddowed", a user could get a very unexpected result from pressing a hotkey. I haven't wrapped my head around this one yet.

chrislee35:
I need some more sleep before I can really dig into this one. I'll take a deeper look this weekend.

bjohas:
Thanks for looking at it!

**How I found it.** A `<hyper>+a` script stopped working after upgrading 0.95.10 → 0.96.0. This was on an older Ubuntu which still had 0.95.10 as default, so it might not affect many others, unless you're in that scenario?

So, a `<hyper>+a` script stopped working and the log showed a *different* item matching:

```
INFO - autokey.service - Matched Phrase "Second phrase" with hotkey and prompt=False
```

"Second phrase" is the shipped sample phrase, and its JSON said `"modes": []`. It should never have been in `ConfigManager.hotKeys`.

**Why `"modes": []` doesn't disable it.** `inject_json_data_base()` assigns `item.modes` from the file and *then* calls `AbstractHotkey.load_from_serialized()`, which routes into `set_hotkey()`:

```python
if key is not None and TriggerMode.HOTKEY not in self.modes:
    self.modes.append(TriggerMode.HOTKEY)
```

The force-add overwrites the modes that were just loaded, so `"modes": []` cannot survive deserialisation while a hotkey is recorded. In 0.95.10 `set_hotkey()` only assigned `modifiers` and `hotKey`, so the identical file loaded as genuinely disabled.

Hope that makes sense... apologies if not. 

**How it arises without hand-editing JSON**. The untick is the **Clear** button beside the hotkey: `on_clear_hotkey_button_pressed` (Qt, `settingswidget.py:195` at v0.95.10) and `on_clearHotkeyButton_clicked` (GTK, `configwindow.py:264`). Both set `hotkey_enabled = False` and reset the dialog. But `SettingsWidget.save()` at v0.95.10 is:

```python
self.current_item.set_modes([])
if self.abbreviation_enabled:
    self.abbr_settings_dialog.save(self.current_item)
if self.hotkey_enabled:
    self.hotkey_settings_dialog.save(self.current_item)
if self.window_filter_enabled:
    self.window_filter_dialog.save(self.current_item)
else:
    self.current_item.set_window_titles(None)     # the filter has an else
```

There is an `else` for the window filter and **none for the hotkey**. So I think what happened is that Clear + Save wrote `"modes": []` while leaving the hotkey recorded — harmless on 0.95.10, where it stayed dormant. 0.96 added `else: unset_hotkey()` in both toolkits, so the *current* GUI no longer creates that state — but every item created that way before the upgrade wakes up on first load under 0.96.

While I can't rule out hand-editing of my items, it seems that it may not be limited to hand-edited configs: it's an upgrade-path issue for anyone who ever pressed Clear on 0.95.x. Either way `"modes": []` is probably a natural encoding of "this trigger is off", and 0.96 silently overrides it.

**Your other questions:**

- *Both front ends* — yes. The bug is in `lib/autokey/model/`, and the GUI-side history above is identical in GTK and Qt.
- *One session, or persistent?* — persistent. The force-add happens on every load, so quitting and restarting doesn't clear it. The only ways to silence such an item are to clear the hotkey itself or delete the item; unticking the trigger has no lasting effect.
- *0.96.0 and develop, not 0.95.10* — correct. `develop` has the same `set_hotkey()` and the same load order, in `model/common.py:inject_json_data_base()`. This PR targets `develop`; the equivalent on `master`/v0.96.0 is the same change duplicated across `model/phrase.py`, `model/script.py` and `model/folder.py`, which I'm happy to send as a follow-up.

On the shadowing you mention: `Service.handle_keypress()` breaks on the first match in `ConfigManager.hotKeys`, and that list is built in folder-walk order, so which item wins is essentially arbitrary. In my case a disabled sample phrase with no window filter was beating a live script that had one.

Hope this helps!

chrislee35:
Reviewed and verified directly against current develop.

Reproduced the bug exactly as described: loaded a phrase JSON with
`modes: []` and a recorded hotkey, and `load_from_serialized()`
returned `modes == ['HOTKEY']` instead of `[]`. Confirmed the root
cause: `AbstractHotkey.set_hotkey()` unconditionally force-adds
`TriggerMode.HOTKEY` whenever a hotkey is present, and
`inject_json_data_base()` assigned `item.modes` from the file before
that call ran, so the force-add clobbered it.

Applied the fix (moving the `item.modes` assignment after the three
`load_from_serialized()` calls) and confirmed: `modes` correctly stays
`[]` while `hotKey`/`modifiers` remain intact, so re-enabling later
still works without re-recording the combo.

Also checked whether running `item.modes` after the other two
`load_from_serialized()` calls (abbreviation, window filter) introduces
any new side effect, since they now execute first. They don't:
`AbstractAbbreviation.load_from_serialized()` never force-adds
`TriggerMode.ABBREVIATION` (that only happens via the explicit
`add_abbreviation()`/`add_abbreviations()` GUI calls, not on
deserialization), and `AbstractWindowFilter` doesn't touch `modes` at
all. So the reorder fixes exactly the one real bug with nothing else
affected.

Full suite green with the fix applied (498 passed, same baseline as
develop).

One gap: no test was added. This is a real, previously-silent
regression with a clean, easy-to-express repro (the JSON snippet
already in the PR description would work almost as-is as a test
fixture). Given the project's general pattern of pairing bug fixes
with a regression test, it'd be good to see one here before merge,
otherwise a future refactor of `inject_json_data_base()`'s ordering
could silently reintroduce this with nothing to catch it.

Otherwise this looks correct and appropriately scoped. Agree with
leaving `set_hotkey()`'s force-add alone given the other callers'
dependency on it.

josephj11 (commented):
@chrislee35 Another bit to review

bjohas:
Thank you for checking this so thoroughly, @chrislee35, including the side effects of the reorder.

Test added in 8c17961, as a new `tests/test_model_common.py`:

- `"modes": []` with a recorded hotkey survives a JSON round trip for `Phrase`, `Script` and `Folder`, and the hotkey itself is kept so re-enabling does not need it re-recorded.
- The enabled case (`[HOTKEY]`) for contrast.
- The reproducer from the description, loaded from a file on disk through `load_from_serialized()`.

The four disabled-trigger cases fail with the old ordering and pass with the fix.

chrislee35:
Verified the tests work as promised. LGTM. Merge it.

Issue #305, opened 2019-08-26
Allow (optionally negated) Window Filter independently on wm_title and wm_class


mmhere:
## Classification:

Feature (New) / Enhancement

## Reproducibility:

Always

## Version

AutoKey version: 0.95.7

Used GUI (Gtk, Qt, or both): GTK (likely same in Qt)

Installed via: dpkg --install *.deb

Linux Distribution:  Ubuntu 16.04

## Summary

Allow (optionally negated) Window Filter independently on wm_title and wm_class

### Two parts.  Part One:

Code in model.py (roughly line 346) tests the GUI-entered Window Filter for Scripts like this:

    return bool(r.match(window_info.wm_title)) or bool(r.match(window_info.wm_class))

This is not always what you want and can match many more windows than you might expect, or not match anything if you have specific/detailed window titles.  For example, my xterm windows display very detailed titles that constantly change based on the version of shell running within, the current directory, whether tmux is active etc.  So matching against them doesn't always work as you expect.  It is very difficult to come up with a single regexp to make this work.

Matching solely against the window class ('xterm.XTerm') in this case, works better.

Therefore, a choice to only match against wm_class (alternately only match against wm_title) would be great.  Maybe an arrangement of check boxes or radio buttons to achieve this.  You'd still like to be able to (r.match(window_info.wm_title) or r.match(window_info.wm_class))  but these effective logic forms would also be very useful:

    return bool(r.match(window_info.wm_title))

or:

    return bool(r.match(window_info.wm_class))

In my copy I hacked model.py to do the latter;  it does what I require.

### Part Two.  Negation:

Sometimes you want to specify a regexp in Window Filter that is an EXCEPTION.  That is, apply to all windows EXCEPT FOR ones matching the regexp.

Inside the script, checking for window title/class to negate it isn't always what you want because by then the keystroke used to activate the script is consumed.  In some apps you can simply pass this back through with send_keys(the_key), but you have to hard code the_key string in the script.

Further, some apps (like xterm) require special settings (allowSendEvents enabled) to make the key come thru.  That setting in turn forcibly disables other settings you might like to use in xterm (allowWindowOps, allowFontOps).

So for now I have a "negative" regexp (that's a hack!) to skip xterm but allow all other windows.  This required the modification to model.py described above.

So a checkbox GUI modification:

    Window Filter:   [____regexp_______]   [x] Negate

or some such.

mmhere:
[...]

Issue #161, opened 2018-06-30
Include more sophisticated regex/exclusion features


ghost:
## Classification:

Feature

## Reproducibility:

Always

## Version

AutoKey-Py3 version 0.90.4 / 3.6

Installed via: `sudo apt install autokey-gtk`

Distro: Ubuntu GNOME 17.10

## Summary 

When creating shortcuts, you can provide a regular expression to make sure the shortcut only works on applications whose class names match. However, I have had much difficulty coming up with a regular expression that *excludes* applications based on their class name. 

For example, I have shortcuts for copy and paste that send the `<ctrl>+c` and `<ctrl>+v` keys, respectively, to any application I'm on. However, I don't want these shortcuts to be enabled for the terminal (URxvt), since it doesn't register those keybindings as copy and paste and it performs a completely irrelevant action upon using those shortcuts. 

It would be nice if I could simply provide a regular expression that excludes certain class names or, better yet, have the ability to simply click on a particular window to exclude it, just like how I can use the detect properties feature to include an application window.

Or maybe I can make it so that I can use one shortcut that sends different keys depending on the application I'm currently on. So `<alt>+d` would send `<ctrl>+c` to my browser while `<alt>+d` would send `<ctrl>+<alt>+c` to my terminal.

## What I've Tried

I've tried using the regexes `(?!^Terminal$)(^.*$)` and `^((?!Terminal).)*$`, to no avail. It doesn't seem like any exclusion filters work, from what I've searched on the internet. I don't even know which type of regex autokey uses. 

I eventually decided on creating a shortcut that bound `<alt>+d` to the following script:
```
 if window.get_active_class() == 'urxvt.URxvt':
        keyboard.send_keys("<ctrl>+<alt>+c")
 else:
        keyboard.send_keys("<ctrl>+c")
```

I did the same for paste (replacing the `c` with `v`) and now copying and pasting works everywhere. However, I feel like for functionality as simple as this, it should be a standard feature and not something the user has to code in.

josephj11:
First of all, this should work, so it's a valid issue. I have never seen an example of a successful exclusion pattern. This is something a number of users (including me) want.

Problem: The version and version number of AutoKey you are reporting is impossible.

[...]

PR #1223, opened 2026-09-18
Add an invert option to the window filter

bjohas:
### Problem

The window filter is include-only: one regex meaning "trigger only in windows matching this". There is no way to express "trigger everywhere **except** here", so anyone who wants that has to enumerate every application that is *not* the one they want to exclude.

A real config, from the folder its author named `Scripts_Ctrl_Emulation_for_all_apps_bar_Terminal_itself`:

```
google-chrome.OpenDevEd.*|.*.Chrome2|.*.Google-chrome|autokey-gtk.Autokey-gtk|
Navigator.Zotero|.*.Io.elementary.files|.*.gnome.Nautilus|.*Navigator.Firefox|
... 28 alternations in total ...
```

The intent is "everything but the terminal". Checked against the windows open on that machine, the allowlist silently missed `brave-browser.Brave-browser`, `evince.Evince` and a bare `google-chrome` class — every new application is broken by default until someone edits the regex again.

**A negative lookahead does not substitute for this.** `_should_trigger_window_title()` applies the regex to `wm_title` and `wm_class` separately and ORs the results:

```python
return bool(r.match(window_info.wm_title)) or bool(r.match(window_info.wm_class))
```

`(?!.*gnome-terminal).*` correctly rejects the class `gnome-terminal-server.Gnome-terminal`, but that window's title is something like `bjohas@host: ~`, which contains no "terminal" — so the title arm matches, the OR succeeds, and the item fires anyway. Excluding by class is impossible unless the excluded string also happens to appear in every title.

Inverting at the match site gives `not (title or class)`: the window is excluded if **either** property matches.

### What this does

Adds `isInverted`, mirroring `isRecursive` — same serialization shape, same setter style, default `False`, one check box in the filter dialog in both toolkits.

Three commits:

1. **Model.** The flag travels *with* the regex through the inheritance walk, so a child inheriting a parent's recursive filter uses the parent's invert flag rather than its own; `get_applicable_filter_inverted()` mirrors `get_applicable_regex()` line for line. Read back with `.get()`, because existing config files have no such key and a `KeyError` here is swallowed by the broad handler in `model/common.py`, which would load the rest of the item incompletely rather than fail loudly. Includes the first unit tests for `AbstractWindowFilter`.

2. **Grab walk.** `__grab_ungrab_recurse()` obtains `WindowInfo` with `traverse=False`, and `_get_window_info()` fills unknown properties with empty strings in that mode, so window-manager frames arrive as `WindowInfo("", "")`. For an include filter that is harmless — the regex fails to match, and the walk descends. For an inverted filter, `not (False or False)` is `True`, so it would grab on the frame and then recurse with `checkWinInfo=False`, grabbing unconditionally across the whole subtree including the excluded application. AutoKey would take the key there and then decline to fire, swallowing the keystroke. `_should_grab_on_window()` makes an unidentifiable window explicitly undecidable; the sibling walk `__recurse_tree_grab_ungrab()` already guarded this case and now shares the decision function.

3. **UI.** The check box stays visible for every item type, unlike the recursive one — inverting is meaningful for phrases and scripts, not just folders. The Qt grid had no row 1, so nothing needed renumbering.

4. **Summary label.** The window filter summary on an item's page showed the bare regex, so an inverted filter and the same filter uninverted looked identical while meaning the opposite — you had to open the dialog to tell. It is now prefixed when the applicable filter is inverted:

   ```
   kitty.*|gnome-terminal.*   ->   ⊘ EXCEPT  kitty.*|gnome-terminal.*
   ```

   Label only: `get_filter_regex()` feeds the dialog's regex entry box, so decorating that would corrupt the round-trip and eventually persist the decoration as part of the pattern. The marker follows the *applicable* filter, so an item inheriting a parent's inverted filter is marked too. The formatter is shared in `UI_common_functions` and used by all four label sites, so the toolkits cannot drift.

### Testing

`tests/test_window_filter.py` is new (16 test functions, 21 cases): the wm_title-vs-wm_class asymmetry above, inheritance in both directions, a parity test guarding the two walks against divergence, back-compat loading of a filter dict with no `isInverted` key, and `GlobalHotkey` — which sets the filter attributes by hand because `AbstractHotkey.__init__` does not call `AbstractWindowFilter.__init__`, hence the class-level default.

`tests/test_interface.py` gains three tests that exercise the grab walk against a mock window tree (`root → frame → client → widget`, only the client carrying a real title and class). Reverting commit 2 makes the inverted case fail: the excluded terminal's frame, client and widget all get grabs.

Full suite locally: 473 passed, 1 pre-existing failure (`test_service.py::test_start`) also present on unmodified `develop`.

### Deliberately out of scope

- `filter_matches()` compares pattern strings, so two items with the same regex but opposite invert flags are reported as conflicting. Conservative — a spurious "already in use" warning, never a missed conflict.
- No scripting API parameter; `isRecursive` is not exposed there either.
- **The flag is not correctly handled on the uinput/Wayland path, and the effect may be worse than a no-op.** `uinput_interface.py:862-867` matches `windowInfoRegex` against `wm_title` only, bypassing the mixin and so also ignoring `wm_class`, inheritance and `isRecursive`. Because that is the *block* decision while the *trigger* decision goes through the model, an inverted filter would disagree with itself there: the key is swallowed in the application the user excluded, and everywhere else both the application and AutoKey act. This is from reading the code: I have no uinput setup to confirm it. Routing that check through `_should_trigger_window_title()` looks right but is a behaviour change to the Wayland path that wants Wayland testing, so it is not in this PR.
- `__needsMutterWorkaround` already over-grabs `SUPER` hotkeys under GNOME/Cinnamon/Unity regardless of the filter; that defeats include filters today and will defeat exclusion too.


josephj11:
I couldn't follow the technical part of this, so I'll just comment at the functional level.

- The title OR class instead of being separated has been a huge pain and really needs to be fixed. There are a couple of related issues [here](https://github.com/autokey/autokey/issues/305) and [here](https://github.com/autokey/autokey/issues/161) for it.
- Using a "negate", "except", or "doesn't match" option has been suggested before and is a very good idea
- In addition to this, what I'd really like to see is a new option that allows the user to supply a Python [window filter script](https://github.com/xpusostomos/autokey/tree/issue305)  that could do anything as long as it returns a true or false and doesn't run too long.
- I'm a Python/OOP newbie, but having an abstract window filter class sounds good to me, especially if it makes it easier for my prior suggestion to be implemented.

josephj11:
tests/test_window_filter.py 1 : Please don't include a generic copyright. There is way too much confusion around copyrights. There's a whole [discussion](https://github.com/autokey/autokey/discussions/1193) of it. 

GPL 3 already applies to the whole project, so including the notice multiple times is just noise. (I know others have done it too.)

josephj11:
@chrislee35 Can you look at this? It addresses a longstanding deficiency in AutoKey. It doesn't have to be in 0.97.0, but it would be great for 0.97.1.

For anyone: The action metadata files include an isRecursive flag that I think this PR takes into account. AFAIK, nothing in AutoKey actually uses it. It eludes me ATM, but someone implemented (but never merged) Chorded Hotkeys for AutoKey where you could press several hotkeys in succession to run an action. I believe that used isRecursive to store subsequent hotkeys "within" the first hotkey.

FYI: This has been suggested in the past:

`(?i)^((?!((my_username[@]my_computer)|(terminal))).)*$`

Case insensitive match all windows which do not have something
constant in the title or something constant in the class.

bjohas:
Thank you for the feedback, that's really helpful — and thanks especially for the pointers to #305 and #161, which I hadn't found.

**Copyright header** — removed and pushed, thanks for explaining. I'd copied the block from another file without thinking, and you're quite right that asserting a copyright holder wasn't mine to do.

**On the lookahead approach** — I did try that first, and I should say it does work; I just found it a bit brittle on my own setup, which is what pushed me towards this. The thing I ran into is that the regex gets matched separately against `wm_title` and `wm_class` and the two results are OR'd, so to exclude an application I had to make sure the pattern excluded *both* of its properties. My terminal's class is `gnome-terminal-server.Gnome-terminal` but its title is just `user@host: ~`, so a lookahead naming only the class still matched via the title and the item fired anyway.

Your example handles that by naming something from the title as well as the class, which I think is exactly right. What I kept tripping over was that my titles change as I work — so a pattern that excluded the terminal correctly in the morning would let it through later. That's probably particular to how I use it rather than a general problem.

**Scope, so I don't overstate it** — this PR doesn't separate title from class, which I think is the deeper fix #305 and #161 are really asking for. It only lets you invert whatever the combined match currently produces. So it helps with "everything except X" but leaves the underlying OR as it is. I didn't want that to be misunderstood as solving those issues.

**On `isRecursive`** — I had a look at this while working on the inheritance, though only at the window-filter path, so I may well be missing the use you're thinking of. As far as I could see it's read in `get_applicable_regex(forChild=True)`: a folder's filter is offered to its children only when the folder has `isRecursive` set, which is how a filter on a folder comes to apply to the scripts inside it. My change follows the same path for the invert flag, so an item that inherits its parent's filter also inherits whether it's inverted — otherwise you could end up with one folder's regex and another's polarity. I didn't come across anything chorded-hotkey related, but I was only reading the filter code.

**On a Python window filter script** — that sounds genuinely useful, and rather more powerful than what I've done here. I don't know the codebase well enough to say whether `AbstractWindowFilter` is the right place to hang it, but if it were, I'd imagine `_should_trigger_window_title()` is the single point everything already funnels through, since both the abbreviation path and the hotkey path end up there.

Happy to adjust any of this if it's heading the wrong way.

josephj11:
Copyright: I wasn't thinking of "your place" or imputing any motive to you. I just want to keep things as simple as possible without being too simple.

Lookahead...: Thomas, one of our former lead developers, suggested that. I don't do lookaheads. I've only gotten as far as lookbacks.  Yes, window titles are highly unpredictable. That's one of the best arguments for handling them separately (which I did not think you were thinking of doing). That's a larger issue.

isRecursive: I've never seen it used for anything. I have never set any properties for a folder explicitly except for one with a hotkey so the folder becomes a menu of actions. The chorded hotkey work was never integrated into our codebase, so there's nothing explicit about it to find.

Filtering by script was my idea, but, so far, that's all it is. In bash, I almost always fall back from sed to awk when a regex gets nested more than two levels deep. It's so much easier to code/read/debug. And here, using Python, we could do the same thing allowing deep magic without an additional line of code in AutoKey itself (beyond making the feature available). I'm certainly not expecting you to do that either.

What I do want is for all these options to inform us so that what we add now doesn't get in the way of adding them in the future.

I still think the except flag is a great idea. It would be just as useful with the proposed script option as it is now. And, more importantly, it's easy to understand.

The one concern I have: AutoKey allows the same hotkey or abbreviation(s) to be associated with more than one action as long as the window filters for each such action appear to be different. (Turns out that verifying that two regexes are disjoint is a very difficult (maybe impossible) problem. AutoKey is happy with testing that they are not identical.)

If there are two actions that both share the same triggers and regex with one of the regexes negated, the code has to recognize that they are different or it will not allow the second one to be saved and activated. This shouldn't be difficult to do, but it is necessary.

bjohas:
That last point is a good catch, and you're right that it's necessary rather than optional — I had it in the PR description as a known limitation, which on reflection was the wrong call. Fixed in the latest push.

`filter_matches()` was comparing only the pattern string, so "only in X" and "everywhere but X" looked identical and the second item couldn't be saved. It now takes the other filter's polarity alongside its pattern, and the callers thread it through: `check_abbreviation_unique()`, `check_hotkey_unique()`, `get_item_with_hotkey()`, `item_has_same_hotkey()`, and the `validate()` paths in both toolkits plus the global hotkey dialog.

The new argument defaults to `False` everywhere, so nothing that doesn't know about inverted filters changes behaviour — the scripting API in particular passes a bare pattern and still compares exactly as it did.

You're quite right that deciding whether two regexes are disjoint is the hard version of this problem. I've not attempted that; this only makes the polarity part of the comparison, so the two filters you describe are now distinguishable in the same cheap "are these identical" way as before.

On not getting in the way of later work — that's a fair thing to hold me to, so for what it's worth: everything goes through `_should_trigger_window_title()`, which is the one place both the abbreviation path and the hotkey path end up. A script-based filter would presumably replace the body of that method, and an inverted flag would either apply to its result or become irrelevant because the script can negate for itself. I don't think this makes that harder, though you'd know better than I would.

The separate title/class handling does look like the bigger prize of the two. I've deliberately not touched it here.

Thanks for the context on the lookahead suggestion and on `isRecursive` — that's useful to know, particularly that the chorded hotkey work never landed, since I'd been assuming there was something I hadn't found.

chrislee35:
Thanks for another solid one, @bjohas. Appreciate the continued contributions here.

Here's where I've gotten to so far:
1. I read through the full diff and ran the PR's own test suite (36 passed, 4 xfailed in the new files, 480 passed for the full suite, matching what's described). 
2. Then I took the part unit tests can't really cover, the grab-walk fix, and tested it on real hardware. 
3. I ran both directions on Linux Mint Cinnamon, a genuine reparenting X11 window manager, not a mock, against a real gnome-terminal window:

It correctly fired and not-fired as the new design dictates.

I checked this against AutoKey's own debug log, not just visually, and saw no swallowed keystrokes and no regression to the existing include-only case through that same shared code path.

Still open, roughly in priority order:
- Inheritance through a real folder (recursive + inverted) in the actual GUI, not just the model-level test
- Both UIs' invert checkbox: does it persist across save/reload, and does the summary label update correctly
- Loading a real pre-PR config that has no `isInverted` key at all
- The polarity-aware `filter_matches()` conflict check, exercised through an actual save attempt
- The Wayland no-op. The PR says it has no effect there, and I'd like to confirm that's a clean no-op rather than something worse before this ships alongside a release that also claims Wayland support

I'll keep poking at these over the next while. Nothing here is blocking, I just want real coverage before I have an opinion on scope or timing. 

@josephj11, curious what you think should gate this versus ride along as a documented follow-up, especially on the Wayland question given the 0.97.0/0.97.1 timing you mentioned.

josephj11:
@chrislee35 Apparently, we can keep coming up with great stuff for you to do! My main concern is that 0.97.0 lands ASAP so that you still have time to devote during the beta and for at least a little while after the full release. I don't want to run out of time on non-essentials that can be added later.

Everything usually takes longer than expected and not everyone will jump on the new release let alone the beta right away, so the more runway we have left the better.

We need to do a 0.97.0 feature/enhancement freeze pretty soon.

The bug fixes will need to stay open a bit longer than that except for critical bugs (if any) that we can't allow to be released.

josephj11:
For both of you:

I wonder what would happen if I manually copy an action and its metadata file into one of my AutoKey folders while AutoKey isn't running and that action has one or more of the error conditions we're testing for when it's added to AutoKey through the menus, etc.

What gets caught and what falls through the cracks and potentially breaks things when it runs?

When something like that breaks, it's typically not obvious what broke or why.

This might be a real problem if someone gets a set of actions e.g. from one of our "featured projects", etc. They are likely to add them manually as above.

This is beyond the scope of each of these smaller PRs, but needs to be considered.

josephj11:
@chrislee35

> @josephj11, curious what you think should gate this versus ride along as a documented follow-up, especially on the Wayland question given the 0.97.0/0.97.1 timing you mentioned.

I couldn't quite parse that, so if my answer above didn't address everything, LMK.

bjohas:
Thank you both! @chrislee35, thank you particularly for testing the grab walk on real hardware. That was the part I could only cover with mocks.

**On the Wayland question:** you asked whether it's a clean no-op or something worse. Having gone and looked properly, I think it's worse...

Here `uinput_interface.py:862-867`:

```python
if item.windowInfoRegex != None:
    window_info = self.mediator.windowInterface.get_window_info()
    if not item.windowInfoRegex.match(window_info.wm_title):
        continue
```

Three things there predate this PR: it reads the raw `windowInfoRegex` attribute rather than `get_applicable_regex()`, so no inheritance and no `isRecursive`; it matches `wm_title` only, so no `wm_class`; and it has no knowledge of the invert flag.

The part that I think matters is that this is the *block* decision - whether to swallow the key - whereas the *trigger* decision goes through `service.py` and `check_hotkey_has_properties()`, i.e. the model, which this PR does make invert-aware. So the two halves would disagree:

| window | block path | trigger path | result |
|---|---|---|---|
| the excluded app | title matches, key swallowed | filter says no | keystroke lost |
| everywhere else | no match, key passes through | filter says yes | app and AutoKey both act |

Which is the same shape as the X11 grab-walk problem this PR fixes, in the other component. So "no effect on Wayland" in my PR description is wrong and I'll correct it.

**HOWEVER - caveat:** this is code reading. I have **no** Wayland or uinput setup here, so I can't demonstrate the above. If it holds, the narrow fix is to route that check through the model the way the X11 path does, but that's a separate change and I don't think it belongs in this PR.

**On scope and timing**: thanks for those notes. If you'd prefer I simply hide the invert checkbox when running under uinput until it's properly supported, that's a small change and I'll do it.

**@josephj11, on hand-copied actions:** I ran into this repeatedly while working on these, so rather than speculate, here's what I saw. The short version is that almost nothing is caught, and the item stays live in a half-initialised state. `model/common.py`:

```python
def load_from_serialized(item):
    try:
        ...
        item.inject_json_data(data)
    except Exception:
        logger.exception("Error while loading json data for " + item.description)
        logger.error("JSON data not loaded (or loaded incomplete)")
```

`inject_json_data()` may already have set some fields before it raises, so what's left in memory is a partially populated item that then gets used as though it were fine. The log says "loaded incomplete", but nothing acts on it and the user won't see it.

Two concrete instances from this work:

- In #601's original 2021 log you can watch it happen: an editor writes `folder.json` in two steps, AutoKey reads the half-written file, `set_hotkey()` raises `AttributeError: 'NoneType' object has no attribute 'sort'`, it's swallowed, and the item stays loaded.
- It's also why this PR reads `data.get("isInverted", False)` rather than `data["isInverted"]`. A subscript would raise `KeyError` on every config file written before this feature existed, and that handler would swallow it, so instead of a clear failure, every item in the file would load incompletely and silently. `sendMode` in `phrase.py` is the only existing precedent I found for the `.get()` form.

So your instinct that this would be hard to diagnose seems right to me: the failure isn't loud at any point, and by the time something behaves oddly the log line is long gone.

I agree it's well beyond these PRs. Fixing it properly probably means deciding that a malformed item should fail loudly rather than load partially, which is a behaviour change and wants its own discussion rather than being settled in a PR about something else.

Going by what you said on #1231 about pre-existing problems wanting their own issues: happy to write this up as one, with the two traces above, if that's useful. I'd be filing the question rather than a proposal, since which way it should go is yours to decide.

josephj11:
@bjohas If you want to file addition issues for the problems that this work exposed, that would be great. Filing an issue does not obligate you to work on it or to have all the answers.

Issue #1224, opened 2026-09-18
0.96.0: held modifiers are never released during expansion (fixed on develop, no release since)


bjohas:
### Summary

In 0.96.0, AutoKey never actually releases the modifier keys the user is holding when a hotkey-triggered phrase or script types its expansion. Every character arrives with those modifiers still applied.

Both defects behind this are already fixed on `develop` — incidentally, by two commits that were not about this bug and came with no tests. So the released version is broken and the development branch is not, and there has been no release since June 2022.

### Symptom

A phrase or script bound to, say, `<ctrl>+<hyper>+d`:

```python
import time
keyboard.send_keys(time.strftime("%Y-%m-%d"))
```

In a terminal that speaks the kitty keyboard protocol, this types ten fragments of the form `\e[<code>;21u` instead of the date — one per character, `21` being `ctrl(4) + hyper(16) + 1`. In an ordinary GUI text field the same bug looks like the expansion silently doing nothing, which is presumably why it has gone unreported.

### Cause 1 — the release is sent by a mechanism the X server does not see

`IoMediator.send_string()` does bracket the expansion correctly:

```python
# lib/autokey/iomediator/iomediator.py:251-257  (v0.96.0)
def _clear_modifiers(self):
    for modifier in list(self.modifiers.keys()):
        if self.modifiers[modifier] and modifier not in (Key.CAPSLOCK, Key.NUMLOCK):
            self.releasedModifiers.append(modifier)
            self.interface.release_key(modifier)
```

but `interface.release_key()` → `__releaseKey()` → `__sendKeyReleaseEvent()` → `focus.send_event(...)` (`interface.py:974-978`). An XSendEvent is delivered to a client; it never enters the server's input pipeline, so the server's key-down array and XKB modifier state are untouched and the modifier stays held.

`IoMediator.release_key()` (`iomediator.py:213-215`) is the one that routes to `interface.fake_keyup()` and XTEST, which does clear it.

### Cause 2 — and the release targets the wrong key anyway

```python
# lib/autokey/interface.py:1443  (v0.96.0)
AK_TO_XK_MAP = dict((v,k) for k, v in XK_TO_AK_MAP.items())
```

`XK_TO_AK_MAP` maps both `XK_Hyper_L` and `XK_Hyper_R` onto `Key.HYPER`, so inverting the dict keeps whichever came last — the right-hand variant. Confirmed by evaluating it:

```
<hyper> -> keysym 65518  (Hyper_R)
[...]

PR #1225, opened 2026-09-18
Add regression tests for modifier release during expansion

bjohas:
Adds tests for behaviour that `develop` already has, but which nothing currently protects.

### Why

Two commits fixed this area, neither of which was about the bug and neither of which added a test:

- **`2ad54f5`** "Refactor: Reduce calls from iomediator to interface" (2021) changed `_clear_modifiers`/`_reapply_modifiers` from `interface.release_key()`/`press_key()` to `IoMediator.release_key()`/`press_key()`. That is the difference between an **XSendEvent**, which is delivered to a client and never enters the X server's input pipeline, and **XTEST**, which does. Only the latter actually clears a modifier the user is holding.
- **`dba3345`** "Add distinction between left and right modifiers" (2023) added the explicit `AK_TO_XK_MAP[Key.HYPER] = XK.XK_Hyper_L` overrides below the dict inversion. `XK_TO_AK_MAP` maps both variants of each modifier onto one `Key`, so inverting it keeps the right-hand one; without the overrides AutoKey releases `Hyper_R` while the user holds `Hyper_L`.

`grep -rn "_clear_modifiers" tests/` currently returns nothing. Either change could be reverted by a future refactor — the first one especially, since `interface.release_key()` still exists and looks like the obvious call — and the test suite would stay green.

Both defects are live in the released 0.96.0, where a hotkey-triggered expansion types `\e[<code>;21u` fragments in a kitty-protocol terminal, or silently nothing in a GUI text field. Filed separately as #1224.

### The tests

In `tests/test_iomediator.py`:

- `test_clear_modifiers_releases_via_xtest_not_xsendevent` — asserts the call reaches `IoMediator.release_key()` and that `interface.release_key()` is **not** called
- `test_reapply_modifiers_presses_via_xtest_not_xsendevent` — the same for the re-press
- `test_capslock_and_numlock_are_not_cleared` — pins the existing exclusion
- `test_modifier_keysyms_resolve_to_the_left_hand_variant` — asserts all six overrides, so removing them fails rather than silently reinstating the right-hand keysyms

All four pass on `develop` and fail against `v0.96.0`, which is what makes them meaningful rather than tautological.

### Note for reviewers

Not changed here, but noticed while writing them: `_reapply_modifiers()` now re-presses via XTEST for real. If the user lets go of the modifier during the expansion, the re-press puts it genuinely down again with no matching physical release. On 0.96.0 that was harmless because the re-press was an XSendEvent no-op. A `self.modifiers[modifier]` re-check before pressing would close it. Happy to add that here or separately, but it is a behaviour change rather than a test so I have left it out.


bjohas:
Updated: I have dropped `test_reapply_modifiers_presses_via_xtest_not_xsendevent`.

While running this branch I hit a stuck-modifier bug caused by the very behaviour that test asserted, so pinning it with a test would have been the wrong thing. `_reapply_modifiers()` presses the user's modifiers back down unconditionally, and because `2ad54f5` made that a real XTEST press rather than an XSendEvent no-op, a modifier released while the expansion is still typing goes down with no physical release to follow and sticks.

Filed separately as #1226, with a reproduction: a script bound to `<alt>+<ctrl>+<hyper>+<shift>+,` that sleeps 100 ms before typing leaves all four modifiers down, and the keyboard stuck in shift.

The three remaining tests are unaffected — they cover the release half and the `AK_TO_XK_MAP` left-hand overrides, neither of which is in question. All still pass on `develop` and fail against `v0.96.0`.

josephj11:
@chrislee35 Please review.

chrislee35:
Reviewed. This is a clean, well-scoped addition: pure test-only (54 lines, one file, no source changes).

Verified both historical claims directly against current develop:
1. `_clear_modifiers`/`_reapply_modifiers` call `self.release_key`/`press_key`, which route through `interface.fake_keyup()`/`fake_keydown()` (XTEST), not `interface.release_key()` (XSendEvent). Matches the `2ad54f5` description exactly.
2. `AK_TO_XK_MAP` has the six explicit left-hand overrides below the dict inversion. Matches the `dba3345` description exactly.

Ran the three tests in isolation: all pass against current develop. Confirmed they're not tautological by reverting `_clear_modifiers` to call `interface.release_key()` directly (the old XSendEvent path), the regression test correctly failed.

The MagicMock usage is worth calling out since it's a nice pattern: `IoMediator._clear_modifiers(mediator)` calls the real unbound method with a mock standing in for `self`, so the actual production logic runs unmodified, only the environment (X11 interface, threading, real IoMediator construction) is faked out. That lets the test assert on exactly the thing that matters here: that `self.release_key` gets called and `self.interface.release_key` does not, which is precisely the regression a future "simplification" could reintroduce.

Two small non-blocking notes:
- The PR description still mentions a fourth test for the re-apply/re-press path. The second commit dropped it deliberately, with a good reason: it would have locked in the stuck-modifier bug now tracked as #1226. Might be worth a one-line edit to the description so it matches what's actually shipping, but the commit message already explains it well.
- This currently conflicts with develop in tests/test_iomediator.py, but it's the same append-append collision as #1231/#1235/#1236 (multiple PRs each adding a new test near the end of the same file). Test-merged locally and it resolves cleanly with no logic changes needed.

Looks good to merge from my end.

josephj11:
@chrislee35 Ready to repair then merge?

chrislee35:
yes.  Do you want me to repair it or are you comfortable doing it this time?  Just remember that if you did it wrong, the automated tests should catch anything that would break the build.

josephj11:
Is this an accept both? If so I'll do it.

Issue #1230, opened 2026-09-19
AltGr detection hardcodes keycode 108, so AltGr characters are silently remapped instead


bjohas:
### Summary

`__usableOffsets` decides which keysym offsets AutoKey can reach when sending a character. Offsets 4 and 5 are the AltGr levels, and they are enabled only when `ISO_Level3_Shift` is found on **keycode 108** at offset 0:

```python
# lib/autokey/interface.py
altList = self.localDisplay.keysym_to_keycodes(XK.XK_ISO_Level3_Shift)
self.__usableOffsets = (0, 1)
for code, offset in altList:
    if code == 108 and offset == 0:
        self.__usableOffsets += (4, 5)
        logger.debug("Enabling sending using Alt-Grid")
        break
```

Keycode 108 is where AltGr sits on a stock layout, but not on every keyboard. Where it sits elsewhere, AutoKey runs with offsets `(0, 1)` — the unshifted and shifted levels only — and every character living on an AltGr level is classified unreachable.

### Why it matters beyond AltGr characters

When `__sendString` finds a character it cannot reach, it does not give up. It rewrites the keyboard mapping with `XChangeKeyboardMapping` to borrow a spare keycode for that character — and never restores it. There is a single call site, no save, no restore, and nothing on shutdown.

The keycodes it borrows are by definition ones carrying no symbols, which are also the ones XKB has no name for. The residue therefore shows up in `xkbcomp` output as definitions on unnamed keycodes:

```
key     <> {         [      odiaeresis,      Odiaeresis ] };
```

— on keycode 8, below the first physical key. That cannot come from any XKB layout; only a core-protocol client can create it.

So a wrong answer here does not merely disable AltGr sending: it silently accumulates modifications to the user's live keyboard mapping.

### Reproduction

On a keyboard where `Control_R` is on 108 and AltGr is on 92:

```
>>> d.keysym_to_keycodes(XK.XK_ISO_Level3_Shift)
[(49, 0), (92, 0), (92, 2), (108, 2)]
```

Two keycodes carry `ISO_Level3_Shift` at offset 0 — 49 and 92 — and `mod5` is mapped to 92, so AltGr plainly works for the user. But no entry matches `code == 108 and offset == 0`, so:

```
old rule -> __usableOffsets = (0, 1)
new rule -> __usableOffsets = (0, 1, 4, 5)
```

[...]

PR #1231, opened 2026-09-19
Detect AltGr wherever it is, not only on keycode 108

bjohas:
Fixes #1230.

### The problem

`__usableOffsets` enables the AltGr levels (offsets 4 and 5) only when `ISO_Level3_Shift` is found on **keycode 108** at offset 0. That is where AltGr sits on a stock layout, and nowhere else.

On a keyboard where `Control_R` is on 108 and AltGr is on 92:

```
keysym_to_keycodes(ISO_Level3_Shift) -> [(49, 0), (92, 0), (92, 2), (108, 2)]

old rule -> (0, 1)
new rule -> (0, 1, 4, 5)
```

Two keycodes carry the symbol at offset 0, and `mod5` is mapped to 92, so AltGr works perfectly well for the user — AutoKey just does not find it.

### Why this is worse than losing AltGr sending

A character AutoKey cannot reach is not skipped. `__sendString` rewrites the keyboard mapping with `XChangeKeyboardMapping` to borrow a spare keycode for it, and never restores it — one call site, no save, no restore, nothing on shutdown. Since the borrowed keycodes are those carrying no symbols, they are also the ones with no XKB name, so the residue appears in `xkbcomp` output as:

```
key     <> {         [      odiaeresis,      Odiaeresis ] };
```

on keycode 8, below the first physical key. So the effect of this check answering wrongly is silent, accumulating modification of the user's live keyboard mapping.

On the machine this was found on, `€` at `(13, 4)` and `£` at `(12, 4)` were both being remapped on every use. With the fix they are sent directly and nothing is rewritten.

### The change

Accept any keycode whose own symbol is `ISO_Level3_Shift`, rather than requiring a particular keycode. That keeps the original intent — the key *is* an AltGr key — without assuming where it lives.

The loop is extracted into `_usable_offsets()`, a static method, so it can be tested without an X server. Six cases: a stock layout, AltGr elsewhere (the failing case), AltGr alone, `ISO_Level3_Shift` present only as a secondary symbol, present only at non-zero offsets, and absent.

### Notes

- Present identically in `v0.96.0` and on `develop`.
- This reduces how often the remapping path is entered, but does not address the remapping never being restored. That seems worth a separate discussion, since a proper fix would mean restoring on shutdown and after each send.


josephj11:
@bjohas IIUC, this should have the side effect of reducing the delays caused by remapping and make AutoKey a bit more responsive. If this is correct, then a comment here should probably reference that issue.

@chrislee35 This looks like another good one to review. It doesn't have to go in 0.97.0, but it would fit there.

bjohas:
Yes, fair - thank you! It's narrower than it might look.

**What it removes.** Each remap is an `XChangeKeyboardMapping` over the whole 8..207 range, which the X server answers with a `MappingNotify` to every client. AutoKey's own handler treats that as a keymap change and regrabs every hotkey. So a character on an AltGr level cost a full remap plus a full regrab, on every expansion containing it. With the fix those characters are sent directly and neither happens.

On the machine I found this on that was `€` and `£` being remapped on every use, plus anything else the layout puts on level 3.

**To put a number on it**, since that's the part your question is really about. `__ignoreRemap` is meant to stop AutoKey reacting to its *own* `XChangeKeyboardMapping`, but it's cleared when sending finishes and the `MappingNotify` arrived about half a second later — so AutoKey treated its own remap as external and regrabbed everything. On this machine that is **8.98 seconds and 10,666 `XGrabKey` round-trips**, per expansion containing an AltGr character. This PR removes the trigger outright for those characters, so the saving there is the whole stall rather than a fraction of it.

I should be straight that part of that figure is contingent: it's that large *because* of the `__ignoreRemap` race. Close the race and AutoKey stops storming itself, and what's left is the cost of the rewrite plus a `MappingNotify` to every other client on the system — smaller, but still work that needn't happen at all.

**What it doesn't.** A character that isn't in the layout at *any* reachable level still needs the remap, so this narrows the set rather than emptying it. And the regrab reaction itself is untouched — anything else that provokes a `MappingNotify` still costs the same 8.98 s.

That reaction is the deeper problem: `on_keys_changed()` regrabs whether or not the mapping actually differs, and in the cases I captured it didn't differ at all. I verified that by snapshotting `xmodmap -pke` and `-pm` around each event — byte-identical before and after.

Worth noting for `develop` specifically: `on_keys_changed()` there calls `__ungrab_all_hotkeys()` and `__delayedInitMappings()` directly rather than through `__enqueue`, so the `time.sleep(0.2)` and the whole regrab run on the X event thread. On 0.96.0 they were queued.

I have a fix for that running locally — compare the mapping and return early when it's unchanged, which also closes the `__ignoreRemap` race — but I've deliberately held it back rather than open yet another PR while these are outstanding. Happy to send it whenever it would be welcome, or to file the measurements as an issue first if that's more useful.

I've also opened #1232 about the remapping never being restored, which is the other half of this area and is a design question rather than something I have a patch for.

By the way, thank you for comments - and sorry for the flurry of my comments. Spending a bit of time ironing out some issues before upgrading.

josephj11:
You're doing great although I can't guarantee all of it will make it into 0.97.0. If not, it should be in 0.97.1.

As for the races, @chrislee35 has been dealing with several of those lately.

From your description, they are preexisting problems in our codebase mostly separate from the current set of PRs, so they should probably have their own issues and PRs.

The more atomic we keep things, the easier it is to  manage and review. If something doesn't work out, then fixing it or reverting it doesn't also pull out other things that do work.

When we first got going on reviving 0.97.0, everything was out of sync between master and develop and we got one PR with over 70 commits in it. It was a nightmare to review. I never want to have to deal with that again!

bjohas:
A correction to my earlier comment, before it misleads anyone.

I said that on `develop`, `on_keys_changed()` calls `__ungrab_all_hotkeys()` and `__delayedInitMappings()` directly rather than through `__enqueue`, so the regrab runs on the X event thread, and that 0.96.0 queued it. **That is wrong.**

Both methods are decorated `@queue_method(queue)`, and the decorator body is:

```python
def wrapper(*args):
    queue.put_nowait((func, args))
```

So calling them enqueues, exactly as 0.96.0's explicit `__enqueue(...)` did. `develop` refactored from explicit enqueueing to a decorator; it did not move the work onto the event thread. I read the call sites and not the decorator, and I offered it as a reason the stall might be worse on `develop`, which it is not.

What is true on both branches is that the `time.sleep(0.2)` in `on_keys_changed()` itself runs inline, before anything is queued.

Apologies for the noise. The measurements in that comment stand: I took them on 0.96.0, where the regrab was queued, so they were never dependent on the claim I got wrong.

I have since filed the two underlying problems separately, as you suggested: #1233 for regrabbing when the mapping has not changed, and #1234 for AutoKey reacting to its own remap. PRs #1235 and #1236.

chrislee35:
Reviewed and verified, including live.

1. Confirmed the bug directly: `__build_usable_offsets` really did require `code == 108` before it would enable the AltGr levels.
2. Ran the 6 new `_usable_offsets` tests and the full suite. All pass, no regressions.
3. Confirmed live on a real X11 VM (Linux Mint Cinnamon), no artificial setup needed. Its actual keymap already has `ISO_Level3_Shift` on keycode 92, not 108. Ran both the old and new logic directly against the live display:

```
OLD usableOffsets: (0, 1)
NEW usableOffsets: (0, 1, 4, 5)
```

The fix is correct: accepting any keycode whose own symbol is `ISO_Level3_Shift` matches the actual intent instead of assuming where the key lives. Good catch on the silent keyboard-mapping corruption too, that's a nasty side effect to have tracked down.

Looks ready to merge.

josephj11:
@bjohas and @chrislee35  Looks like I may have merged PRs out of order. LMK how to fix it. Sorry.

chrislee35:
This PR now conflicts with `develop` after #1237, #1235, and #1236 merged ahead of it.

1. @bjohas: Rebase `fix/altgr-detection-develop` onto current `develop`.
2. In `lib/autokey/interface.py`, keep both the `_usable_offsets`/`__build_usable_offsets` change here and the `__remap_characters` change from #1236.
3. In `tests/test_interface.py`, keep both this PR's `test_usable_offsets` and the `TestOnKeysChanged`/`TestSelfRemapDoesNotRegrab` classes from #1235/#1236.
4. Push the rebased branch and confirm CI passes.
5. Merge once green.

josephj11:
@chrislee35 Do we ever use revert? Couldn't the changes be reverted then reapplied in the right order?

chrislee35:
Technically yes, it is possible. It will cause problems if anyone patched against our current develop branch. I don't recommend this, but here are the steps:

1. git fetch origin && git checkout develop && git reset --hard origin/develop
2. git revert -m 1 e8d1c53 # (revert #1236 merge)
3. git revert -m 1 69154d5 # (revert #1235 merge)
4. git push origin develop
5. gh pr merge 1231 --merge # (should be conflict-free now)
6. git fetch origin && git checkout develop && git pull
7. git merge --no-ff fix/regrab-storm-develop # resolve any conflicts against #1231's changes, then git push origin develop
8. git merge --no-ff fix/self-remap-regrab-develop # resolve any conflicts, then git push origin develop

Run the test suite / CI after each reapply (steps 5, 7, 8) before moving to the next.

josephj11:
@chrislee35 I wouldn't do anything this delicate myself. Mainly, I just want to understand. Why is it a bad idea if we effectively undo it and then fix it immediately after? AFAIK, this would only affect patches done over a few days. Would there be consequences beyond that?  Is revert actually good for anything?
----
I just asked an AI and it explained it enough for me. It looks good, but it's a bad idea.

Issue #1233, opened 2026-09-19
on_keys_changed() regrabs every hotkey on any MappingNotify, even when the mapping has not changed


bjohas:
## Classification

Bug (performance / responsiveness)

## Version

`develop` (`3cf0630`); also present in 0.96.0. Measured on 0.96.0, code checked on both.

## Summary

`on_keys_changed()` reacts to every `MappingNotify` by ungrabbing, re-reading and regrabbing every hotkey. But `MappingNotify` means the mapping *may* have changed, not that it did. Clients re-apply the same keyboard mapping wholesale, and AutoKey does a full regrab pass for each one.

On this machine a single such event costs **8.98 seconds** and **10,666 `XGrabKey` round-trips**, for a rewrite that is byte-identical every time.

## Why it matters beyond the wasted work

While that pass runs, the X event queue is not being drained. Further keypresses land inside the window, so pressing a key repeatedly makes it worse rather than better. `on_keys_changed()` also sleeps 0.2 s before starting, on whatever thread delivered the event.

This is, I think, the mechanism behind AutoKey "becoming unresponsive": it is alive and busy, not wedged. I had a wrapper script to restart it periodically for years before finding this.

## Evidence that the mapping had not changed

I snapshotted `xmodmap -pke` and `xmodmap -pm` immediately before and after each event. Byte-identical in every capture. The events were real; the change was not.

The log also shows what AutoKey thinks of its own work:

```
10,690 Xlib.error.BadAccess on major_opcode 33 (GrabKey) in one session
```

`BadAccess` on `GrabKey` means the grab is already held: AutoKey colliding with its own grabs, because a second pass starts while the first is still going.

## Reproducing

Any expansion that sends synthetic keys provokes it here; a three-letter abbreviation cost 10,864 round-trips. The emitter of the redundant `MappingNotify` on this machine is most likely `ibus-x11 --xim`, but I have not pinned that down, and I do not think it matters: re-applying a keyboard mapping is a legitimate thing for a client to do.

## Suggested fix

Remember the mapping and compare before doing any of the work. If it cannot be read, or has never been read, fall through to the regrab rather than skipping it, so an unreadable mapping cannot silently stop AutoKey reacting to genuine layout changes.

I have this running here and will open a PR. Measured after the change: the `BadAccess` flood is gone, and a session shows only the benign `BadWindow` races from grabbing windows that disappear mid-walk.

## Scope

[...]

Issue #1234, opened 2026-09-19
AutoKey regrabs every hotkey in response to its own keyboard remapping: __ignoreRemap loses the race


bjohas:
## Classification

Bug (performance / responsiveness)

## Version

`develop` (`3cf0630`); also present in 0.96.0.

## Summary

`__sendString()` rewrites the keyboard mapping to borrow spare keycodes for characters the current layout cannot reach. `__ignoreRemap` exists so that AutoKey does not react to the `MappingNotify` its own rewrite provokes.

It loses the race. The flag is set before the rewrite and cleared once the string has finished sending, but the event arrives asynchronously. On this machine it was consistently about **half a second** late, by which time the flag was already `False` and AutoKey treated its own remap as another client's, doing a full ungrab and regrab of every hotkey.

Sequence:

```
__remap_characters()   __ignoreRemap = True
change_keyboard_mapping()
...string sends...
__sendString() ends    __ignoreRemap = False
  ~0.5 s later         MappingNotify arrives -> flag is already False -> full regrab
```

## Cost

The regrab is the expensive part: 8.98 s and 10,666 `XGrabKey` round-trips here. So every expansion containing a character the layout cannot reach pays that, on top of the keymap rewrite itself. In my own library 51 characters need remapping.

## Relationship to the other issue

I have filed the general case separately as #1233: `on_keys_changed()` regrabs even when the mapping has not changed at all. These are different defects and I would not want them confused:

- that one is about events from **other** clients, where the mapping is unchanged
- this one is about AutoKey's **own** rewrite, where the mapping genuinely has changed, so a comparison alone does not help

Fixing the first does not fix this.

## Suggested fix

Rather than tightening when the flag is cleared - which would still be a race, only a narrower one - record the mapping AutoKey has just written. The event it provokes then compares equal to what we last saw and returns early. That treats the problem as a question of what the events mean rather than of timing.

This does build on the comparison introduced by the fix for #1233, so the two want to go in that order.

## An aside on `__ignoreRemap`

The flag is doing two jobs: suppressing our own remap, and guarding re-entrancy while `on_keys_changed()` is mid-pass (it sets the flag itself, and `__delayedInitMappings()` clears it). That overloading is part of why the first job is easy to get wrong. I have not touched it, since the fix above makes the first job unnecessary, but it may be worth a look separately.

<br/>
<hr/>

[...]

PR #1235, opened 2026-09-19
Do not regrab every hotkey when the keyboard mapping has not changed

bjohas:
Fixes #1233.

### The change

`on_keys_changed()` reacts to every `MappingNotify` by ungrabbing, re-reading and regrabbing every hotkey. `MappingNotify` means the mapping *may* have changed, not that it did, and clients re-apply the same mapping wholesale.

Remember the mapping and compare before doing any of that work.

### Measurements

On this machine, one such event cost **8.98 seconds** and **10,666 `XGrabKey` round-trips**, for a rewrite that was byte-identical every time. Verified byte-identical by snapshotting `xmodmap -pke` and `xmodmap -pm` either side of each event.

Before: 10,690 `Xlib.error.BadAccess` on `major_opcode 33` in one session, which is AutoKey colliding with its own grabs because a second pass starts while the first is still running.

After, over a 24-minute session: 196 errors, all `BadWindow`, in two short bursts. `BadWindow` on `GrabKey` is the benign race of grabbing a window that disappeared during the tree walk. The pathological signature is gone.

### Failure direction

When the mapping cannot be read, or has never been read, the code falls through to the regrab rather than skipping it. An unreadable mapping should not be able to silently stop AutoKey reacting to a genuine layout change: the cost of being wrong that way is a hotkey that stops working, which is far worse than a wasted regrab.

### Tests

Five cases in `tests/test_interface.py`, no X server needed: unchanged mapping does not regrab; changed mapping does; the new mapping is remembered for next time; a first event with no baseline regrabs; and an unreadable mapping regrabs rather than skipping.

Two of the five fail without the change, three pass either way because they assert existing behaviour is preserved.

### Scope

This addresses mapping events from **other** clients. AutoKey's own remapping genuinely does change the mapping, so this comparison does not help there; that is #1234 and a separate PR that builds on this one.

Note that `on_keys_changed()` also sleeps 0.2 s before starting, on whatever thread delivered the event. I have left that alone.


josephj11:
@chrislee35 0.97.1 stuff I think.

chrislee35:
Reviewed and verified, including live.

1. Confirmed `on_keys_changed()` does the full ungrab/regrab on every `MappingNotify` unconditionally, no comparison at all.
2. Confirmed the fail-safe direction is right: an unreadable or never-read mapping falls through to the regrab rather than skipping it.
3. Ran the 5 new tests and the full suite. All pass, no regressions.
4. Confirmed live on a real X11 VM (Linux Mint Cinnamon): forced an external `MappingNotify` by re-applying the identical keymap with `xmodmap`. The first event went through a full regrab (no baseline yet for that moment), then five identical events in a row all logged "Keymap change event with no actual change - not regrabbing." Exactly the intended behavior.

Those measurements are pretty striking. Good fix, and I like that it's scoped to just the "did anything actually change" question rather than trying to guess whose event caused it.

Looks ready to merge.

PR #1236, opened 2026-09-19
Do not regrab hotkeys in response to AutoKey's own keyboard remapping

bjohas:
Fixes #1234. Builds on #1235, which should go first.

Because it is stacked on that branch, this PR currently shows **both** commits. Only the second one, `621524a`, belongs to this change; the diff here will reduce to that single commit once #1235 merges. I kept them separate rather than sending one combined change so that either can be reverted without taking the other with it.

### The change

`__sendString()` rewrites the keyboard mapping to borrow spare keycodes for characters the layout cannot reach. `__ignoreRemap` exists to stop AutoKey reacting to the `MappingNotify` that rewrite provokes, but it is cleared when sending finishes while the event arrives asynchronously, about half a second later on this machine. By then the flag is `False` and AutoKey treats its own remap as another client's.

Record the mapping we have just written, so the event compares equal in `on_keys_changed()` and returns early.

### Why this rather than fixing the flag

Clearing `__ignoreRemap` when the event arrives instead of when sending ends would narrow the window rather than close it, and it introduces a new failure: if the expected event never arrives, the flag stays set and the next genuine layout change is ignored.

Comparing the mapping answers the question the code actually cares about, which is whether anything changed, rather than trying to guess which client caused an event.

### Cost of the current behaviour

The regrab is the expensive part: 8.98 s and 10,666 `XGrabKey` round-trips here. Every expansion containing a character the layout cannot reach paid that. In my own library, 51 characters need remapping.

Worth noting that #1231, if it lands, reduces how often this path is taken on layouts where AltGr is not on keycode 108, because those characters stop being treated as unreachable. It does not remove the path.

### Tests

Two cases in `tests/test_interface.py`: a remap records what it wrote, and a call that does not remap leaves the remembered mapping alone. The first fails without the change.

### An aside

`__ignoreRemap` is doing two jobs, suppressing our own remap and guarding re-entrancy while `on_keys_changed()` is mid-pass. This change makes the first job unnecessary but leaves the flag in place, since the second job is still real. Untangling that is a separate question and I have not touched it.


josephj11:
@chrislee35 0.97.1?

chrislee35:
Reviewed and verified, on top of #1235.

1. Confirmed the race directly: `__ignoreRemap` is cleared in `send_string()` once sending finishes, while the `MappingNotify` from `__remap_characters()`'s own write arrives later, after the flag's already reset.
2. Confirmed recording the mapping right after the write is the more robust fix, versus retiming the flag, which would just narrow the same race.
3. Ran the 2 new tests and the full suite. All pass, no regressions.
4. Tried to reproduce live on a real X11 VM, with a partial result worth flagging. I triggered AutoKey's own remap path for real (a phrase needing a borrowed keycode for an unmapped character), and confirmed the character sent correctly. But on that machine the resulting `MappingNotify` arrived in single-digit milliseconds every time, well before `__ignoreRemap` clears, so it kept getting caught by the old flag-based guard instead of exercising the new mechanism. I widened the send with a much longer phrase to try to push past that window and still couldn't hit it, so the ~0.5s delay you measured looks genuinely machine-dependent. I did confirm the mechanism this depends on works live (see #1235), and confirmed by reading the diff that this records the mapping immediately after writing, so the logic holds even though I didn't get to watch this specific race close.

Makes sense to land this right after #1235 given the dependency. Looks ready to merge.

PR #1237, opened 2026-09-19
Do not select or expand a tree path that no longer exists (fixes the GTK segfault in #601)

bjohas:
Fixes #601.

### The crash

`expanded_rows` and `last_open` hold positional `GtkTreePath` strings such as `"3:1:2"`, persisted in settings across restarts. They are restored by resolving them against whatever model exists at the time, which may have been rebuilt since, or loaded from a config that changed on disk. So a remembered path can easily name a row that is no longer there.

`Gtk.TreePath.new_from_string()` does not validate against any model, so a stale path constructs happily and only fails later, inside `select_path()`:

```
Fatal Python error: Segmentation fault
  gi/overrides/Gtk.py:1434 in select_path
  gtkui/configwindow.py:913 in config_modified
  UI_common_functions.py:167 in path_created_or_modified
```

Verified on a plain `Gtk.TreeStore`, no display needed:

```
new_from_string("3:7") on a 1-row store  ->  "3:7"   # constructs, does NOT validate
get_iter_from_string("3:7")              ->  ValueError: invalid tree path
```

There is a second, non-fatal case. An empty string raises rather than returning `None`:

```
Gtk.TreePath.new_from_string("")
  Gtk-CRITICAL: gtk_tree_path_new_from_string: assertion '*path != '\000'' failed
  TypeError: constructor returned NULL
```

That one surfaces in a callback where the exception is swallowed, so the effect is a silently incomplete restore rather than a crash.

### Reproducing it without an external edit

`UI_common_functions` calls `config_modified()` from both `path_created_or_modified` and `path_removed`, neither inside a `try`. **Deleting a folder** makes the remembered path stale by construction, so this does not need an unlucky external editor write to hit.

### The change

`_resolve_tree_path()` resolves a remembered path against the current model using `get_iter_from_string()`, and returns `None` when the row has gone. Callers skip instead of restoring a selection that cannot exist.

Applied at all three sites that restore remembered paths:

| site | what it did |
|---|---|
| startup selection | `select_path(self.last_open)`, guarded only by the model being non-empty |
| `config_modified()` | `select_path(new_from_string(self.last_open))` after `rebuild_tree()`: the crash above |
| `__initTreeWidget()` expand loop | `p = path.new_from_string(row)` then `if not p is None` |

That third guard looks like it already handles this, but it cannot fire: `new_from_string()` **raises** rather than returning `None`, so control never reaches the check.

### Tests

New `tests/test_configwindow.py`, 12 cases against a real `Gtk.TreeStore`: no display, no running application. `_resolve_tree_path` is a static method taking the model precisely so it can be tested this way.

Covered: paths that exist; stale top-level, stale child, over-deep, and the index a deleted folder used to occupy; empty and `None`; malformed; a missing model; and the full #601 sequence of remember → rebuild → restore.

The module needs the gettext `_` builtin at import time (via `settingsdialog`), which the running application normally installs, so the test file installs it. That is why there is no existing GTK test module to append to.

### Notes on #601

#601 was closed in 2021 against #638, titled *"Potential fix #601"*. That PR changed button-label sensitivity in `configwindow.py` and `unset_hotkey`'s `self.modifiers = None` → `[]`. The second targets the `AttributeError: 'NoneType' object has no attribute 'sort'` visible in the reported log, which the reporter had already ruled out:

> These errors appear when reloading on the same window too, without crash. So they are not the errors relating directly to this crash

`set_hotkey()` now also has `if not modifiers: modifiers = []`, so that symptom is handled twice over, while `select_path` stayed unguarded. I do not think anyone was careless here: the log in #601 is dominated by those tracebacks and the `select_path` warnings are two quiet lines in the middle of them. But the crash itself was never addressed, and it is still reachable on `develop` today.

### Scope

- GTK only. The Qt tree does not restore paths this way, which matches #601's original report that Qt was unaffected.
- Behaviour change is limited to the case that currently crashes or silently fails: a path that resolves is still selected and expanded exactly as before.


chrislee35:
Reviewed and verified.

1. Confirmed all three restore sites against current develop: they match exactly as described.
2. Independently reproduced the underlying GTK behavior in a bare Python session, no display needed: `new_from_string("3:7")` constructs fine on a 1-row store with no validation, and only `get_iter_from_string("3:7")` catches it, raising `ValueError`. An empty string raises `TypeError` instead of returning `None`, confirming the existing guard at the third site really can never fire.
3. Ran the 12 new tests against a real `Gtk.TreeStore`. All pass. Full suite green (481 passed), no regressions.

Nice archaeology on #601 too, separating the two tangled symptoms and showing #638 only fixed one of them.

One minor, non-blocking thing: `_resolve_tree_path` parses the string twice, once via `get_iter_from_string` to validate and again via `new_from_string` to build the return value. Could use `model.get_path(iter)` instead and skip the second parse, but it's not worth holding this up over.

Looks ready to merge.

josephj11:
Great! I thought we were going to have to postpone this one.
I'll merge it.

But, can one of you explain this "store" thing to me (in Gitter, etc., not here)? I don't think I've run into it before. It sounds like a database of sorts and I'm not aware of AutoKey using anything like that.

chrislee35:
Certainly!

